PR #27, opened 2019-09-17
D3ASIM-1429: Refactoring variable names.

faizan2590:


d3a_interface/constants_limits.py
```diff
@@ -77,21 +78,22 @@ class StorageSettings:
...
+        # Energy sell-range, storage never sell outside this limit.
         # Unit is ct/kWh.
-        BREAK_EVEN_BUY = 24.9
-        # Minimum acceptable buy rate for the battery, in ct/kWh.
-        MIN_BUYING_RATE = 24.9
-        # Maximum acceptable sell rate for the battery, in ct/kWh.
-        MAX_SELLING_RATE = 30
-        # Min allowed battery SOC, range is [0, 0.1].
-        MIN_ALLOWED_SOC = 0.1
+        SELLING_RANGE = [30, 25]
```
hannesdiedrich: Please add some descriptive comments here ( for all new settings, apart from the ranges). Thanks!

d3a_interface/constants_limits.py
```diff
@@ -77,21 +78,22 @@ class StorageSettings:
...
+        INITIAL_SELLING_RANGE = [0, 1000]
+        FINAL_SELLING_RANGE = [0, 1000]
+        # Min allowed battery SOC, range is [0, 100] %.
+        MIN_ALLOWED_SOC = 10
         # Controls whether energy is sold only on the most expensive market, default is
         # to sell to all markets
         SELL_ON_MOST_EXPENSIVE_MARKET = False
+        RATE_CHANGE_PER_UPDATE = [0, 100]
```
hannesdiedrich: Minor, but as you add this to all the devices, how about making this a GeneralSetting?

d3a_interface/constants_limits.py
```diff
@@ -102,11 +104,13 @@ class LoadSettings:
         # Max load energy rate, in ct/kWh
         FINAL_BUYING_RATE = 35
         FINAL_BUYING_RATE_RANGE = [0, 10000]
+        RATE_CHANGE_PER_UPDATE = [0, 100]
 
     class PVSettings:
         PANEL_COUNT_RANGE = [1, 10000]
         MIN_SELL_RATE_RANGE = [0, 10000]
         INITIAL_RATE_RANGE = [0, 10000]
+        RATE_CHANGE_PER_UPDATE = [0, 100]
```
hannesdiedrich: Why the limit to 100? I imagine, that the user wants to update the the price more that 10 Euros for each step, if the Final buying rate can be 10000. Also: Please make it consistent and use 10000 for all the rate ranges (sometimes you used 1000) Thanks!

d3a_interface/constants_limits.py
hannesdiedrich: EDIT, forget it. Its there. The diff view confuses me sometimes. Thanks!

d3a_interface/constants_limits.py
spyrostz: Personally, I think that all these limits are overly exaggerated. Still, there might be the case when the user might want to increase the price more than 1 Euro/kWh per update (100 cts/kWh -> 1 Euro/kWh). Therefore I agree with Hannes to harmonise this to the other rate parameters.

PR #30, opened 2019-09-26
Feature/d3 asim 1478

hannesdiedrich:


d3a_interface/constants_limits.py
```diff
@@ -39,6 +39,8 @@ class GeneralSettings:
         # Boolean flag which forces d3a to run in real-time
         RUN_REAL_TIME = False
         KEEP_PAST_MARKETS = False
+        # Boolean flag which forces d3a to dispatch events via redis channels
+        EVENT_DISPATCHING_VIA_REDIS = False
```
faizan2590: Judging from the comment one line above, should this flag be set to `True` by default?

hannesdiedrich: No. We dont want to despatch events via redis on normal circumstances. Only when deploying, we set it to True.

PR #31, opened 2019-09-27
D3ASIM-1524: Changing ranges' constant parameter from list to namedtu…

faizan2590:
…ple data-structure.


d3a_interface/constants_limits.py
```diff
@@ -19,6 +19,9 @@
 import sys
 from datetime import date, datetime
 from pendulum import duration
+from collections import namedtuple
+
+Ranges = namedtuple('Ranges', ('min', 'max'))
```
spyrostz: Trivial, but Ranges is a quite generic name. IMO RangeLimit looks better to my eyes.
PS: you can merge this now, and do the refactoring in the following PR, if you are in a rush

PR #34, opened 2019-10-01
D3ASIM-1524: Adding a common interface for devices' parameter's valid…

faizan2590:
…ation

### DIFF
--- a/d3a_interface/device_validator.py
+++ b/d3a_interface/device_validator.py
@@ -16,8 +16,8 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 
-import json
 from d3a_interface.constants_limits import ConstSettings
+from d3a_interface.exceptions import DeviceException
 
 
 GeneralSettings = ConstSettings.GeneralSettings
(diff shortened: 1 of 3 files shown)


d3a_interface/device_validator.py
```diff
@@ -0,0 +1,245 @@
...
+LoadSettings = ConstSettings.LoadSettings
+PvSettings = ConstSettings.PVSettings
+StorageSettings = ConstSettings.StorageSettings
+
+
+def validate_load_device(**kwargs):
+    if ("avg_power_W" in kwargs and kwargs["avg_power_W"] is not None) and \
+            not (LoadSettings.AVG_POWER_RANGE.min <= kwargs["avg_power_W"] <=
+                 LoadSettings.AVG_POWER_RANGE.max):
+        return json.dumps({"mis_configuration": [f"avg_power_W should be in between "
```
spyrostz: Why not raise an exception here ? That way it will be far easier for the calling code to just catch the exception. From the d3a and d3a-web PRs, I saw that a boolean variable was returned and an exception was raised according to the variable. If you raise a ValueError (or, even better, a new exception type eg. D3AValidateException) you will not have to care about raising an exception on the calling code at least on d3a. On d3a-web you will just have to catch this exception and re-raise it as a GraphQLError.

tests/test_device_validator.py
```diff
@@ -0,0 +1,185 @@
...
+GeneralSettings = ConstSettings.GeneralSettings
+LoadSettings = ConstSettings.LoadSettings
+PvSettings = ConstSettings.PVSettings
+StorageSettings = ConstSettings.StorageSettings
+
+
+class TestValidateDeviceSettings(unittest.TestCase):
+
+    def test_load_device_setting(self):
+        assert validate_load_device(avg_power_W=-1) == \
```
spyrostz: If this was the reason for returning a boolean value instead of an exception, you can check whether an exception is raised with the context manager assertRaises: 
`with self.assertRaises(ValueError):
    validate_load_device(avg_power_W=-1)`

tests/test_device_validator.py
```diff
@@ -0,0 +1,136 @@
...
+            validate_storage_device(final_buying_rate=15,
+                                    final_selling_rate=14)
+
+        with self.assertRaises(D3ADeviceException):
+            validate_storage_device(energy_rate_increase_per_update=-5)
+
+        with self.assertRaises(D3ADeviceException):
+            validate_storage_device(energy_rate_decrease_per_update=-3)
+
+        with self.assertRaises(D3ADeviceException):
```
hannesdiedrich: @faizan2590 Can you please add a test for the case, where it shouldn't raise an Exception?
Thanks!

faizan2590: Okay, doing it!

PR #35, opened 2019-10-07
D3ASIM-1544: Improving device_validator and adding more related unitt…

faizan2590:
…ests.

### DIFF
--- a/d3a_interface/device_validator.py
+++ b/d3a_interface/device_validator.py
@@ -64,7 +64,7 @@ def validate_load_device(**kwargs):
                                    f"{LoadSettings.INITIAL_BUYING_RATE_RANGE.max}"]})
     if ("initial_buying_rate" in kwargs and kwargs["initial_buying_rate"] is not None) and \
             ("final_buying_rate" in kwargs and kwargs["final_buying_rate"] is not None) and \
-            kwargs["initial_buying_rate"] > kwargs["final_buying_rate"]:
+            (kwargs["initial_buying_rate"] > kwargs["final_buying_rate"]):
         raise D3ADeviceException({"mis_configuration": [f"initial_buying_rate should be "
                                                         f"less than final_buying_rate."]})
     if ("hrs_of_day" in kwargs and kwargs["hrs_of_day"] is not None) and \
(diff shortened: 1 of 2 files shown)


d3a_interface/device_validator.py
```diff
@@ -155,85 +155,85 @@ def validate_pv_device(**kwargs):
...
+            (StorageSettings.MIN_SOC_RANGE.min <= kwargs["min_allowed_soc"] <=
+             StorageSettings.MIN_SOC_RANGE.max):
         raise D3ADeviceException(
             {"mis_configuration": [f"min_allowed_soc should be in between "
                                    f"{StorageSettings.MIN_SOC_RANGE.min} & "
                                    f"{StorageSettings.MIN_SOC_RANGE.max}."]})
     if ("initial_soc" in kwargs and kwargs["initial_soc"] is not None) and \
             ("min_allowed_soc" in kwargs and kwargs["min_allowed_soc"] is not None) and \
-            kwargs["initial_soc"] < kwargs["min_allowed_soc"]:
+            (kwargs["initial_soc"] < kwargs["min_allowed_soc"]):
```
spyrostz: Parentheses here do not hurt, therefore it is ok as it is, but I would like to point out that they are not really needed, because the `and` operator has lower precedence than `<` and comparisons in general. Better safe than sorry though, leave it as it is.

spyrostz:
> @spyrostz @faizan2590 What do you think on including the scenario and project name into the error messages?

@hannesdiedrich I agree that this is potentially useful information. I would suggest to add these here too. The only problem is that these objects cannot access the database, neither do they have information about projects and configuration names. Therefore, I would suggest to raise/catch/reraise these exceptions from a place where this information is accessible.

hannesdiedrich:
https://github.com/gridsingularity/d3a-interface/pull/35#issuecomment-539008961
@spyrostz Actually I meant parsing these two information via a function parameter. But I see now, that this validate function is called somewhere in d3a-web were these information are not at hand now. So i suggest to add another enhancement story to deal with this. What do you think?

spyrostz:
> [#35 (comment)](https://github.com/gridsingularity/d3a-interface/pull/35#issuecomment-539008961)
> @spyrostz Actually I meant parsing these two information via a function parameter. But I see now, that this validate function is called somewhere in d3a-web were these information are not at hand now. So i suggest to add another enhancement story to deal with this. What do you think?

@hannesdiedrich that's fine by me, if it is outside the scope of this story then a new enhancement story should be created.

hannesdiedrich:
LGTM

PR #42, opened 2019-10-28
D3ASIM-1593: Adding device validations for CEP/MMR/InfinitePower.

faizan2590:


d3a_interface/device_validator.py
```diff
@@ -265,3 +267,85 @@ def validate_storage_device(**kwargs):
...
+        raise D3ADeviceException(
+            {
+                "misconfiguration": [f"energy_rate should be in between "
+                                     f"{CepSettings.ENERGY_RATE_RANGE.min} & "
+                                     f"{CepSettings.ENERGY_RATE_RANGE.max}."]})
+
+
+def _validate_rate_profile(energy_rate_profile):
+    for date, value in energy_rate_profile.items():
+        print(f"date: {date}")
```
hannesdiedrich: Please delete

d3a_interface/device_validator.py
```diff
@@ -265,3 +267,85 @@ def validate_storage_device(**kwargs):
...
+        if date == "filename":
```
hannesdiedrich: There could also be another entry in the profile that is not the "filename" entry, the header.
So, i suggest to duplicate or move the check for the date format from the d3a.strategy.read_user_profile here.

hannesdiedrich:
@faizan2590 How about the second part of the ticket: `Please change the naming of SELLING_RANGE and BUYING_RANGE constants and also the naming of the NamedTuple RangeLimit  from min/max to initial/final` ?

d3a_interface/device_validator.py
```diff
@@ -265,3 +267,85 @@ def validate_storage_device(**kwargs):
...
+            continue
+        value = float(value) if type(value) == str else value
+        if not CepSettings.ENERGY_RATE_RANGE.min <= value <= CepSettings.ENERGY_RATE_RANGE.max:
```
hannesdiedrich: Why not reusing the `_validate_rate` function here?

hannesdiedrich: Ah, i see, you want to show the timestamp as well. Ok. Forget it then.

d3a_interface/device_validator.py
```diff
@@ -265,3 +267,85 @@ def validate_storage_device(**kwargs):
...
+            type(kwargs["grid_connected"]) is not bool:
+        raise D3ADeviceException(
+            {"misconfiguration": [f"grid_connected must be a boolean value."]})
+
+
+def validate_finite_diesel_generator(**kwargs):
+    if "max_available_power_kW" in kwargs and kwargs["max_available_power_kW"] is not None:
+        if type(kwargs["max_available_power_kW"]) is float or \
+                type(kwargs["max_available_power_kW"]) is int:
+            if not (CepSettings.MAX_POWER_KW_RANGE.min <= kwargs["max_available_power_kW"] <=
```
hannesdiedrich: I know, I should have mentioned this earlier, but a similar check is done a lot of times in the d3a-interface. How about of abstracting this away into a function that only takes the inputs [variable_name, value, min_limit, max_limit]?

d3a_interface/device_validator.py
```diff
@@ -265,3 +267,85 @@ def validate_storage_device(**kwargs):
...
+        if not CepSettings.ENERGY_RATE_RANGE.min <= value <= CepSettings.ENERGY_RATE_RANGE.max:
+            raise D3ADeviceException(
+                {"misconfiguration": [f"energy_rate should at time: {date} be in between "
+                                      f"{CepSettings.ENERGY_RATE_RANGE.min} & "
+                                      f"{CepSettings.ENERGY_RATE_RANGE.max}."]})
+
+
+def validate_commercial_producer(**kwargs):
+    if "energy_rate" in kwargs and kwargs["energy_rate"] is not None:
+        if type(kwargs["energy_rate"]) is float or type(kwargs["energy_rate"]) is int:
```
spyrostz: Better to use isinstance instead of type for checking the type, since it allows you to compare multiple types at once.

d3a_interface/device_validator.py
faizan2590: Sorry, I already discussed offline with @spyrostz . If you have a deeper look, I have skipped validating uploaded_profile because of its complexity.
This should have already removed, so I will remove this check now

d3a_interface/device_validator.py
```diff
@@ -42,26 +45,29 @@ def validate_load_device(**kwargs):
...
-            not (LoadSettings.HOURS_RANGE.min <= kwargs["hrs_per_day"] <=
-                 LoadSettings.HOURS_RANGE.max):
+            not _validate_range_limit(LoadSettings.HOURS_RANGE.initial,
+                                      kwargs["hrs_per_day"],
+                                      LoadSettings.HOURS_RANGE.final):
         raise D3ADeviceException(
             {"misconfiguration": [f"hrs_per_day should be in between "
-                                  f"{LoadSettings.HOURS_RANGE.min} & "
-                                  f"{LoadSettings.HOURS_RANGE.max}."]})
+                                  f"{LoadSettings.HOURS_RANGE.initial} & "
```
hannesdiedrich: Sorry, maybe I wasn't as explicit as i should have been. IMO For the HOURS_RANGE initial and final doesnt make sense. 
I thought you ad another Rangelimit for the rates and keep this one for the non-rate values. Does it make sense?

d3a_interface/device_validator.py
hannesdiedrich: Sorry, What happened to this comment. Do you plan to implement it?

hannesdiedrich: Ah now I see. You only abstracted the min<=x<=max check away. But what i meant was also to put in the raise of the D3ADeviceException in there. That is why I proposed 4 inout variables.

d3a_interface/device_validator.py
faizan2590: Im not sure if its really that important, at least from the wordings of the ticket `Please change the naming of SELLING_RANGE and BUYING_RANGE constants and also the naming of the NamedTuple RangeLimit  from min/max to initial/final`. I had an impression to simply change min/max to initial/final. 
If its important to classify in two different namedtuples then please also tell me where it need to (min, max) namedtuple and where it should be (initial, final)

d3a_interface/constants_limits.py
```diff
@@ -46,8 +47,8 @@ class GeneralSettings:
         # Boolean flag which forces d3a to dispatch events via redis channels
         EVENT_DISPATCHING_VIA_REDIS = False
 
-        RATE_DECREASE_PER_UPDATE_RANGE = RangeLimit(0, 1000)
-        RATE_CHANGE_PER_UPDATE = RangeLimit(0, 1000)
+        RATE_DECREASE_PER_UPDATE_RANGE = RateRange(0, 1000)
+        RATE_CHANGE_PER_UPDATE = RateRange(0, 1000)
```
spyrostz: Isn't this a range too ? Then why not being consistent with the naming above and renaming it to RATE_CHANGE_PER_UPDATE_RANGE ?

d3a_interface/constants_limits.py
```diff
@@ -67,7 +68,7 @@ class StorageSettings:
         # possible range of state of charge
         INITIAL_CHARGE_RANGE = RangeLimit(10, 100)
 
-        BREAK_EVEN_RANGE = RangeLimit(0, 10000)
+        BREAK_EVEN_RANGE = RateRange(0, 10000)
```
spyrostz: min/max are ok here, therefore you can keep the RangeLimit namedtuple

d3a_interface/constants_limits.py
```diff
@@ -76,14 +77,16 @@ class StorageSettings:
         MAX_ABS_POWER_RANGE = RangeLimit(0.0001, 2000000)
         # Energy buy-range, storage never buys outside this limit.
         # Unit is ct/kWh.
-        BUYING_RANGE = RangeLimit(0, 24.9)
-        INITIAL_BUYING_RANGE = RangeLimit(0, 10000)
-        FINAL_BUYING_RANGE = RangeLimit(0, 10000)
+        INITIAL_BUYING_RATE = 0
```
spyrostz: you could keep the BUYING_RATE variable and just assign a RateRange to it. That way you could access its parameters like BUYING_RATE.initial and BUYING_RATE.final, which is much more readable and easily maintainable IMO compared to 2 separate variables.

d3a_interface/constants_limits.py
```diff
@@ -76,14 +77,16 @@ class StorageSettings:
...
+        FINAL_BUYING_RATE = 24.9
+        INITIAL_BUYING_RANGE = RateRange(0, 10000)
+        FINAL_BUYING_RANGE = RateRange(0, 10000)
         # Energy sell-range, storage never sell outside this limit.
         # Unit is ct/kWh.
-        SELLING_RANGE = RangeLimit(30, 25)
-        INITIAL_SELLING_RANGE = RangeLimit(0, 10000)
-        FINAL_SELLING_RANGE = RangeLimit(0, 10000)
+        INITIAL_SELLING_RATE = 30
```
spyrostz: Same as BUYING_RATE

d3a_interface/constants_limits.py
hannesdiedrich: Maybe I am to blame here because i might have confused Faizan.
My proposal was: Create a Raterange namedtuple with "intital" and "final" memebers and assign it to only `BUYING_RANGE` and `SELLING_RANGE` . The other limits should be a Limitrange or as you called it a Rangelimit with members "min" and "max". I hope this is clear now.

d3a_interface/constants_limits.py
faizan2590: Thanks for the catch, `RATE_CHANGE_PER_UPDATE ` isn't used anymore. If you agree, I will remove it.

d3a_interface/constants_limits.py
faizan2590: `BREAK_EVEN_RANGE` isnt used anymore too, if you agree, I will remove it too

d3a_interface/constants_limits.py
faizan2590: Okay, to be 100% clear this this time and to be on the same page. Following changes needs to be done:

- Replace `INITIAL_BUYING_RATE` & `FINAL_BUYING_RATE` with `BUYING_RATE_RANGE = RateRange(0, 24.9)`.
- Replace `INITIAL_SELLING_RATE` & `FINAL_SELLING_RATE` with `SELLING_RATE_RANGE = RateRange(30, 25)`

d3a_interface/constants_limits.py
faizan2590: Sorry, Im keeping/updating `RATE_CHANGE_PER_UPDATE_RANGE`, as it looks more generic to me compared to `RATE_DECREASE_PER_UPDATE_RANGE `

d3a_interface/constants_limits.py
```diff
@@ -76,14 +75,14 @@ class StorageSettings:
         MAX_ABS_POWER_RANGE = RangeLimit(0.0001, 2000000)
         # Energy buy-range, storage never buys outside this limit.
         # Unit is ct/kWh.
-        BUYING_RANGE = RangeLimit(0, 24.9)
-        INITIAL_BUYING_RANGE = RangeLimit(0, 10000)
-        FINAL_BUYING_RANGE = RangeLimit(0, 10000)
+        BUYING_RATE_RANGE = RateRange(0, 24.9)
+        INITIAL_BUYING_RANGE = RateRange(0, 10000)
```
hannesdiedrich: Sorry, no! It is a limit, isnt it=?

d3a_interface/constants_limits.py
```diff
@@ -46,72 +47,66 @@ class GeneralSettings:
...
         # Max battery capacity in kWh.
         CAPACITY = 1.2
-        CAPACITY_RANGE = RangeLimit(0.0001, 2000000)
+        CAPACITY_LIMIT = RangeLimit(0.0001, 2000000)
         # Maximum battery power for supply/demand, in Watts.
         MAX_ABS_POWER = 5
-        MAX_ABS_POWER_RANGE = RangeLimit(0.0001, 2000000)
-        # Energy buy-range, storage never buys outside this limit.
+        MAX_ABS_POWER_LIMIT = RangeLimit(0.0001, 2000000)
+        # Energy buy-limit, storage never buys outside this limit.
```
hannesdiedrich: Minor, but this is a range (of course it is a limit somwhow, but in order to not get confused in the future, please change it to range again. Thanks!)

faizan2590: Okay, should it still use `RangeLimit` as a namedtuple or `RateRange`

hannesdiedrich: yes of course!

PR #45, opened 2019-11-12
D3ASIM-1644: Added eta_seconds to ResultsSchema.

hannesdiedrich:


d3a_interface/schemas.py
```diff
@@ -134,6 +134,7 @@ class ResultsSchemas:
                             "bills": {"type": "object"},
                             "tree_summary": {"type": "object"},
                             "status": {"type": "string"},
+                            "eta_seconds": {"type": "number"},
```
faizan2590: One clarification, would this key value pair also contains, simulation queued time?

hannesdiedrich: No, because the simulation queued time is only determined on the fly by d3-web.

PR #46, opened 2019-11-12
D3ASIM-1635: Added check in load device validator.

hannesdiedrich:


tests/test_device_validator.py
```diff
@@ -77,6 +77,18 @@ def test_load_device_setting(self):
...
+            validate_load_device(avg_power_W=100,
+                                 daily_load_profile=[1, 2, 3, 4, 5, 6])
+
+        with self.assertRaises(D3ADeviceException):
+            validate_load_device(hrs_per_day=1,
+                                 daily_load_profile=[1, 2, 3, 4, 5, 6])
+
+        with self.assertRaises(D3ADeviceException):
+            validate_load_device(hrs_of_day=[1, 2, 3, 4, 5, 6],
+                                 daily_load_profile=[1, 2, 3, 4, 5, 6])
```
faizan2590: One more check that would be nice to have a type check via ast.literal_eval is of `daily_load_profile`. If I am right `daily_load_profile`  should be in dict format having key as an operating hour and value as energy_demanded for that hour.

hannesdiedrich: Ok, will take the opportunity to add it.

d3a_interface/device_validator.py
```diff
@@ -113,6 +113,16 @@ def validate_load_device(**kwargs):
             {"misconfiguration": [f"daily_load_profile and all or one [hrs_per_day, hrs_of_day, "
                                   f"avg_power_W] can't be set together."]})
 
+    if "daily_load_profile" in kwargs and kwargs["daily_load_profile"] is not None:
+        print()
+        if isinstance(kwargs["daily_load_profile"], str):
+            _validate_energy_profile(ast.literal_eval(kwargs["daily_load_profile"]))
```
faizan2590: Can you add this `validate_load_device` validator in [DefinedLoadStrategy](https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/strategy/predefined_load.py#L79) where `daily_load_profile` could be a profile path.

hannesdiedrich: Sorry, i dont get it. 
In the d3a the `read_arbitrary_profile` is dealing with reading and validating the daily_load_profile. 
This is why i suggested to you in one of your last PRs in this repo to move this function from the d3a to here. 
But you had a good reason not to. But i can't remember. Can you?

faizan2590: I can't recall it either, well then skip it for now and we can look into later when needed.

PR #50, opened 2019-12-05
D3ASIM-1756: Added DISPATCH_EVENTS_BOTTOM_TO_TOP flag.

hannesdiedrich:


spyrostz:
This thing is in my mind for a while now, so I will mention it here since it starts to exhibit more and more. D3a-interface is supposed to be a repo for code that is shared across d3a and d3a-web (of course there is also code there that we potentially plan to add for the future, but this is not the norm). Therefore I would suggest to use d3a.constants module for this variable, since this will probably never be exposed via the d3a-web.

PR #54, opened 2020-01-15
Feature/d3 asim 973

spyrostz:


d3a_interface/sim_results/aggregate_results.py
```diff
@@ -0,0 +1,198 @@
...
+        if area_uuid not in global_pe:
+            global_pe[area_uuid] = deepcopy(market_pe[area_uuid])
+        else:
+            global_pe[area_uuid]["price-energy-day"].extend(
+                market_pe[area_uuid]["price-energy-day"])
+    return global_pe
+
+
+def merge_device_statistics_results_to_global(market_device, global_device):
+    if not global_device:
```
hannesdiedrich: Minor, but:
`if not {}` yields True. This was new to me. Not sure what I think of it considering the readability. 
Also, the name "*_device" was confusing at first. Maybe append a `_dict` everywhere (in order to not append `_statistics` that would be too long I understand.)

d3a_interface/utils.py
```diff
@@ -0,0 +1,49 @@
+"""
+Copyright 2018 Grid Singularity
```
faizan2590: Im not sure, does it needs to be updated as per current year?

spyrostz: I can ask Ana, but I think that this should refer to the start date of the licensing.

d3a_interface/sim_results/aggregate_results.py
spyrostz: Regarding the latter comment, I would suggest to introduce `typing` here, and to denote that these are dicts. That way we can have that information and a potential type check (if we decide to use it in the future)

PR #76, opened 2020-03-24
D3ASIM-1967: Added validation for the constant grid fee parameter.

spyrostz:


d3a_interface/area_validator.py
```diff
@@ -18,17 +18,22 @@
...
         error_message = {"misconfiguration": [f"grid_fee_percentage should be in between "
                                               f"{AreaSettings.GRID_FEE_PERCENTAGE_LIMIT.min} & "
                                               f"{AreaSettings.GRID_FEE_PERCENTAGE_LIMIT.max}."]}
         validate_range_limit(AreaSettings.GRID_FEE_PERCENTAGE_LIMIT.min,
                              kwargs["grid_fee_percentage"],
                              AreaSettings.GRID_FEE_PERCENTAGE_LIMIT.max, error_message)
+
+    if "grid_fee_constant" in kwargs and kwargs["grid_fee_constant"] is not None:
+        if kwargs["grid_fee_constant"] < 0:
+            raise D3ADeviceException("Grid fee constant should be a positive value.")
```
faizan2590: Shouldn't we also restrict user by setting the ridiculously high value  like we have done for device parameters ?

spyrostz: Agreed, setting the upper limit to 200 cts/kWh. Thanks

PR #78, opened 2020-03-31
Feature/d3 asim 2175

hannesdiedrich:


d3a_interface/constants_limits.py
```diff
@@ -122,6 +122,8 @@ class WindSettings:
         MAX_WIND_TURBINE_OUTPUT_W = 160
 
     class IAASettings:
+        GRID_FEE_TYPE = 1
```
spyrostz: Reason for not using an enum is because you cannot json-serialize it, right ?

d3a_interface/constants_limits.py
```diff
@@ -122,6 +122,8 @@ class WindSettings:
...
+        VALID_FEE_TYPES = [1, 2]
```
spyrostz: BTW, can you also add a small comment that defines which value is the constant fees and which is the percentage fees ? Thanks

d3a_interface/constants_limits.py
hannesdiedrich: yes. I dont like the int either, but it is the same for the market_type.

d3a_interface/constants_limits.py
hannesdiedrich: Good idea! Will do!

PR #80, opened 2020-04-14
D3ASIM-2213: Enhancing area validation.

faizan2590:


d3a_interface/utils.py
```diff
@@ -57,3 +57,11 @@ def wait_until_timeout_blocking(functor, timeout=10, polling_period=0.01):
         time.sleep(polling_period)
         current_time += time.time() - start_time
     assert functor()
+
+
+def key_in_dict_and_not_none(d, key):
```
spyrostz, line 62: Have you deleted the corresponding function in d3a-web ? If not it would be great to do so and use the d3a-interface function instead, that way the function will not be duplicated.

PR #84, opened 2020-04-17
D3ASIM-1951: Adding loss_function validator.

faizan2590:


faizan2590:
> Did you try to run dd3a integrationtests with this change?

Yes, and it hasn't complained

PR #94, opened 2020-05-26
D3ASIM-2348: Renaming HIGH_RESOLUTION_PLOTS flag

faizan2590:


d3a_interface/constants_limits.py
```diff
@@ -41,7 +41,7 @@ class GeneralSettings:
         ENERGY_RATE_DECREASE_PER_UPDATE = 1  # rate decrease in cents_per_update
         SETUP_FILE_PATH = None  # Default path of the available setup files
         EXPORT_DEVICE_PLOTS = True
-        EXPORT_ENERGY_TRADE_PROFILE_HR = False
+        HIGH_RESOLUTION_PLOTS = False
```
hannesdiedrich, line 44: Not sure about this renaming. There might be more HR plots in the future.
And we explicitly prependet `EXPORT` before.

faizan2590: Discussed offline, adding a separate flag for OFFER/BID/TRADE graph.

PR #96, opened 2020-06-05
D3ASIM-2231: Adding separate validation for infinite bus.

faizan2590:


d3a_interface/device_validator.py
```diff
@@ -393,11 +393,21 @@ def validate_market_maker(**kwargs):
...
+    if "energy_rate_profile" in kwargs and kwargs["energy_rate_profile"] is not None and \
+            ("energy_rate_profile_uuid" not in kwargs or
+             kwargs["energy_rate_profile_uuid"] is None):
         raise D3ADeviceException(
-            {"misconfiguration": [f"buying_rate can only be configured for InfiniteBus mode."]})
+            {"misconfiguration": [f"energy_rate_profile must have a uuid."]})
+    if key_in_dict_and_not_none(kwargs, "energy_rate_profile_uuid") and not \
+            isinstance(kwargs["energy_rate_profile_uuid"], str):
+        raise D3ADeviceException(
+            {"misconfiguration": [f"energy_rate_profile_uuid must have a string type."]})
```
hannesdiedrich, line 407: Totally minor, but this looks like it could be refactored into its own function (as it is used below too).
If you dont think so, skip it. Thanks!

PR #98, opened 2020-06-08
D3ASIM-2427: Added bids_offers_trades as part of the simulation resul…

spyrostz:
…ts. Moved time conversion functions to d3a-interface in order to be reused by the d3a-web.


d3a_interface/utils.py
```diff
@@ -81,3 +84,34 @@ def str_to_pendulum_datetime(input_str):
...
+    return datetime.format(DATE_TIME_FORMAT_SECONDS)
+
+
+def convert_pendulum_to_str_in_dict(indict, outdict, ui_format=False, unix_time=False):
+    for key, value in indict.items():
+        if isinstance(key, DateTime):
+            outdict[format_datetime(key, ui_format, unix_time)] = indict[key]
+        elif isinstance(value, DateTime):
+            outdict[key] = format_datetime(value, ui_format, unix_time)
+        elif isinstance(indict[key], list):
```
faizan2590, line 109:
```suggestion
        elif isinstance(value, list):
```
For code uniformity.

d3a_interface/utils.py
```diff
@@ -81,3 +84,34 @@ def str_to_pendulum_datetime(input_str):
...
+            outdict[key] = [convert_pendulum_to_str_in_dict(element, {}, ui_format, unix_time)
+                            for element in indict[key]]
+        elif isinstance(indict[key], dict):
```
faizan2590, line 112: [Same](https://github.com/gridsingularity/d3a-interface/pull/98/files#r436543100) applies here.

d3a_interface/utils.py
spyrostz: I agree, and it is about time that we fixed that, will do so. Thanks!

d3a_interface/utils.py
spyrostz: Same as above, will fix this.

PR #108, opened 2020-07-08
D3ASIM-2562: Adding jwt_token expiry as a common variable to be used …

faizan2590:
…across different services.


hannesdiedrich:
As discussed offline, I have some concern exposing this variable to the public. But we can discuss this later when more team members are present.

PR #115, opened 2020-08-11
D3ASIM-1890: Adding check_redis_health method to commonly to used amo…

faizan2590:
…ng different services.

### DIFF
--- a/d3a_interface/utils.py
+++ b/d3a_interface/utils.py
@@ -22,8 +22,9 @@ from copy import copy
 from threading import Timer
 from d3a_interface.constants_limits import DATE_TIME_UI_FORMAT, DATE_TIME_FORMAT, TIME_FORMAT, \
     DATE_TIME_FORMAT_SECONDS
+from redis.exceptions import ConnectionError
 import time
-
+import logging
 
 def convert_datetime_to_str_in_list(in_list, ui_format=False):
     """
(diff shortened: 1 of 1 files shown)


d3a_interface/utils.py
```diff
@@ -132,5 +134,18 @@ def mkdir_from_str(directory: str, exist_ok=True, parents=True):
 class RepeatingTimer(Timer):
     def run(self):
         while not self.finished.is_set():
-            self.function(*self.args, **self.kwargs)
             self.finished.wait(self.interval)
+            self.function(*self.args, **self.kwargs)
```
spyrostz, line 138: Is it really necessary to change the order here? I am not negative on that, however it would be great if you checked the simulation results via the UI after this change, to validate that the GraphQL subscriptions (which also use this timer) work as expected

d3a_interface/utils.py
```diff
@@ -132,5 +134,18 @@ def mkdir_from_str(directory: str, exist_ok=True, parents=True):
...
+
+
+def check_redis_health(redis_db):
+    result = False
```
spyrostz, line 142: If you can rename it to `is_connected` would be great, it would better reflect the intent here

d3a_interface/utils.py
faizan2590: Remnant of testing, reverting it

PR #148, opened 2020-12-28
D3ASIM-3139: migration to github actions

mclrch:


hannesdiedrich (approved):
LGTM

.github/workflows/d3a-interface_ci.yaml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      env:
+        TOXENV: travis
```
spyrostz, line 18: Could be a good time to rename the `travis` tox parameter to something like `ci`

mclrch: nice catch
done

.github/workflows/d3a-interface_ci_cd_development.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      env:
+        TOXENV: travis
+      run: |
+        bash .support/travis-install.sh
+        bash .support/travis-run.sh
```
spyrostz, line 20: I am not sure if it is good to do it in the context of this story, but I think it would be good to get rid of these files somehow. These 2 files actually do 2 things (at the moment at least, some time ago they were also used to install packages related to blockchain):
- Install tox. This operation is probably not even needed, because tox is already a dependency in the Python requirements. 
- Set the environment variable that controls the interface branch that needs to be used. 
Both operations can be performed here as far as I see

spyrostz (approved):
LGTM once the comments are resolved

.github/workflows/d3a-interface_ci_cd_development.yaml
faizan2590, line 20: Sorry if I missing something here but I couldn't find `travis-install.sh` inside `d3a-interface`.

mclrch: fixed
thanks

mclrch: fixed
thanks

PR #150, opened 2021-01-05
D3ASIM-3084: Added check for new settings parameter slot_length_realt…

hannesdiedrich:
…ime.

### DIFF
--- a/d3a_interface/settings_validators.py
+++ b/d3a_interface/settings_validators.py
@@ -15,10 +15,11 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-from d3a_interface.constants_limits import ConstSettings, GlobalConfig
-from d3a_interface.exceptions import D3ASettingsException
 from pendulum import duration, Duration
 from datetime import timedelta
+
+from d3a_interface.constants_limits import ConstSettings, GlobalConfig
+from d3a_interface.exceptions import D3ASettingsException
 
 
 def validate_global_settings(settings_dict):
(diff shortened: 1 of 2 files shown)


Hassan754 (approved):

d3a_interface/settings_validators.py
```diff
@@ -76,6 +82,9 @@ def validate_global_settings(settings_dict):
             int(settings_dict["grid_fee_type"]) not in ConstSettings.IAASettings.VALID_FEE_TYPES:
         raise D3ASettingsException(f'Invalid value for grid_fee_type '
                                    f'({settings_dict["grid_fee_type"]}).')
+    if "slowdown" in settings_dict and settings_dict["slowdown"] > 0 and \
+            slot_length_realtime > GlobalConfig.slot_length_realtime:
+        raise D3ASettingsException("'slowdown' and 'slot_length_realtime' can not be set together")
```
faizan2590, line 87: Can you confirm `slot_length_realtime ` even be greater than the actual market_slot, like `market_slot=5mins & slot_length_realtime =30mins`. Personally it would look super-wierd to me like simulation running slower than the slot_length

hannesdiedrich: IMO there should not be a limit for `slot_length_realtime` as we deprecate slowdown now. There could be the case when the user set the `slot_length` to 30 seconds, but wants to monitor it in slower speed, hence setting `slot_length_realtime` to a higher value. 
If you agree, I will now close this PR.

hannesdiedrich: One addition to that:
Of course, there is always the limit of the maximum simulation speed, meaning if the slot needs 5 mins to pass, the users setting of e.g. slot_length_realtime=1 min will not have an effect.

PR #154, opened 2021-01-17
D3ASIM-2961: Added grid_connected field inside the load schema

Hassan754:


d3a_interface/schemas.py
```diff
@@ -34,16 +34,16 @@ class ScenarioSchemas:
...
+                        "type": "array",
+                        "items": {"anyOf": [
+                            {"$ref": "#/definitions/area"},
+                            {"$ref": "#/definitions/pv"},
+                            {"$ref": "#/definitions/load"},
+                            {"$ref": "#/definitions/infinite_power_plant"},
+                            {"$ref": "#/definitions/finite_power_plant"},
+                            {"$ref": "#/definitions/storage"}
+                        ]}, "default": []},
+                        {"type": "null"}]}
```
hannesdiedrich, line 46: I see a lot of whitespace corrections in your PRs. Are you using a special linter that complains? Or are these just personal preference?

Hassan754: I am using Pycharm's linter , let me check its settings

Hassan754: @hannesdiedrich This is from the Pycharm official docs : 

> PyCharm adheres to PEP8 rules and requirements for arranging and formatting Python code.

I can be extra cautious when refactoring the code in the future

hannesdiedrich: Hmm, so pycharm added these whitespace changes for you? 
I am also using the pycharm linter, but is is only notifying me with grey underlines and it is not notifying me in these places here. Which version of pycharm are you using?

Hassan754: I am using Pycharm 2019.3.1 , maybe try pressing `Command` + `Shift` + `L` ?

fievelk: In pyCharm:
- `Command` + `Option` + `L`: reformat code
- `Command` + `Option` + `Shift` + `L`: reformat file

I think it's not the default behaviour on save, but rather an action that you can trigger manually.

The issue I see with committing this kind of changes is that they hide the history of that specific line in the git tree. For example, if I want to know why some code was implemented/edited, I would usually use git blame or check the history. However, if the line was changed because of linting fixes, I would find commits that are unrelated to that line. In theory, linting should be done when the line is created (but it's not always the case, especially with older code).

hannesdiedrich: Yeah, cant you just omit pressing  Command + Shift + L ?
I agree with Pierpaolo and add, that these changes also mask the real changes in PRs (in this PR it is just one line, right?)

Hassan754: Yes , Deal .

spyrostz: also agree with @fievelk , just to cover all cases here, @Hassan754 can you also check your Pycharm settings? Mine are like this:
tabsvsspaces

Hassan754: Screen Shot 2021-01-18 at 13 05 12

Hassan754: @spyrostz There is an option to export/import settings , maybe if you find this refactoring style convenient , all of us can agree to use it

spyrostz: Good, so your tabs / space settings are intact. No need for exporting the settings since we all use the default Pycharm ones, the only reason why I asked you to check it out was to rule out the case that these settings were modified and were the root cause of the diff (they were not).

d3a_interface/schemas.py
```diff
@@ -104,13 +104,14 @@ class ScenarioSchemas:
...
-                    "avg_power_W":  {"anyOf": [{"type": "number"}, {"type": "null"}]},
-                    "hrs_per_day":  {"anyOf": [{"type": "number"}, {"type": "null"}]},
+                    "avg_power_W": {"anyOf": [{"type": "number"}, {"type": "null"}]},
+                    "hrs_per_day": {"anyOf": [{"type": "number"}, {"type": "null"}]},
                     "hrs_of_day": {"anyOf": [{"type": "array"}, {"type": "null"}]},
                     "initial_buying_rate": {"type": "number"},
                     "final_buying_rate": {"anyOf": [{"type": "number"}, {"type": "null"}]},
                     "fit_to_limit": {"type": "boolean"},
-                    "update_interval":  {"anyOf": [{"type": "number"}, {"type": "null"}]},
+                    "grid_connected": {"type": "boolean"},
```
hannesdiedrich, line 113: So this is the only line that was added, right?
I am still overwhelmed by the whitespace changes. I propose to revert them.

Hassan754: @hannesdiedrich Reverted

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #156, opened 2021-01-27
Outsourced modules iterator to d3a-interface

Hassan754:
Outsourcing modules iterator as it will be used in both `d3a` and `d3a-api-client`


hannesdiedrich (approved):
LGTM

faizan2590 (approved):
lgtm

PR #160, opened 2021-02-10
D3ASIM-3292: Also allowing None values for initial_soc in the storage…

hannesdiedrich:
… representation.

### DIFF
--- a/d3a_interface/schemas.py
+++ b/d3a_interface/schemas.py
@@ -79,7 +79,7 @@ class ScenarioSchemas:
                     "number_of_clones": {"type": "number"},
                     "uuid": {"type": "string"},
                     "libraryUUID": {"anyOf": [{"type": "string"}, {"type": "null"}]},
-                    "initial_soc": {"anyOf": [{"type": "string"}, {"type": "null"}]},
+                    "initial_soc": {"anyOf": [{"type": "number"}, {"type": "null"}]},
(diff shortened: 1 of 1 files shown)


Hassan754 (approved):

d3a_interface/schemas.py
```diff
@@ -79,7 +79,7 @@ class ScenarioSchemas:
                     "number_of_clones": {"type": "number"},
                     "uuid": {"type": "string"},
                     "libraryUUID": {"anyOf": [{"type": "string"}, {"type": "null"}]},
-                    "initial_soc": {"type": "number"},
+                    "initial_soc": {"anyOf": [{"type": "string"}, {"type": "null"}]},
```
spyrostz, line 82: I am a bit surprised that the soc is not sent as a float but as a string. As an extra precaution and in order to protect ourselves from changes in the future, I would recommend to also allow the number type.

spyrostz (dismissed):
One comment, LGTM once resolved

d3a_interface/schemas.py
hannesdiedrich: I actually replaces the string with number, Dont know why this was there.
Also tried it locally, works.

spyrostz (approved):
LGTM

Hassan754 (approved):
LGTM

PR #161, opened 2021-02-10
D3ASIM-3279: Converted bills internal calculation from area name-base…

spyrostz:
…d to uuid-based. Refactored the update method to better describe the steps of the bills calculation process, and added comments accordingly. Added explicit unit tests for some of the bills class methods.

### DIFF
--- a/d3a_interface/sim_results/bills.py
+++ b/d3a_interface/sim_results/bills.py
@@ -267,24 +267,24 @@ class MarketEnergyBills:
         # The flattened dict contains only area_uuid -> energy_bills_dict, no children
         # or aggregated data like Totals, External Trades etc.
         self._flatten_energy_bills(bills, flattened)
-        # Adds children to the flattened dict by iterating over the area_dict, finding out the children
-        # of each area, and copying the children bills under their respective parent. Only 1 level
-        # child hierarchy.
-        # The same function populates the Eternal Trades, Accumulated Trades, Market Fees and Totals
-        # bills for the areas that are not leaves (and have children).
+        # Adds children to the flattened dict by iterating over the area_dict, finding out the
+        # children of each area, and copying the children bills under their respective parent. Only
+        # 1 level child hierarchy.
+        # The same function populates the Eternal Trades, Accumulated Trades, Market Fees and
+        # Totals bills for the areas that are not leaves (and have children).
         # The naming format of the areas is still uuid
         bills = self._accumulate_by_children(area_dict, flattened, {})
         # Keep the state of the unformatted bills in order to be reused by _energy_bills method.
         # We need this in order to read past results when calculating the new energy bills.
         self.current_raw_bills = deepcopy(bills)
-        # Converts the children of the bills from uuids to names, because the UI uses these as headers in
-        # the table where these data are reported.
+        # Converts the children of the bills from uuids to names, because the UI uses these
+        # as headers in the table where these data are reported.
         bills = self._swap_children_uuids_to_names(area_dict, bills)
-        # Converts the keys of the result dict from uuids to names, in order to end up in the CSV files
-        # in a human-readable format.
+        # Converts the keys of the result dict from uuids to names, in order to end up in
+        # the CSV files in a human-readable format.
         self.bills_results = self._bills_local_format(area_dict, bills)
-        # Rounds the precision of the results to 3 decimal points, in order for the UI to report them
-        # correctly.
+        # Rounds the precision of the results to 3 decimal points, in order for the UI to report
+        # them correctly.
         self.bills_redis_results = self._round_results_for_ui(deepcopy(bills))
 
     def restore_area_results_state(self, area_dict, last_known_state_data):
--- a/d3a_interface/sim_results/bills.py
+++ b/d3a_interface/sim_results/bills.py
@@ -16,7 +16,8 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from copy import deepcopy, copy
-from d3a_interface.utils import round_floats_for_ui, key_in_dict_and_not_none
+from d3a_interface.utils import round_floats_for_ui, get_area_uuid_name_mapping, \
+    key_in_dict_and_not_none
 from d3a_interface.sim_results import is_load_node_type, is_pv_node_type, \
     area_name_from_area_or_iaa_name, get_unified_area_type
 from d3a_interface.constants_limits import ConstSettings
@@ -26,7 +27,8 @@ class CumulativeBills:
     def __init__(self):
         self.cumulative_bills_results = {}
 
-    def _calculate_device_penalties(self, area, area_core_stats):
+    @classmethod
+    def _calculate_device_penalties(cls, area, area_core_stats):
         if len(area['children']) > 0 or area_core_stats == {}:
             return 0.0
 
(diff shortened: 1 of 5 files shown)


d3a_interface/sim_results/bills.py
```diff
@@ -250,13 +254,38 @@ def _accumulate_grid_fee_charged(self, area_dict, area_core_stats):
             self._accumulate_grid_fee_charged(child, area_core_stats)
 
     def update(self, area_dict, area_core_stats):
+        # Updates the self.market_fees dict which keeps track of the accumulated market fees for
+        # each area.
         self._update_market_fees(area_dict, area_core_stats)
+        # Calculates the cumulative_fee_charged_per_market , which in reality is the sum of the
+        # grid fees for all markets for the whole simulation duration.
         self._accumulate_grid_fee_charged(area_dict, area_core_stats)
```
hannesdiedrich, line 262: I just had a look into the code of both functions an AFAIU, they are doing the same. 
If you agree, i suggest to get rid of one of them.

d3a_interface/sim_results/bills.py
```diff
@@ -250,13 +254,38 @@ def _accumulate_grid_fee_charged(self, area_dict, area_core_stats):
...
+        # Flatten the energy bill tree.
+        # The flattened dict contains only area_uuid -> energy_bills_dict, no children
+        # or aggregated data like Totals, External Trades etc.
         self._flatten_energy_bills(bills, flattened)
-        self.bills_results = self._accumulate_by_children(area_dict, flattened, {})
-        self._bills_for_redis(area_dict, deepcopy(self.bills_results))
+        # Adds children to the flattened dict by iterating over the area_dict, finding out the
+        # children of each area, and copying the children bills under their respective parent. Only
+        # 1 level child hierarchy.
+        # The same function populates the Eternal Trades, Accumulated Trades, Market Fees and
```
hannesdiedrich, line 273:
```suggestion
        # The same function populates the External Trades, Accumulated Trades, Market Fees and
```

d3a_interface/sim_results/bills.py
```diff
@@ -250,13 +254,38 @@ def _accumulate_grid_fee_charged(self, area_dict, area_core_stats):
...
+        bills = self._accumulate_by_children(area_dict, flattened, {})
+        # Keep the state of the unformatted bills in order to be reused by _energy_bills method.
+        # We need this in order to read past results when calculating the new energy bills.
+        self.current_raw_bills = deepcopy(bills)
+        # Converts the children of the bills from uuids to names, because the UI uses these
+        # as headers in the table where these data are reported.
+        bills = self._swap_children_uuids_to_names(area_dict, bills)
+        # Converts the keys of the result dict from uuids to names, in order to end up in
+        # the CSV files in a human-readable format.
+        self.bills_results = self._bills_local_format(area_dict, bills)
```
hannesdiedrich, line 285: We could save this effort for the results_service, right?
If you agree, we could add a `should_export_plots` parameter to the `MarketEnergyBills` (just like in other results endpoints). What do you think?

d3a_interface/sim_results/bills.py
```diff
@@ -348,32 +377,59 @@ def _generate_external_and_total_bills(self, area_dict, results):
...
+                bills_results[child['name']] = bills_results_uuids[child['uuid']]
+        return bills_results
+
+    @classmethod
+    def _create_uuid_name_mapping(cls, area_dict, results):
+        results[area_dict["uuid"]] = area_dict["name"]
+        if key_in_dict_and_not_none(area_dict, "children"):
+            for child in area_dict["children"]:
+                results.update(cls._create_uuid_name_mapping(child, results))
+        return results
```
hannesdiedrich, line 414: Minor, but we have the `get_area_name_uuid_mapping`  in utils. When you passs `area_dict` to `serialized_scenario` it should return the same.

d3a_interface/sim_results/bills.py
```diff
@@ -348,32 +377,59 @@ def _generate_external_and_total_bills(self, area_dict, results):
...
+        area_uuid_name_mapping = self._create_uuid_name_mapping(area_dict, {})
+        final_result = {}
+        for k, children in bills_results.items():
+            children_result = {}
+            if "bought" in children:
+                final_result[k] = children
+                continue
+            for uuid, child in children.items():
+                if uuid in ["Totals", "Accumulated Trades", "External Trades", "Market Fees"] or \
+                        uuid not in area_uuid_name_mapping:
```
hannesdiedrich, line 426: This surprised my t first, but it is really only a precaution, right?

d3a_interface/sim_results/bills.py
spyrostz: I assume you are talking about the methods `_update_market_fees` and `_accumulate_grid_fee_charged`. Unfortunately they do not do the same thing, _accumulate_grid_fee_charged is actually calculating a sum of all market fees of all areas as a single number (therefore it only changes a float value), and the _update_market_fees maintains a dict with a mapping of area_uuid -> accumulated grid fees for that area for the whole simulation.
What could be improved here is that we could avoid the second recursion over the area_dict and calculate in one recursion both stats. If you agree I'll do this change.

d3a_interface/sim_results/bills.py
spyrostz: Thanks!

d3a_interface/sim_results/bills.py
spyrostz: I assume you are talking about the `uuid not in area_uuid_name_mapping`, yes it is a precaution in order to not lose entries by mistake. Only the `uuid in ["Totals", "Accumulated Trades", "External Trades", "Market Fees"]` should be sufficient

d3a_interface/sim_results/bills.py
spyrostz: Yep, why not. I'll add it, thanks

d3a_interface/sim_results/bills.py
hannesdiedrich: Ah, thanks a lot for the explanation, i looked again and was probably confused by the naming of `cumulative_fee_charged_per_market` in the `_accumulate_grid_fee_charged` which should be named rather `cumulative_fee_charged_whole_sim` or similar (but please skip this if you can not bother to open a d3a-web PR for that)
I agree with you proposal to only loop once and combine the functions. Thanks a lot!

tests/test_bills.py
```diff
@@ -0,0 +1,111 @@
...
+            load_uuid: self._empty_bills,
+            pv2_uuid: self._empty_bills,
+        }
+
+        uuid_results[grid_uuid]["house1"]["bought"] = 0.987654321
+
+        results = self.bills._bills_local_format(area_dict, uuid_results)
+        assert set(results.keys()) == {"grid", "house1", "house2", "load", "pv", "pv2"}
+        # Validate that no rounding takes place for the local results
+        assert results["grid"]["house1"]["bought"] == 0.987654321
```
hannesdiedrich, line 91: ❤️

hannesdiedrich (dismissed):
Thanks a lot for the detailed comments and unittests!
Left some refactoring comments, LGTM nevertheless

d3a_interface/sim_results/bills.py
spyrostz: I think you are right, it is better to fix this now and name it properly, otherwise more people will be confused here. Will create a d3a-web PR to change the usage of this variable

d3a_interface/sim_results/bills.py
spyrostz: The problem is that here I need the opposite, uuids -> names and not names -> uuids . However one good point is that this method could be ported to utils. If you agree let me know and I'll move it there.

hannesdiedrich: Ah, yes, sorry. Yes, please port to utils.

hannesdiedrich (dismissed):
LGTM

faizan2590 (approved):
LGTM

PR #164, opened 2021-02-15
D3ASIM-3179: Updated nested child accumulated trades or add if area a…

faizan2590:
…dded via live event.

### DIFF
--- a/d3a_interface/sim_results/cumulative_grid_trades.py
+++ b/d3a_interface/sim_results/cumulative_grid_trades.py
@@ -31,25 +31,27 @@ class CumulativeGridTrades:
         self.current_balancing_trades = {}
         self.accumulated_trades = {}
         self.accumulated_balancing_trades = {}
+        self._restored = False
 
     def update(self, area_dict, flattened_area_core_stats_dict):
         self.export_cumulative_grid_trades(
-            area_dict, flattened_area_core_stats_dict, self.accumulated_trades
+            area_dict, flattened_area_core_stats_dict, self.accumulated_trades, self._restored
         )
 
     def restore_area_results_state(self, area_uuid, last_known_state_data):
+        self._restored = True
         if area_uuid not in self.accumulated_trades:
             self.accumulated_trades[area_uuid] = last_known_state_data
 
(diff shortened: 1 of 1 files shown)


d3a_interface/sim_results/cumulative_grid_trades.py
```diff
@@ -184,14 +184,18 @@ def _update_area_children_in_accumulated_trades_dict(cls, accumulated_trades, ar
...
+            c['uuid'] for c in accumulated_trades[area['uuid']]['children']
         ]
         for child in area.get("children", []):
-            if child["name"] not in current_child_names:
+            if child["uuid"] not in current_child_uuids:
                 accumulated_trades[area['uuid']]['children'].append(
                     cls._generate_accumulated_trades_child_dict(accumulated_trades, child)
                 )
+            else:
+                child_index = current_child_uuids.index(child['uuid'])
```
Hassan754, line 196: IMO you should use ```enumerate``` , this way you have the index of the current iteration without invoking list.index() , which takes O(N) in each time called

Hassan754: Update, I see that you're not looping over the same list, so my proposal doesn't work

Hassan754 (dismissed):
LGTM

d3a_interface/sim_results/cumulative_grid_trades.py
```diff
@@ -31,25 +31,27 @@ def __init__(self):
...
+        self._restored = False
 
     def update(self, area_dict, flattened_area_core_stats_dict):
         self.export_cumulative_grid_trades(
-            area_dict, flattened_area_core_stats_dict, self.accumulated_trades
+            area_dict, flattened_area_core_stats_dict, self.accumulated_trades, self._restored
         )
 
     def restore_area_results_state(self, area_uuid, last_known_state_data):
+        self._restored = True
```
spyrostz, line 42: Please revert it to False at the end of the update() method.

spyrostz (approved):
As a general comment, instead of passing the _restored flag throughout the methods, better to remove the @classmethod decorator in the functions that you really need to. 
Once this is done LGTM

PR #167, opened 2021-02-19
Feature/d3asim 3307 Created an interface for the results classes. Added a manager class for all the simulation result objects. 

spyrostz:
Potential improvement: currently the unit tests for the results classes are located either in d3a or in d3a-web, one more improvement to the process would be to port these to the d3a-interface, so that the code is tested in the same repo that it is created.

### DIFF
--- a/d3a_interface/sim_results/area_throughput_stats.py
+++ b/d3a_interface/sim_results/area_throughput_stats.py
@@ -28,7 +28,7 @@ class AreaThroughputStats(ResultsBaseClass):
         self.imported_energy = {}
 
     def update(self, area_result_dict=None, core_stats=None, current_market_slot=None):
-        if not self._validate_update_parameters(
+        if not self._has_update_parameters(
                 area_result_dict, core_stats, current_market_slot):
             return
         self.results = {}
--- a/d3a_interface/sim_results/area_throughput_stats.py
+++ b/d3a_interface/sim_results/area_throughput_stats.py
@@ -15,22 +15,25 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+from typing import Dict
 from d3a_interface.utils import round_floats_for_ui, create_subdict_or_update
+from d3a_interface.sim_results.results_abc import ResultsBaseClass
 
 
-class AreaThroughputStats:
+class AreaThroughputStats(ResultsBaseClass):
     def __init__(self):
         self.results = {}
         self.results_redis = {}
         self.exported_energy = {}
         self.imported_energy = {}
 
-    def update(self, area_dict, core_stats, current_market_time_slot_str):
-        if current_market_time_slot_str == "":
+    def update(self, area_result_dict=None, core_stats=None, current_market_slot=None):
+        if not self._validate_update_parameters(
+                area_result_dict, core_stats, current_market_slot):
             return
         self.results = {}
         self.results_redis = {}
-        self.update_results(area_dict, core_stats, current_market_time_slot_str)
+        self.update_results(area_result_dict, core_stats, current_market_slot)
 
     def update_results(self, area_dict, core_stats, current_market_time_slot_str):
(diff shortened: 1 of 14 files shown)


d3a_interface/sim_results/export_unmatched_loads.py
```diff
@@ -17,9 +17,11 @@
 """
```
hannesdiedrich, line 17: We should really deal with D3ASIM-3324 soon!

hannesdiedrich (dismissed):
Great! Thank you! LGTM

d3a_interface/sim_results/results_abc.py
```diff
@@ -0,0 +1,34 @@
+from abc import ABC, abstractmethod
+from typing import Dict, List
+
+
+class ResultsBaseClass(ABC):
+
+    @staticmethod
+    def _validate_update_parameters(area_result_dict, core_stats, current_market_slot):
```
faizan2590, line 8:
```suggestion
    def _has_update_parameters(area_result_dict, core_stats, current_market_slot):
```
Since it expects to return boolean `True or False` rather than itself raising any exception.

faizan2590 (commented):
one comment, other than that LGTM.

d3a_interface/sim_results/results_abc.py
spyrostz: Why not, I'll adapt it, thanks!

d3a_interface/sim_results/energy_trade_profile.py
```diff
@@ -145,3 +151,44 @@ def convert_timestamp_strings_to_datetimes(profile):
...
+
+    def restore_area_results_state(self, area_uuid, last_known_state_data):
+        pass
+
+    @property
+    def plot_results(self):
+        return self.convert_timestamp_strings_to_datetimes(
+            convert_pendulum_to_str_in_dict(
+                self.traded_energy_profile, {}, ui_format=True)
+        )
```
hannesdiedrich, line 184: Sorry, I know this has been in the d3a for a while, but why do we first have to convert first to string and then to datetimes? Isn't the `plot_energy_profile` able to handle pendulum objects?

d3a_interface/sim_results/area_throughput_stats.py
```diff
@@ -87,3 +90,26 @@ def update_results(self, area_dict, core_stats, current_market_time_slot_str):
         for child in area_dict['children']:
             if child['type'] == "Area":
                 self.update_results(child, core_stats, current_market_time_slot_str)
+
+    @staticmethod
+    def merge_results_to_global(market_trade: Dict, global_trade: Dict, *_):
```
hannesdiedrich, line 95: AFAIU, `*_` is a way to ignore all additional arguments, right?
Why is this needed here?

hannesdiedrich (dismissed):
Two comments or rather questions, LGTM

d3a_interface/sim_results/area_throughput_stats.py
spyrostz: Because `slot_list` from the abc is required only in some implementations of the `merge_results_to_global` method. Therefore if I created a method with only 2 arguments, Python would complain that the abstractmethod does not adhere to the same number of arguments as the base class method.

d3a_interface/sim_results/energy_trade_profile.py
spyrostz: Good question! I would also love to remove this, let me dig a bit in d3a export.py and I'll let you know what I find

faizan2590 (dismissed):
LGTM,  though one old [comment](https://github.com/gridsingularity/d3a-interface/pull/167#discussion_r581066910) from my side to be addressed

d3a_interface/sim_results/energy_trade_profile.py
spyrostz: Thanks for the suggestion, I removed it and also added the call of the `add_sold_bought_lists` to this method too, in order to avoid the direct import of the class in the d3a code.

PR #168, opened 2021-02-23
Create helper function that returns the actual size of object in memory

Hassan754:


faizan2590 (approved):
LGTM

d3a_interface/utils.py
```diff
@@ -253,3 +256,21 @@ def iterate_over_all_modules(modules_path):
...
+
+def deep_size_of(input_obj):
+    """
+    Gets the real size of a python object taking into consideration the nested objects
+    """
+    memory_size = 0
+    ids = set()
+    objects = [input_obj]
+    while objects:
+        new = []
```
hannesdiedrich, line 269: Minor, but how about renaming `new` to something like `child_objects` ?
AFAIU, these are the child objects, e.g. an `Offer` object inside a dict tree, right?

hannesdiedrich (approved):
One minor comment. LGTM

d3a_interface/utils.py
Hassan754: @hannesdiedrich actually they are the referents objects not the direct children ,
we can name the new to referents_objs if you agree

hannesdiedrich: Thanks! yes, would be good.

PR #169, opened 2021-02-23
D3ASIM-3270: Adding cumulative net energy flow recepie.

faizan2590:


d3a_interface/utils.py
```diff
@@ -253,3 +253,17 @@ def iterate_over_all_modules(modules_path):
...
+
+
+def area_bought_from_child(trade, area_name, child_names):
+    return area_name_from_area_or_iaa_name(trade['buyer']) == area_name and \
+           area_name_from_area_or_iaa_name(trade['seller']) in child_names
+
+
+def area_sells_to_child(trade, area_name, child_names):
+    return area_name_from_area_or_iaa_name(trade['seller']) == area_name and \
+           area_name_from_area_or_iaa_name(trade['buyer']) in child_names
```
spyrostz, line 269: These are ok to have in d3a-interface if you need them here, however I would expect one more PR that removes them from d3a.

spyrostz (approved):
LGTM

hannesdiedrich (approved):
Once Spyros comment is addressed. LGTM

d3a_interface/utils.py
faizan2590: Unfortunately in d3a it uses trade object whereas here it is using trade dict

spyrostz: I see, you could potentially refactor it to handle both, by using type inspection, however it is an overkill, leave it as is. Some type hints could be potentially helpful here in order for the caller of the code to not confuse the 2 functions.

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #170, opened 2021-02-23
D3ASIM-3074: Create helper function that returns the actual size of object in memory

Hassan754:


hannesdiedrich (dismissed):
Thanks a lot! LGTM

spyrostz (commented):
One comment, about the context in which this method will be used. The reason I am asking, is because finding out how much memory an object allocates is very difficult to calculate due to references. For instance, if object A holds a reference to another object B, that was originally created by another object C, then the memory of object B will be counted on both objects A and B. Under the mindset of the normal user of this method though this is not correct, because the creator of object B is object C, and usually the user of this function would not want to include the memory of B, while calculating the memory of A. Or he might want to, but you get the idea, the usage of this function depends on the context. We could make implicit rules, by taking into account the reference count of each object, however even this depends on the context where this function will be used. 

Having said all the above, if we are talking about plain dicts or lists that contain simple types, or objects that are essentially mainly containers of data, then this will of course work and generate good results, that's why I would like to see where you actually need to use it before approving the PR. 

Finally, a comment that more or less describes the limitations of the function for the future user would be great to have here. Thanks!

Hassan754:
> One comment, about the context in which this method will be used. The reason I am asking, is because finding out how much memory an object allocates is very difficult to calculate due to references. For instance, if object A holds a reference to another object B, that was originally created by another object C, then the memory of object B will be counted on both objects A and B. Under the mindset of the normal user of this method though this is not correct, because the creator of object B is object C, and usually the user of this function would not want to include the memory of B, while calculating the memory of A. Or he might want to, but you get the idea, the usage of this function depends on the context. We could make implicit rules, by taking into account the reference count of each object, however even this depends on the context where this function will be used.
> 
> Having said all the above, if we are talking about plain dicts or lists that contain simple types, or objects that are essentially mainly containers of data, then this will of course work and generate good results, that's why I would like to see where you actually need to use it before approving the PR.
> 
> TLDR, looks good, however a comment that more or less describes the limitations of the function for the future user would be great to have here. Thanks!

@spyrostz Thanks for the informative comment !
ATM the goal of this function is to be passed to the TTLCache in [here](https://github.com/gridsingularity/d3a-web/blob/bada70b97ddd346067501cce78f8ef63f6b06ca6/simulation_results/results_calculation_helper.py#L26)
the TTLCache is a container of data (a dict), it will contain some objects (They will all be declared for the specific key, so they should not be used somewhere else )
In our case I think this is working fine, so I'll keep it if you agree and I will carry your proposal to add an explanatory comment

spyrostz:
> > One comment, about the context in which this method will be used. The reason I am asking, is because finding out how much memory an object allocates is very difficult to calculate due to references. For instance, if object A holds a reference to another object B, that was originally created by another object C, then the memory of object B will be counted on both objects A and B. Under the mindset of the normal user of this method though this is not correct, because the creator of object B is object C, and usually the user of this function would not want to include the memory of B, while calculating the memory of A. Or he might want to, but you get the idea, the usage of this function depends on the context. We could make implicit rules, by taking into account the reference count of each object, however even this depends on the context where this function will be used.
> > Having said all the above, if we are talking about plain dicts or lists that contain simple types, or objects that are essentially mainly containers of data, then this will of course work and generate good results, that's why I would like to see where you actually need to use it before approving the PR.
> > TLDR, looks good, however a comment that more or less describes the limitations of the function for the future user would be great to have here. Thanks!
> 
> @spyrostz Thanks for the informative comment !
> ATM the goal of this function is to be passed to the TTLCache in [here](https://github.com/gridsingularity/d3a-web/blob/bada70b97ddd346067501cce78f8ef63f6b06ca6/simulation_results/results_calculation_helper.py#L26)
> the TTLCache is a container of data (a dict), it will contain some objects (They will all be declared for the specific key, so they should not be used somewhere else )
> In our case I think this is working fine, so I'll keep it if you agree and I will carry your proposal to add an explanatory comment

@Hassan754 agreed, thanks for dealing with this!

spyrostz (approved):
Thanks for adding the comment ! LGTM

faizan2590 (approved):
LGTM

hannesdiedrich (approved):
LGTM

PR #174, opened 2021-02-26
D3ASIM-3307: Handled case for empty dict in the restore state methods.

spyrostz:


Hassan754 (approved):

d3a_interface/sim_results/kpi.py
```diff
@@ -225,22 +225,24 @@ def update(self, area_dict, core_stats, current_market_slot):
...
+                last_known_state_data.get('total_energy_demanded_wh', 0.0)
             self.state[area_dict['name']].total_energy_produced_wh = \
-                last_known_state_data['total_energy_produced_wh']
+                last_known_state_data.get('total_energy_produced_wh', 0.0)
             self.state[area_dict['name']].total_self_consumption_wh = \
-                last_known_state_data['total_self_consumption_wh']
+                last_known_state_data.get('total_self_consumption_wh', 0.0)
             self.state[area_dict['name']].self_consumption_buffer_wh = \
-                last_known_state_data['self_consumption_buffer_wh']
+                last_known_state_data.get('self_consumption_buffer_wh', 0.0)
```
faizan2590, line 245: I would rather keep it strict as earlier, so at least we know via log to debug later why the targeted keys' value aren't there

faizan2590 (dismissed):
One suggestion, other than that LGTM

hannesdiedrich (approved):
I agree with Faizan.
LGTM

d3a_interface/sim_results/kpi.py
spyrostz: Added these in order to handle malformed messages too, but you are right, better to see the issue immediately. Reverting this

faizan2590 (approved):
LGTM

PR #175, opened 2021-02-26
D3ASIM-3307: Fixed restoring the state of the net energy for areas th…

spyrostz:
…at do not support it.

### DIFF
--- a/d3a_interface/sim_results/cumulative_net_energy_flow.py
+++ b/d3a_interface/sim_results/cumulative_net_energy_flow.py
@@ -33,7 +33,13 @@ class CumulativeNetEnergyFlow(ResultsBaseClass):
 
     def restore_area_results_state(self, area_dict: Dict, last_known_state_data: Dict):
         area_uuid = area_dict["uuid"]
-        if area_uuid not in self.net_area_flow:
+        # TODO: Restore the state only if it is a float. At the moment, the default value for
+        # the DB column is an empty dict, which will be added for areas that do not support
+        # the net energy flow. These need to be omitted from the state restore.
+        # Needs to be refactored by changing the DB field from JSON to a float value,
+        # and denote absence of the value by leaving the field empty.
+        if area_uuid not in self.net_area_flow and \
+                isinstance(last_known_state_data, float):
             self.net_area_flow[area_uuid] = last_known_state_data
 
     def _update_results(self, area_dict, core_stats, current_market_time_slot_str):


faizan2590 (approved):
Thanks, LGTM

Hassan754:
As a side task Spyros, can you add the 'status' key in `results_mapping` in `all_results` , 
in my task https://github.com/gridsingularity/d3a-web/pull/1093, I need it , I may do the update side from the d3a web .
Anyway if you think this should be handled in a new PR, it's okay

hannesdiedrich (approved):
LGTM

spyrostz:
> As a side task Spyros, can you add the 'status' key in `results_mapping` in `all_results` ,
> in my task [gridsingularity/d3a-web#1093](https://github.com/gridsingularity/d3a-web/pull/1093), I need it , I may do the update side from the d3a web .
> Anyway if you think this should be handled in a new PR, it's okay

@Hassan754 for the simulation status I think it is better to retrieve it from the SimulationResults table, which includes all simulation results that are not dependent on time, but are denoting the current state of the simulation. The d3a-interface part only deals with aggregated and timeseries data. From the d3a-web you can easily poll it via SImulationResults.objects.filter(simulation_id=<sim_id>).status , however I am fairly certain that it is already fetched by the simulation_results Django app, so it is probably there already.

Hassan754:
> > As a side task Spyros, can you add the 'status' key in `results_mapping` in `all_results` ,
> > in my task [gridsingularity/d3a-web#1093](https://github.com/gridsingularity/d3a-web/pull/1093), I need it , I may do the update side from the d3a web .
> > Anyway if you think this should be handled in a new PR, it's okay
> 
> @Hassan754 for the simulation status I think it is better to retrieve it from the SimulationResults table, which includes all simulation results that are not dependent on time, but are denoting the current state of the simulation. The d3a-interface part only deals with aggregated and timeseries data. From the d3a-web you can easily poll it via SImulationResults.objects.filter(simulation_id=<sim_id>).status , however I am fairly certain that it is already fetched by the simulation_results Django app, so it is probably there already.

@spyrostz it is indeed there as data['status'], but this is only available for the sim_resu that triggered the event
For sims in cache. this info is not there, anyway I think that I'll query them from the db

Hassan754 (approved):

spyrostz:
> > > As a side task Spyros, can you add the 'status' key in `results_mapping` in `all_results` ,
> > > in my task [gridsingularity/d3a-web#1093](https://github.com/gridsingularity/d3a-web/pull/1093), I need it , I may do the update side from the d3a web .
> > > Anyway if you think this should be handled in a new PR, it's okay
> > 
> > 
> > @Hassan754 for the simulation status I think it is better to retrieve it from the SimulationResults table, which includes all simulation results that are not dependent on time, but are denoting the current state of the simulation. The d3a-interface part only deals with aggregated and timeseries data. From the d3a-web you can easily poll it via SImulationResults.objects.filter(simulation_id=<sim_id>).status , however I am fairly certain that it is already fetched by the simulation_results Django app, so it is probably there already.
> 
> @spyrostz it is indeed there as data['status'], but this is only available for the sim_resu that triggered the event
> For sims in cache. this info is not there, anyway I think that I'll query them from the db

@Hassan754  you should not need the status to be in the cache, nor read it from the DB IMO. When a simulation finishes or stops, you will immediately receive a new result set with the status "finished". Then you can purge immediately the results from the cache. 
Same goes for stopping a simulation btw, when a simulation is stopped a result set with the stopped status will be reported.

PR #176, opened 2021-03-01
D3ASIM-3321: Tracking area kpis using uuids.

faizan2590:


hannesdiedrich (approved):
You probably checked multiple times whether the UI is not braking right?
LGTM

spyrostz (approved):
LGTM, I also think that the UI should not break with these changes, but it is worth it to check it out before merging.

faizan2590:
> LGTM, I also think that the UI should not break with these changes, but it is worth it to check it out before merging.

Tested to validate it

https://user-images.githubusercontent.com/33482427/109816052-98dfa480-7c30-11eb-9417-dd8ff73f521e.mov

spyrostz (approved):
LGTM

PR #181, opened 2021-03-10
D3ASIM-3376: Port scenario_representation_traversal to d3a-interface

Hassan754:


fievelk (commented):
Is the `parent` ever used? Do you foresee it being used in some cases?

Hassan754:
> Is the `parent` ever used? Do you foresee it being used in some cases?

Not at this moment, but I thought this can be handy in the future

spyrostz (approved):
Great addition to decouple the recursion logic from the actual logic, LGTM

PR #183, opened 2021-03-15
D3ASIM-3383: Fixed: Only count self consumed energy when bought by th…

hannesdiedrich:
…e infinteBus in the same market-area.

### DIFF
--- a/d3a_interface/sim_results/kpi.py
+++ b/d3a_interface/sim_results/kpi.py
@@ -123,7 +123,8 @@ class KPIState:
         the InfiniteBus has consumed/demanded.
         """
         if trade['buyer_origin_id'] in self.buffer_list and \
-                trade['seller_origin_id'] in self.producer_list:
+                trade['seller_origin_id'] in self.producer_list \
+                and trade['seller_origin_id'] == trade['seller_id']:
             self.total_self_consumption_wh += trade['energy'] * 1000
             self.demanded_buffer_wh += trade['energy'] * 1000
 


d3a_interface/sim_results/kpi.py
```diff
@@ -123,7 +123,8 @@ def _dissipate_infinite_consumption(self, trade):
         the InfiniteBus has consumed/demanded.
         """
         if trade['buyer_origin_id'] in self.buffer_list and \
-                trade['seller_origin_id'] in self.producer_list:
+                trade['seller_origin_id'] in self.producer_list \
+                and trade['seller_origin_id'] == trade['seller_id']:
```
Hassan754, line 127: A side question, what is the difference between seller and seller_origin ?

Hassan754 (approved):

d3a_interface/sim_results/kpi.py
hannesdiedrich: The seller_origin is the asset that created the first offer/bid. It gets passed on to every forwarded offer/bid. If a bid/offer is forwarded, the seller is changed to the IAA as now the IAA is offering/bidding energy. 
So, imagine a PV is offering energy and this offer gets forwarded from the House market to the community market,
the offer properties are as follows in the markets.
1. House market:
   Offer(1kWh, 30cents, seller_origin="PV", seller: "PV")
2. Community market:
   Offer(1kWh, 30cents, seller_origin="PV", seller: "House")

Hassan754: Understood, Thanks!

faizan2590 (approved):
LGTM

PR #188, opened 2021-03-23
D3ASIM-3218: Adding simulation config validator.

faizan2590:


hannesdiedrich (commented):
The ideal way would be to use jsonschema for this and also test the type of the values. 
But OK for me to do it later (please add a story for that).
However, if not already started, please add this validator to the download of the simulation config file on the d3a-web side.

d3a_interface/simulation_config/__init__.py
```diff
@@ -0,0 +1,11 @@
+
+
+class SimulationConfigException(Exception):
+    pass
+
+
+def validate_simulation_config(simulation_config):
```
hannesdiedrich, line 7: In order to know which simulation_config is meant here in the future, please pre- or append something like `api_client` to the name. Also to the exception and the package name. Thanks!

d3a_interface/simulation_config/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+    if set(simulation_config.keys()) != \
+            {"username", "name", "uuid", "domain_name",
```
fievelk, line 9: Please use single quotes for strings whenever possible :) (see the "naming and style conventions" on Jira)

d3a_interface/schemas.py
```diff
@@ -14,184 +14,219 @@
...
+                                    'attributes': {'type': 'string'},
+                                    'registered': {'type': 'string'},
+                                    'aggregator_name': {'anyOf': [{'type': 'string'},
+                                                                  {'type': 'null'}]},
+                                    'children': {'anyOf': [{'type': 'array'}, {'type': 'null'}]}
+                                },
+                                'required': ['username', 'name', 'uuid', 'domain_name',
+                                             'web_socket_domain_name', 'global_settings',
+                                             'registry']
+                                }
```
hannesdiedrich, line 232: ❤️

hannesdiedrich (approved):
LGTM

d3a_interface/schemas.py
```diff
@@ -14,184 +14,219 @@
...
+                                        {'eta_seconds': {'type': 'number'},
+                                         'elapsed_time_seconds': {'type': 'number'},
+                                         'percentage_completed': {'type': 'number'}},
+                                    'slot_length_realtime': {'type': 'string'}
+                                },
+                                'registry': {
+                                    'name': {'type': 'string'},
+                                    'type': {'type': 'string'},
+                                    'uuid': {'type': 'string'},
+                                    'attributes': {'type': 'string'},
```
fievelk, line 223: Can't `attributes` be `null`? I think this is the current default: https://github.com/gridsingularity/d3a-web/blob/master/simulation/graphql/user_roles/helpers.py#L96

fievelk: Same for `registered`. I think that `aggregator_name` correctly handles this.

d3a_interface/api_simulation_config/__init__.py
```diff
@@ -0,0 +1,3 @@
+
+class ApiSimulationConfigException(Exception):
```
fievelk, line 2: This one is only used in the `d3a-api-client` so far, right? In that case, I would just put it in that repository.

d3a_interface/schemas.py
```diff
@@ -1,4 +1,4 @@
-"""
+'''
```
fievelk, line 1: I think this came from a global "find and replace" in the file. For docstrings `"""` are generally used :D Nothing big anyway of course

d3a_interface/api_simulation_config/__init__.py
faizan2590: Also on `d3a-web` side to validate the its structure before being served to user https://github.com/gridsingularity/d3a-web/pull/1282

fievelk (approved):

d3a_interface/api_simulation_config/__init__.py
fievelk: I can't seem to find where it is being used in the PR you linked :S

faizan2590: Now with this specific link to that line?https://github.com/gridsingularity/d3a-web/pull/1282/files#diff-0708084461f72c4397539a8ca5f0a90cc67058b596fb151ebff2dadb65d72f14R173

fievelk: It seems like you're pointing to a reference to the `validate_api_simulation_config()` method, but I don't see the `ApiSimulationConfigException` there.

faizan2590: It will only raise the `ValidationError` in case the returned format isn't in the expected format. Anyways, lets move that discussion to the targeted PR

PR #189, opened 2021-03-24
D3ASIM-3375: Simulation assets  list results handler

Hassan754:


d3a_interface/sim_results/simulation_assets_info.py
```diff
@@ -0,0 +1,43 @@
...
+                updated_results_dict['number_of_load_type'] += 1
+                updated_results_dict['total_energy_demand_kwh'] += area_result['total_energy_demanded_wh']/ 1000.0
+            elif 'pv_production_kWh' in area_result:
+                updated_results_dict['number_of_pv_type'] += 1
+                updated_results_dict['total_energy_generated_kwh'] += area_result['pv_production_kWh']
+            elif 'Storage' in area_result:  # Needs to update the result sent from d3a to send the capacity
+                pass
+            elif 'FinitePowerPlant' in area_result:  # Needs to update the result sent from d3a to send the power
+                pass
+
```
Hassan754, line 33: Next step is dividing the accumulated values over the number of assets

d3a_interface/sim_results/simulation_assets_info.py
```diff
@@ -0,0 +1,43 @@
...
+    def update(self, area_result_dict, core_stats, current_market_slot):
+        updated_results_dict = {field: 0 for field in SimulationAssetsInfo.FIELDS}
+        for area_uuid, area_result in core_stats.items():
+            if 'total_energy_demanded_wh' in area_result:
+                updated_results_dict['number_of_load_type'] += 1
+                updated_results_dict['total_energy_demand_kwh'] += area_result['total_energy_demanded_wh']/ 1000.0
+            elif 'pv_production_kWh' in area_result:
+                updated_results_dict['number_of_pv_type'] += 1
+                updated_results_dict['total_energy_generated_kwh'] += area_result['pv_production_kWh']
+            elif 'Storage' in area_result:  # Needs to update the result sent from d3a to send the capacity
```
hannesdiedrich, line 29: AFAIU, only the capacity of the storage is needed, which will only change with live events. The capacity is already in the configuration No need to send it again from the d3a.

d3a_interface/sim_results/simulation_assets_info.py
```diff
@@ -0,0 +1,43 @@
...
+                pass
+            elif 'FinitePowerPlant' in area_result:  # Needs to update the result sent from d3a to send the power
```
hannesdiedrich, line 31: Here, too. As it is a finite power plant, the production is known. 
However, I don't know what they want to see in this endpoint. If they want to see how much energy is sold by all power plants, we need to track that independently, indeed. However, I am strongly agains this. 
AFAIU, this endpoint is to get a rough number and not to summarize the results of all the markets.
So, we should keep the amount of work for us and for the servers low, IMO.
I see this being deprecated at some point too. 
But I might be totally wrong here and am curious about any other opinions.

d3a_interface/sim_results/simulation_assets_info.py
```diff
@@ -0,0 +1,43 @@
+from typing import Dict, List
+
+from d3a_interface.sim_results.results_abc import ResultsBaseClass
+
+
+class SimulationAssetsInfo(ResultsBaseClass):
+    FIELDS = ['number_of_house_type', 'avg_assets_per_house', 'number_of_load_type',
```
hannesdiedrich, line 7: Fields like `avg_assets_per_house` or `number_of_storage_type` could be acquired solely from the configuration and thus only on the d3a-web side.

hannesdiedrich (commented):
left some general comments, hope that I dont start an endless discussion.

d3a_interface/sim_results/simulation_assets_info.py
spyrostz: IIRC latest configuration is updated immediately once the live event is created and sent to d3a, and not when it is really applied to d3a. This means that the latest configuration contains values that have not affected the simulation yet. For normal simulations this is not a problem, because the slots are executed fast enough for the user to not notice. However, for CNs, this process might take 15 mins, therefore the user will be able to see the updated capacity due to a live event that has not affected the simulation yet, thus being led to wrong conclusions. 

Having said that, I agree that capacity values could potentially be populated with data from the DB, and it is not necessary that we send the capacity values from d3a. This part of the code is mainly called by the simulation_results service, which has access to the DB, therefore it is feasible to read the capacity values directly from the DB. Also, since this function is triggered after the market slot has finished, we will not fall into the pitfall that I explained in the first paragraph. 

Of course, the interface of the `def update(self, area_result_dict, core_stats, current_market_slot):` method will need to be adapted to also incorporate one more argument for the latest representation from the ConfigurationSettings, however it is not that hard to adapt. An alternative approach would be to create a new method that will calculate only values based on the ConfigurationSettings (we could also include it in the abstractmethods in case it is useful for other simulation results in the future, but I would avoid it for now). 

All of the above apply to the rest of the comments below. I also hope that I do not start an endless discussion with this comment :(

hannesdiedrich: OK, understood and convinced with `However, for CNs, this process might take 15 mins, therefore the user will be able to see the updated capacity due to a live event that has not affected the simulation yet, thus being led to wrong conclusions.`
I also would vote for `An alternative approach would be to create a new method that will calculate only values based on the ConfigurationSettings (we could also include it in the abstractmethods in case it is useful for other simulation results in the future, but I would avoid it for now).`
Yeah, lets not drag this endlessly here. :-)

Hassan754: @spyrostz how do you suggest adding the extra config_obj parameter to the update interface of my new handler class without adding it to the all_results's ResultHandler update method ? AFAIS this function invokes other update methods

d3a_interface/sim_results/simulation_assets_info.py
```diff
@@ -0,0 +1,68 @@
...
+              'max_power_plant_power_kw']
+
+    def __init__(self):
+        self.assets_info = {field: 0 for field in SimulationAssetsInfo.FIELDS}
+
+    @staticmethod
+    def merge_results_to_global(market_results: Dict, global_results: Dict, slot_list: List):
+        pass
+
+    def update(self, area_result_dict, core_stats, current_market_slot):
```
faizan2590, line 21:
```suggestion
    def update(self, area_result_dict: Dict, core_stats: Dict, current_market_slot: str):
```
Since parameter type hint is already added in other methods of this module.

d3a_interface/sim_results/simulation_assets_info.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def restore_area_results_state(self, area_dict: Dict, last_known_state_data: Dict):
+        pass
+
+    @property
+    def raw_results(self):
+        return self.assets_info
+
+    def memory_allocation_size_kb(self):
+        return 0
```
faizan2590, line 68:
```suggestion
        return self._calculate_memory_allocated_by_objects([self.assets_info])
```

d3a_interface/sim_results/all_results.py
```diff
@@ -60,6 +62,14 @@ def update(self, area_dict: Dict, core_stats: Dict, current_market_slot: str):
             v.update(area_dict, core_stats, current_market_slot)
         self._update_memory_utilization()
 
+    def update_from_repr(self, area_representation: Dict):
+        """
+        Can be added as an abstract method to the results_abc if needed
+        """
+        for key, value in self.results_mapping.items():
+            if hasattr(value, 'update_from_repr'):
```
spyrostz, line 70: Rather than doing this, I would create a base class method that does nothing by default, and only override it in subclasses that you actually need it. That way you can avoid using introspection / hasattr.

d3a_interface/sim_results/simulation_assets_info.py
hannesdiedrich: @Hassan754 Sorry, I can not find this in your d3a-web PR. Can you maybe point me to it?

d3a_interface/sim_results/simulation_assets_info.py
```diff
@@ -0,0 +1,68 @@
+from typing import Dict, List
+
+from d3a_interface.sim_results import is_finite_power_plant_node_type, is_prosumer_node_type
+from d3a_interface.sim_results.results_abc import ResultsBaseClass
+from d3a_interface.utils import scenario_representation_traversal
+
+
+class SimulationAssetsInfo(ResultsBaseClass):
```
hannesdiedrich, line 8: Please unittest at least `update` and `update_from_repr`. Thanks!

hannesdiedrich (commented):
Two comments, will review once addressed.

d3a_interface/sim_results/all_results.py
spyrostz: This still needs to be dealt with

d3a_interface/sim_results/all_results.py
```diff
@@ -60,6 +62,14 @@ def update(self, area_dict: Dict, core_stats: Dict, current_market_slot: str):
             v.update(area_dict, core_stats, current_market_slot)
         self._update_memory_utilization()
 
+    def update_from_repr(self, area_representation: Dict):
+        """
+        Can be added as an abstract method to the results_abc if needed
+        """
+        for key, value in self.results_mapping.items():
```
spyrostz, line 69: Still we do not need the key here, how about:
```suggestion
        for result_object in self.results_mapping.values():
```
Also would be great if you did the same refactoring for the `update()` method right above

spyrostz (approved):
2 comments, will LGTM once these and Hannes' comments are dealt with

d3a_interface/sim_results/simulation_assets_info.py
```diff
@@ -0,0 +1,75 @@
...
+                updated_results_dict['total_energy_capacity_kwh'] +=\
+                    area_dict.get('battery_capacity_kWh', 0)
+            elif is_finite_power_plant_node_type(area_dict):
+                updated_results_dict['number_of_power_plant_type'] += 1
+                updated_results_dict['max_power_plant_power_kw'] +=\
+                    area_dict.get('max_available_power_kW', 0)
+
+        updated_results_dict['number_of_house_type'],\
+            updated_results_dict['avg_assets_per_house'] = \
+            HomeRepresentationUtils.calculate_home_area_stats_from_repr_dict(area_representation)
```
hannesdiedrich, line 63: AFAIS, `calculate_home_area_stats_from_repr_dict` is not returning the number of homes in the first element of the tuple. I think you need to count then separately with the check `is_home_area`

hannesdiedrich (approved):
LGTM

d3a_interface/sim_results/results_abc.py
```diff
@@ -35,6 +35,13 @@ def update(self, area_result_dict, core_stats, current_market_slot):
         """
         pass
 
+    @abstractmethod
```
spyrostz, line 38: I would recommend to not make it an abstractmethod, and this will remove the need to override the method in the subclasses. If/when more subclasses need it, we'll convert this method to abstract then.

spyrostz (approved):

hannesdiedrich (approved):
LGTM

d3a_interface/sim_results/__init__.py
```diff
@@ -30,11 +30,11 @@ def is_bulk_power_producer(area):
 def is_pv_node_type(area):
     return area['type'] in ["PVStrategy", "PVUserProfileStrategy", "PVPredefinedStrategy",
                             "PVExternalStrategy", "PVUserProfileExternalStrategy",
-                            "PVPredefinedExternalStrategy", "PVForecastExternalStrategy"]
+                            "PVPredefinedExternalStrategy", "PVForecastExternalStrategy", 'PV']
 
 
 def is_finite_power_plant_node_type(area):
-    return area['type'] == "FinitePowerPlant"
+    return area['type'] in ['FinitePowerPlant', 'FiniteDieselGenerator']
```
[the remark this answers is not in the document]
fievelk: Actually, I think the same can be applied to many functions in this module (not required for this PR anyway)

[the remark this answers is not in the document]
Hassan754: I see your concern, however for your suggested solution we'll keep the objects in the memory of the service.
They're not that heavy though, but neither is instantiating them imo

spyrostz (approved):
LGTM

PR #190, opened 2021-03-24
D3ASIM-3370: Added key_in_dict_and_is_true.

hannesdiedrich:


d3a_interface/utils.py
```diff
@@ -151,6 +151,10 @@ def key_in_dict_and_not_none(d, key):
     return key in d and d[key] is not None
 
 
+def key_in_dict_and_is_true(d, key):
+    return key in d and d[key] and d[key] is True
```
faizan2590, line 155:
```suggestion
    return key in d and d[key] is True
```

hannesdiedrich: upsi, thanks!

faizan2590 (approved):
LGTM

Hassan754 (approved):

d3a_interface/utils.py
fievelk: To be sure: this is checking just for the `True` values, right? It needs to fail if the boolean value evaluates to true, but it's not the `True` object?

fievelk: More details about the question I asked: [PEP8 - Programming recommendations](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)
```
# Wrong:
if greeting is True:
```

d3a_interface/utils.py
```diff
@@ -151,6 +151,10 @@ def key_in_dict_and_not_none(d, key):
     return key in d and d[key] is not None
 
 
+def key_in_dict_and_is_true(d, key):
+    return key in d and d[key] is True
```
fievelk, line 155: If we only need to check for identity with the `True` boolean object, would this work for you?
```suggestion
    return d.get(key) is True
```
At this point, I think the need for a separate method in the d3a-interface is not so big (since it's shorter than the method's name). What do you think?

hannesdiedrich: Ok, considering your former comment, it would boil down to 
```
return d.get(key, False)
```
Right?
Then I guess indeed, the comment from @faizan2590  here https://github.com/gridsingularity/d3a-web/pull/1267#discussion_r599977021 is not really needed, right?

Hassan754: Sorry for jumping here, but `d.get(key)` will throw an exception if the key doesn't exist
we can add a default value `d.get(key, False)`

hannesdiedrich: @Hassan754 https://github.com/gridsingularity/d3a-interface/pull/190#discussion_r601615940

fievelk: `d.get(key)` is the same as `d.get(key, None)`, so it shouldn't raise an exception unless I got confused (which happens often when I talk about `.get()`) :D

fievelk: @Hannes let's just keep in mind that `d.get(key)` is different than `d.get(key) is True`:
```python
d = {'foo': 'a string'}

if d.get('foo'):
    print('true')
else:
    print('false')
# true

if d.get('foo') is True:
    print('true')
else:
    print('false')
# false
```

hannesdiedrich: OK, lets get this over with here. What is is your proposal @fievelk ?
- calling `d.get(key) is True` on the d3a-web side directly
- keeping this here in this function to be reused

fievelk: I would vote for the first option:
> calling d.get(key) is True on the d3a-web side directly
It's a one-liner, it's clear enough and it doesn't require using an external library (like d3a-interface)

faizan2590: Considering @fievelk's suggestion none of dict's key member checking is needed in `d3a-interface` as I understood:

- `key_in_dict_and_not_none` could directly be used via `if d.get(key) is not None`
- `key_in_dict_and_not_none_and_not_str_type` could directly be used via `if isinstance(d.get(key), str)`
- `key_in_dict_and_not_none_and_list_type` could directly be used via `if isinstance(d.get(key), list)`
- `key_in_dict_and_not_none_and_list_type` could directly be used via `if isinstance(d.get(key), dict)`

fievelk: My personal preference would be not to define them, given the pros-cons. If they are being used often on the same attributes, it might mean that those checks can probably be grouped into some kind of Validator class rather than multiple small helper functions.

PR #192, opened 2021-03-29
D3ASIM-3366: Added the home statistics calculation.

spyrostz:


d3a_interface/utils.py
```diff
@@ -427,3 +428,28 @@ def scenario_representation_traversal(sc_repr, parent=None):
...
 
     yield sc_repr, parent
+
+
+class HomeRepresentationUtils:
+    @staticmethod
+    def _is_home(representation):
+        home_devices = ["PV", "Storage", "Load", "MarketMaker"]
+        return all("type" in c and c["type"] in home_devices
+                   for c in representation["children"])
```
Hassan754, line 438: we should check if `children` key exists in here instead of making sure before calling it

d3a_interface/utils.py
```diff
@@ -427,3 +428,28 @@ def scenario_representation_traversal(sc_repr, parent=None):
...
+
+    @classmethod
+    def is_home_area(cls, representation):
+        is_market_area = not key_in_dict_and_not_none(representation, "type") or representation["type"] == "Area"
```
Hassan754: What if the 'type' key doesn't exists ? this will pass the first check and then it will raise an exception in the `representation["type"] == "Area"`
```suggestion
        is_market_area = representation.get("type", True) == "Area"
```

d3a_interface/utils.py
```diff
@@ -427,3 +428,28 @@ def scenario_representation_traversal(sc_repr, parent=None):
...
+        has_home_devices = key_in_dict_and_not_none(representation, "children") and representation["children"] and \
```
Hassan754: It is safe to omit the `and representation["children"]` check imo

d3a_interface/utils.py
```diff
@@ -427,3 +428,28 @@ def scenario_representation_traversal(sc_repr, parent=None):
...
+        return is_market_area and has_home_devices
+
+    @classmethod
+    def calculate_home_area_stats_from_repr_dict(cls, representation):
+        devices_per_home = [
+            len(area["children"])
+            for area, _ in scenario_representation_traversal(representation)
+            if cls.is_home_area(area)
+        ]
+        return len(devices_per_home), mean(devices_per_home)
```
Hassan754, line 456: Thanks a lot!

Hassan754 (approved):
LGTM

d3a_interface/utils.py
spyrostz: Safe but not optimal, if I omit it the _is_area will still be called with an empty list. However the impact is negligible, thus if you find it easier to read then let me know to remove it.

Hassan754: It's okay then :)

d3a_interface/utils.py
spyrostz: Actually it is working as it is, have a look at the unit tests where an Area type is not explicitly defined. The case of "type" key missing is covered by the `not key_in_dict_and_not_none(representation, "type")` statement, which will immediately return True if there is no type key.

d3a_interface/utils.py
spyrostz: It is already checked in the is_home_area method, however I could move the checks in the _is_home method. Doing it

hannesdiedrich (approved):
LGTM

Hassan754 (approved):

PR #196, opened 2021-04-20
D3ASIM-2540: Add exception handling to profile parsing

Hassan754:


d3a_interface/read_user_profile.py
```diff
@@ -388,6 +388,9 @@ def read_profile_without_config(input_profile: Dict, slot_length_mins=15) -> Dic
 
 
 def read_and_convert_identity_profile_to_float(profile):
-    parsed_profile = ast.literal_eval(str(profile))
+    try:
+        parsed_profile = ast.literal_eval(str(profile))
```
faizan2590, line 392:
```suggestion
        parsed_profile = ast.literal_eval(profile)
```
I think explicit casting to str for literal_eval can be avoided here as the later is covered in exception raised as `ValueError `

d3a_interface/read_user_profile.py
```diff
@@ -388,6 +388,9 @@ def read_profile_without_config(input_profile: Dict, slot_length_mins=15) -> Dic
 
 
 def read_and_convert_identity_profile_to_float(profile):
-    parsed_profile = ast.literal_eval(str(profile))
+    try:
+        parsed_profile = ast.literal_eval(profile)
+    except ValueError:
+        parsed_profile = profile
```
hannesdiedrich, line 394: Hmm, I don't understand. What is the issue here? Chris is passing a filename string right?

spyrostz: According to the ticket this is the problem, and I think that changing the interface library is the wrong way to go. I think that if the infinite_bus.py used the `read_arbitrary_profile` method, then the input would have been detected that it is a filename and not a dict profile.

PR #201, opened 2021-04-29
D3ASIM-3450: remove validation of area names

fievelk:
The simulation results that can be downloaded by the users contain a directory structure based on the names of the areas in the scenario. If these names contain invalid characters (like `/`) the export process will either fail or be compromised.

Our previous solution was to restrict the allowed characters in area names. However, this created incompatibilities with several simulations that had already been created by users in the past. We therefore decided to allow those characters again, but to sanitize them when generating the paths for the downloaded results (see https://github.com/gridsingularity/d3a-web/pull/1326)


hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

faizan2590 (approved):
LGTM

PR #206, opened 2021-05-17
D3ASIM-3447: Port client connections interfaces to d3a-interface

Hassan754:


d3a_interface/client_connections/websocket_connection.py
```diff
@@ -0,0 +1,84 @@
...
+from d3a_interface.client_connections.utils import retrieve_jwt_key_from_server
+
+
+class WebsocketMessageReceiver(ABC):
+    def __init__(self, rest_client):
+        self.client = rest_client
+
+    @abstractmethod
+    def received_message(self, message):
+        ...
```
spyrostz, line 24: Out of curiosity, is this preferred over the old-fashioned `pass` statement ?

spyrostz (approved):
LGTM, probably needs to be merged together with the api client PR

hannesdiedrich (approved):
LGTM

d3a_interface/client_connections/websocket_connection.py
Hassan754: It looks it is only a matter of preference, there are no rules about this in the PEPs

faizan2590 (approved):

d3a_interface/client_connections/websocket_connection.py
fievelk: If that's ok for you, I would then use `pass` (as it's used all across our codebase and it's more common I think).

Hassan754: @fievelk Yea I'm okay with both options, I'll leave this here though if it's fine

PR #210, opened 2021-05-25
D3ASIM-3479: Changed the sign of cumulativeNetEnergyFlow to be consis…

hannesdiedrich:
…tent with the EnergyTradeProfile.

This has to be tested by an e2e-test. Gave up thinking about a way of testing the consistency between cumulativeNetEnergyFlow and EnergyTradeProfile.

### DIFF
--- a/d3a_interface/sim_results/cumulative_net_energy_flow.py
+++ b/d3a_interface/sim_results/cumulative_net_energy_flow.py
@@ -51,13 +51,14 @@ class CumulativeNetEnergyFlow(ResultsBaseClass):
 
         child_names = [c['name'] for c in area_dict['children']]
         for trade in core_stats.get(area_dict['uuid'], {}).get('trades', []):
+            # export
             if area_bought_from_child(trade, area_dict['name'], child_names):
-                add_or_create_key(
+                subtract_or_create_key(
                     self.net_area_flow, area_dict['uuid'], trade['energy']
                 )
             # import
             if area_sells_to_child(trade, area_dict['name'], child_names):
-                subtract_or_create_key(
+                add_or_create_key(
                     self.net_area_flow, area_dict['uuid'], trade['energy']
                 )
 


spyrostz (approved):
LGTM, but unfortunately this will affect only new simulation runs, otherwise we need a migration to revert the sign of the field in the DB. Would be great to clarify it with frontend / Sarah.

hannesdiedrich:
> LGTM, but unfortunately this will affect only new simulation runs, otherwise we need a migration to revert the sign of the field in the DB. Would be great to clarify it with frontend / Sarah.

Actually you are right, I save myself the hassle and dont ask, because this numbers are shown in the singularity map (as coloured circles). Will add a migration right away.

hannesdiedrich:
> > LGTM, but unfortunately this will affect only new simulation runs, otherwise we need a migration to revert the sign of the field in the DB. Would be great to clarify it with frontend / Sarah.
> 
> Actually you are right, I save myself the hassle and dont ask, because this numbers are shown in the singularity map (as coloured circles). Will add a migration right away.

@spyrostz This is not straightforward (as you probably already know). 
I had a look into the d3a-web code now and found the `_populate_null_cumulative_net_energy_flow_entry`, but this is only run when the `cumulative_net_energy_flow` was not populated yet. I can not think of a (resource inexpensive) way to run these online migrations again for the sign change. 
If you do not have a better idea, I will try to explain this to Sarah and that they should live with this or deal with it on the frontend (for that historical runs).

spyrostz:
> > > LGTM, but unfortunately this will affect only new simulation runs, otherwise we need a migration to revert the sign of the field in the DB. Would be great to clarify it with frontend / Sarah.
> > 
> > 
> > Actually you are right, I save myself the hassle and dont ask, because this numbers are shown in the singularity map (as coloured circles). Will add a migration right away.
> 
> @spyrostz This is not straightforward (as you probably already know).
> I had a look into the d3a-web code now and found the `_populate_null_cumulative_net_energy_flow_entry`, but this is only run when the `cumulative_net_energy_flow` was not populated yet. I can not think of a (resource inexpensive) way to run these online migrations again for the sign change.
> If you do not have a better idea, I will try to explain this to Sarah and that they should live with this or deal with it on the frontend (for that historical runs).

That's why I mentioned that we should make sure that this needs to be done before implementing it actually. In order to achieve it it is probably a better idea to create a new migration that does not recalculate the value, but only flips the sign of the existing value. Not pretty, better to be avoided unless necessary.
However, how about instead of changing it in the database, to change the sign convention in the API? That way you will be able to easily switch the sign from positive to negative, and also adapt it if the UI uses a different sign convention in the future.

hannesdiedrich:
> > > > LGTM, but unfortunately this will affect only new simulation runs, otherwise we need a migration to revert the sign of the field in the DB. Would be great to clarify it with frontend / Sarah.
> > > 
> > > 
> > > Actually you are right, I save myself the hassle and dont ask, because this numbers are shown in the singularity map (as coloured circles). Will add a migration right away.
> > 
> > 
> > @spyrostz This is not straightforward (as you probably already know).
> > I had a look into the d3a-web code now and found the `_populate_null_cumulative_net_energy_flow_entry`, but this is only run when the `cumulative_net_energy_flow` was not populated yet. I can not think of a (resource inexpensive) way to run these online migrations again for the sign change.
> > If you do not have a better idea, I will try to explain this to Sarah and that they should live with this or deal with it on the frontend (for that historical runs).
> 
> That's why I mentioned that we should make sure that this needs to be done before implementing it actually. In order to achieve it it is probably a better idea to create a new migration that does not recalculate the value, but only flips the sign of the existing value. Not pretty, better to be avoided unless necessary.
> However, how about instead of changing it in the database, to change the sign convention in the API? That way you will be able to easily switch the sign from positive to negative, and also easily adapt it if the UI uses a different sign convention in the future.

Thanks a lot! I agree and have created the corresponding d3a-web PR: https://github.com/gridsingularity/d3a-web/pull/1337

Closing this PR now

PR #212, opened 2021-05-26
D3ASIM-3453: Added Kafka communication package.

spyrostz:


d3a_interface/kafka_communication/kafka_consumer.py
```diff
@@ -0,0 +1,36 @@
...
+                      "sasl_plain_password": KAFKA_PASSWORD,
+                      "security_protocol": KAFKA_COMMUNICATION_SECURITY_PROTOCOL,
+                      "ssl_context": create_kafka_new_ssl_context(),
+                      "sasl_mechanism": KAFKA_SASL_AUTH_MECHANISM,
+                      "api_version": KAFKA_API_VERSION,
+                      "fetch_max_bytes": KAFKA_MAX_MESSAGE_SIZE_PER_TOPIC,
+                      "max_partition_fetch_bytes": KAFKA_MAX_MESSAGE_SIZE_PER_TOPIC,
+                      "max_poll_records": KAFKA_MAX_POLL_RECORDS}
+        else:
+            kwargs = {"bootstrap_servers": DEFAULT_KAFKA_URL}
```
Hassan754, line 28: It looks that the default value for this is the same as provided, however we do may need to change the default url in the future so let's keep it as it is

d3a_interface/kafka_communication/kafka_producer.py
```diff
@@ -0,0 +1,60 @@
...
+KAFKA_PUBLISH_RETRIES = 5
+KAFKA_BUFFER_MEMORY = 2048000000
+KAFKA_MAX_REQUEST_SIZE = 2048000000
+
+
+def kafka_connection_factory():
+    try:
+        return KafkaConnection()
+    except Exception:
+        logging.info("Running without Kafka connection for simulation results.")
```
Hassan754, line 19: When is this supposed to happen ?

d3a_interface/kafka_communication/kafka_producer.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def publish(self, endpoint_buffer):
+        pass
+
+    @staticmethod
+    def is_enabled():
+        return False
+
+
+class KafkaConnection:
```
Hassan754, line 35: To make sure we have the same interfaces here do you think it's okay if we extend the `DisabledKafkaConnection` class here ?

d3a_interface/kafka_communication/kafka_producer.py
```diff
@@ -0,0 +1,60 @@
+import logging
+import json
+from zlib import compress
+from kafka import KafkaProducer
+
+from d3a_interface.kafka_communication import KAFKA_URL, DEFAULT_KAFKA_URL, KAFKA_USERNAME, \
+    KAFKA_PASSWORD, KAFKA_COMMUNICATION_SECURITY_PROTOCOL, KAFKA_SASL_AUTH_MECHANISM, \
+    KAFKA_API_VERSION, create_kafka_new_ssl_context, KAFKA_TOPIC
```
Hassan754, line 8: nit. , We can use parantheses in all 2+ lines imports

d3a_interface/kafka_communication/kafka_producer.py
```diff
@@ -0,0 +1,60 @@
...
+                      "retries": KAFKA_PUBLISH_RETRIES,
+                      "buffer_memory": KAFKA_BUFFER_MEMORY,
+                      "max_request_size": KAFKA_MAX_REQUEST_SIZE}
+        else:
+            kwargs = {"bootstrap_servers": KAFKA_URL}
+
+        self.producer = KafkaProducer(**kwargs)
+
+    def publish(self, results, job_id):
+        results = json.dumps(results).encode("utf-8")
```
Hassan754, line 54: I'm not sure what the default encoding `json.dumps` returns but as for the .encode(), we can omit passing the "utf-8" parameter as it's the default

d3a_interface/kafka_communication/kafka_producer.py
```diff
@@ -0,0 +1,60 @@
...
+import json
+from zlib import compress
+from kafka import KafkaProducer
+
+from d3a_interface.kafka_communication import KAFKA_URL, DEFAULT_KAFKA_URL, KAFKA_USERNAME, \
+    KAFKA_PASSWORD, KAFKA_COMMUNICATION_SECURITY_PROTOCOL, KAFKA_SASL_AUTH_MECHANISM, \
+    KAFKA_API_VERSION, create_kafka_new_ssl_context, KAFKA_TOPIC
+
+KAFKA_PUBLISH_RETRIES = 5
+KAFKA_BUFFER_MEMORY = 2048000000
```
Hassan754, line 11: If possible add the units here

d3a_interface/kafka_communication/kafka_consumer.py
```diff
@@ -0,0 +1,36 @@
...
+
+        self._consumer = KafkaConsumer(KAFKA_TOPIC, **kwargs)
```
Hassan754, line 30: Won't we need to create several kafka consumers with different topics in the future ?

Hassan754 (dismissed):
LGTM

hannesdiedrich (commented):
LGTM

d3a_interface/kafka_communication/kafka_producer.py
spyrostz: Typically in 2 circumstances:
- when running d3a using the CLI 
- when running integration tests 
During these 2 usecases, we do not need Kafka at all. Therefore the factory method is responsible for constructing the DisabledKafkaConnection class.

d3a_interface/kafka_communication/kafka_producer.py
spyrostz: Sorry for not following this but I would rather be explicit here and state the encoding type. Reason for that was that in the former implementation ascii encoding was used, which should be avoided, and I believe that stating the encoding explicitly will deter people from changing it in the future.

d3a_interface/kafka_communication/kafka_producer.py
spyrostz: Agreed, thanks!

d3a_interface/kafka_communication/kafka_producer.py
spyrostz: It's clear that it is bytes, however I'll add it

d3a_interface/kafka_communication/kafka_consumer.py
spyrostz: YAGNI. When the time comes it will be trivial to modularize it to accept multiple topics.

d3a_interface/kafka_communication/kafka_producer.py
spyrostz: Well by only inheriting from the class we do not ensure that we have the same interfaces, however I understand your point and I agree that it can be achieved by converting the DisabledKafkaConnection to an abstract base class. Let me try that, thanks!

d3a_interface/kafka_communication/kafka_producer.py
Hassan754: Understood, thanks!

fievelk (approved):

d3a_interface/kafka_communication/kafka_producer.py
spyrostz: I unfortunately had to revert this fix because I forgot that one cannot instantiate an ABC, there needs to be a third class that inherits from the ABC. In order to not have 3 classes I think it is simpler to not define an ABC at all, at least until the interface becomes more complex that justifies the extra class for the interface.

PR #216, opened 2021-06-01
D3ASIM-3353: implement Home Meter device

fievelk:
Implement features for the new Home Meter device that should be used by both d3a and d3a-web.

# Reference

- gridsingularity/d3a#1159

### DIFF
--- a/d3a_interface/device_validator.py
+++ b/d3a_interface/device_validator.py
@@ -160,8 +160,8 @@ class HomeMeterValidator(DeviceValidator):
             raise D3ADeviceException(
                 {"misconfiguration": [
                     "initial_selling_rate/market_maker_rate should be greater than or equal to "
-                    "final_selling_rate. Please adapt the market_maker_rate of the configuration or "
-                    "the initial_selling_rate"]})
+                    "final_selling_rate. Please adapt the market_maker_rate of the configuration "
+                    "or the initial_selling_rate"]})
 
         if (
                 kwargs.get("fit_to_limit") is True
@@ -601,4 +601,3 @@ def validate_finite_diesel_generator(**kwargs):
 def validate_range_limit(initial_limit, value, final_limit, error_message):
     if not initial_limit <= value <= final_limit:
         raise D3ADeviceException(error_message)
-
--- a/d3a_interface/device_validator.py
+++ b/d3a_interface/device_validator.py
@@ -97,7 +97,7 @@ class HomeMeterValidator(DeviceValidator):
                 "misconfiguration": [
                     "initial_buying_rate should be less than or equal to final_buying_rate/"
                     "market_maker_rate. Please adapt the market_maker_rate of the configuration "
-                    "or the initial_buying_rate"]})
+                    "or the initial_buying_rate."]})
 
         if kwargs.get("energy_rate_increase_per_update") is not None:
             error_message = {
@@ -151,7 +151,7 @@ class HomeMeterValidator(DeviceValidator):
                 {"misconfiguration": [
                     "initial_selling_rate/market_maker_rate should be greater than or equal to "
                     "final_selling_rate. Please adapt the market_maker_rate of the configuration "
-                    "or the initial_selling_rate"]})
+                    "or the initial_selling_rate."]})
 
         if (kwargs.get("fit_to_limit") is True
                 and kwargs.get("energy_rate_decrease_per_update") is not None):
@@ -243,11 +243,10 @@ def validate_load_device_price(**kwargs):
     if ("initial_buying_rate" in kwargs and kwargs["initial_buying_rate"] is not None) and \
             ("final_buying_rate" in kwargs and kwargs["final_buying_rate"] is not None) and \
             (kwargs["initial_buying_rate"] > kwargs["final_buying_rate"]):
-        raise D3ADeviceException({"misconfiguration": [f"initial_buying_rate should be "
-                                                       f"less than or equal to final_buying_rate/"
-                                                       f"market_maker_rate. Please adapt the "
-                                                       f"market_maker_rate of the configuration "
-                                                       f"or the initial_buying_rate"]})
+        raise D3ADeviceException({"misconfiguration": [
+            "initial_buying_rate should be less than or equal to final_buying_rate/"
+            "market_maker_rate. Please adapt the market_maker_rate of the configuration "
+            "or the initial_buying_rate."]})
     if "energy_rate_increase_per_update" in kwargs and \
             kwargs["energy_rate_increase_per_update"] is not None:
         error_message = \
@@ -317,10 +316,10 @@ def validate_pv_device_price(**kwargs):
             ("final_selling_rate" in kwargs and kwargs["final_selling_rate"] is not None) and \
             (kwargs["initial_selling_rate"] < kwargs["final_selling_rate"]):
         raise D3ADeviceException(
-            {"misconfiguration": [f"initial_selling_rate/market_maker_rate should be greater "
-                                  f"than or equal to final_selling_rate. Please adapt the "
-                                  f"market_maker_rate of the configuration or the "
-                                  f"initial_selling_rate"]})
+            {"misconfiguration": [
+                "initial_selling_rate/market_maker_rate should be greater than or equal to "
+                "final_selling_rate. Please adapt the market_maker_rate of the configuration or "
+                "the initial_selling_rate."]})
     if ("fit_to_limit" in kwargs and kwargs["fit_to_limit"] is True) and \
             ("energy_rate_decrease_per_update" in kwargs and
              kwargs["energy_rate_decrease_per_update"] is not None):
--- a/d3a_interface/device_validator.py
+++ b/d3a_interface/device_validator.py
@@ -16,6 +16,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import ast
+from abc import ABCMeta, abstractmethod
 
 from d3a_interface.constants_limits import ConstSettings
 from d3a_interface.exceptions import D3ADeviceException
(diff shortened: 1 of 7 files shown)


d3a_interface/schemas.py
```diff
@@ -121,6 +122,27 @@ class ScenarioSchemas:
                                                      {'type': 'string'}]}
                 }
             },
+            'home_meter': {
+                'type': 'object',
+                'properties': {
```
hannesdiedrich, line 127: I dont see `home_meter_profile` and `home_meter_profile_uuid` in here. Is this wanted?

d3a_interface/device_validator.py
```diff
@@ -24,10 +25,164 @@
...
+                kwargs.get("initial_selling_rate") is not None
+                and kwargs.get("final_selling_rate") is not None
+                and kwargs["initial_selling_rate"] < kwargs["final_selling_rate"]):
+            raise D3ADeviceException(
+                {"misconfiguration": [
+                    "initial_selling_rate/market_maker_rate should be greater than or equal to "
+                    "final_selling_rate. Please adapt the market_maker_rate of the configuration or "
+                    "the initial_selling_rate"]})
+
+        if (
```
hannesdiedrich, line 166: Please also adapt https://github.com/gridsingularity/d3a-web/pull/1340#discussion_r645450872 here. Thanks!

hannesdiedrich (commented):
One minor and one important comment. LGTM once addressed

d3a_interface/schemas.py
fievelk: Good catch. I'm actually wondering though: the profile uuid is also not present in the other devices (with the exception of the Load). The validations don't fail because we don't declare any attribute as required (except `name`) but maybe this should be changed to make the validations stricter.

hannesdiedrich: Yeah, I think this was done on purpose (not expecting all values to be present) because the Class API on the d3a side is not expecting these values either. AFAIR, these schemas where introduced to make sure that the right parameter types are sent to the d3a. But yes, in the future we could be more strickt.

d3a_interface/sim_results/device_statistics.py
```diff
@@ -136,46 +136,36 @@ def calculate_stats_for_infinite_bus(cls, area_dict, subdict, core_stats, curren
...
+    def _compute_key_name_for_node(area: Dict):
+        if is_pv_node_type(area):
+            return "pv_production_kWh"
+        if is_prosumer_node_type(area):
+            return "soc_history_%"
+        if is_load_node_type(area):
+            return "load_profile_kWh"
+        if area["type"] == "HomeMeterStrategy":
+            return "home_meter_profile_kWh"
+        if area["type"] == "FinitePowerPlant":
```
Hassan754, line 149: There is a function that checks this `is_finite_power_plant_node_type` in `d3a_interface/sim_results/__init__.py:36`
I suggest adding a similar function for the HomeMeter + adding it in the `get_unified_area_type` function residing in the same file

Hassan754: Plus, this function can be replaced with a mapping dict of `unified_area_type` as keys and corresponding `key_name_for_node` as values
This way the usage would be 
```python
key_name = mapping_dict[area_dict["type"]]
```
However if you think that in the future we need to add more logic here, feel free to keep it

d3a_interface/device_validator.py
```diff
@@ -24,10 +25,159 @@
...
 StorageSettings = ConstSettings.StorageSettings
 CepSettings = ConstSettings.CommercialProducerSettings
 
 
+class DeviceValidator(metaclass=ABCMeta):
+    """Interface for devices' validator classes."""
+
+    @classmethod
+    @abstractmethod
+    def validate(cls, **kwargs):
```
Hassan754, line 38: Abstract methods should not define implementations, we should get rid of the `@ abstractmethod` decorator here

tests/test_device_validator.py
```diff
@@ -266,3 +268,49 @@ def test_finite_diesel_generator(self):
         self.assertIsNone(validate_finite_diesel_generator(energy_rate=100))
         with self.assertRaises(D3ADeviceException):
             validate_finite_diesel_generator(energy_rate=-1)
+
+
+class TestHomeMeterValidator(unittest.TestCase):
+    """Tests for the HomeMeterValidator class."""
+
+    @classmethod
+    def setUpClass(cls) -> None:
```
Hassan754, line 277: No need for the return type hint imo

tests/test_device_validator.py
```diff
@@ -266,3 +268,49 @@ def test_finite_diesel_generator(self):
...
+        """The validation fails when incompatible arguments are provided."""
+        failing_arguments = [
+            {"initial_buying_rate": 15, "final_buying_rate": 13},
+            {"energy_rate_increase_per_update": 1, "fit_to_limit": True},
+            {"initial_selling_rate": 10, "final_selling_rate": 13},
+            {"energy_rate_decrease_per_update": 1, "fit_to_limit": True}
+        ]
+        for arguments in failing_arguments:
+            with self.subTest(arguments=arguments):
+                with self.assertRaises(D3ADeviceException):
```
Hassan754, line 315: I'm not convinced why should we use subTest here? 
IMO the test is mature enough without it, no need for individual context for each of these assertions and the test should fail as soon as any validation passed without raising `D3ADeviceException`

Hassan754 (commented):
Left some comments

d3a_interface/sim_results/device_statistics.py
fievelk: I wanted to add a dict or a class to do the mapping, I think it's a good idea. I decided to leave it out of this PR to avoid too much refactoring, since the PR is already big. I can make a new PR and point to this one, so that we can keep things separate.

d3a_interface/device_validator.py
fievelk: In theory, we couldn't have a concrete method inside an abstract class, so that would also be a bit unconventional. Python isn't strict about the definition of abstract methods/classes, so it's possible to take advantage of it (see: https://pymotw.com/2/abc/#concrete-methods-in-abcs).

The reason why I made the `validate()` method abstract is that I wanted to make it very explicit that it is necessary for all validators, and it will be called by external objects.
But I definitely see your point and, in terms of final use, it wouldn't make a big difference. I can check again and see if changing it would improve it!

tests/test_device_validator.py
fievelk: Why not? Actually, that's also automatically filled up by PyCharm.

tests/test_device_validator.py
fievelk: We have several scenarios in which the validation should fail. You can think of each of these cases as a separate method (and they should be, in my opinion, but that's out of scope). 
Using just one set of failing arguments would not cover all cases. Suppose I mistakenly change the implementation of one of the `if` clauses in the `validate_price()` method, I will want tests to fail.

tests/test_device_validator.py
fievelk: Type hints add more verbosity to the code, but they help to spot mistakes while coding in IDEs or while using static type checkers. I would actually encourage doing it everywhere in the code, when possible.

d3a_interface/device_validator.py
```diff
@@ -24,10 +25,153 @@
...
+
+    @classmethod
+    def validate(cls, **kwargs):
+        """Validate both rate and energy values of the device."""
+        cls.validate_price(**kwargs)
+        cls.validate_energy(**kwargs)
+
+    @classmethod
+    @abstractmethod
+    def validate_price(cls, **kwargs):
```
faizan2590, line 44:
```suggestion
    def validate_rate(cls, **kwargs):
```

fievelk: I've used the same naming that was being used in the other validators (for load, pv, and storage). I think you're right though, since we're validating rates and not prices. I think the naming should be changed for the other validators too, but I would like to do it in a separate PR (to use validator classes for all of the devices).

d3a_interface/schemas.py
```diff
@@ -121,6 +122,27 @@ class ScenarioSchemas:
...
+                "properties": {
+                    "name": {"type": "string"},
+                    "type": {"enum": ["HomeMeter"]},
+                    "number_of_clones": {"type": "number"},
+                    "uuid": {"type": "string"},
+                    "libraryUUID": {"anyOf": [{"type": "string"}, {"type": "null"}]},
+                    "initial_selling_rate": {"anyOf": [{"type": "number"}, {"type": "null"}]},
+                    "final_selling_rate": {"type": "number"},
+                    "initial_buying_rate": {"type": "number"},
+                    "final_buying_rate": {"anyOf": [{"type": "number"}, {"type": "null"}]},
```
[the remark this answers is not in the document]
fievelk: Here I followed the same schema that was already defined for `pv` and `storage` devices. I didn't want to change it because I wasn't sure about the reason why it was defined that way. Could you please give me some info about it?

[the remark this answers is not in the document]
fievelk: Maybe that's because `initial_selling_rate` and `final_buying_rate` can be `None`, in case `use_market_maker_rate` is `True`.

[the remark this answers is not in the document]
hannesdiedrich: I think you got it right Pierpaolo. In case `use_market_maker_rate = True` , `final_buying_rate` and  `initial_selling_rate` could be None. The others should never be None.

hannesdiedrich (approved):
LGTM

d3a_interface/device_validator.py
```diff
@@ -24,10 +25,153 @@
...
+    def validate_rate(cls, **kwargs):
+        """Validate the rate values of the device."""
+
+    @classmethod
+    @abstractmethod
+    def validate_energy(cls, **kwargs):
+        """Validate the energy values of the device."""
+
+
+class HomeMeterValidator(DeviceValidator):
```
spyrostz, line 53: For now it is ok, but if we plan to employ it in other strategies then this class will repeat itself. This validator actually is more appropriate for the TemplateStrategyBidUpdater / TemplateStrategyOfferUpdater. The reason why (I guess) you had to implement it in the strategy class, was because these limits are different for each strategy class, and they needed to differentiate. Nothing to do now, just another enhancement topic would be to refactor the other strategies too to use these validation methods. Please leave it for another PR though, since this will delay the review process.

fievelk: Yes, right now we're just using individual functions to validate (e.g. `validate_storage_device(**kwargs)`, `validate_pv_device(**kwargs)` etc.). These are used both inside d3a strategies, and inside GraphQL output objects in d3a-web.
I decided to use a class to begin a gradual refactoring of this validation process. The next step would be to create classes for the other devices too (removing the individual functions). Once that is done, we can try to improve the internal implementation to avoid too much repetition (not sure how yet).

spyrostz (approved):
LGTM

PR #218, opened 2021-06-03
D3ASIM-3447: first myco api client release

Hassan754:


hannesdiedrich (approved):
LGTM

fievelk (dismissed):

fievelk:
@Hassan754 since you've been working with this, do you think it would make sense to turn these constants into Enums? (Not in the context of this PR)

d3a_interface/enums.py
```diff
@@ -0,0 +1,14 @@
+from enum import Enum
+
+
+class BidOfferMatchAlgoEnum(Enum):
+    PAY_AS_BID = 1
+    PAY_AS_CLEAR = 2
+    EXTERNAL = 3
+
+
+class SpotMarketTypeEnum(Enum):
```
hannesdiedrich, line 10: Just saw this. Do you think you could homogenize this with https://github.com/gridsingularity/d3a-web/blob/41bcce49f00dedf83723ac868f0e393ac141236a/simulation/graphql/graphql_enums.py#L28 ?

Hassan754: Done

hannesdiedrich (dismissed):
LGTM

d3a_interface/utils.py
```diff
@@ -454,3 +456,60 @@ def calculate_home_area_stats_from_repr_dict(cls, representation):
...
+    Returns: List[Dict]
+
+    """
+
+    return sorted(input_list,
+                  key=lambda obj: obj.get(attribute),
+                  reverse=reverse_order)
+
+
+def perform_pay_as_bid_match(market_offers_bids_list_mapping):
```
fievelk, line 480: I can't find usages of this method outside of the `myco-api-client` repository. If that's the case, I would think that this function might belong there, rather than in `d3a-interface`. What do you think? Are there other usages I couldn't find?

tests/test_utils.py
```diff
@@ -1,5 +1,8 @@
 import unittest
-from d3a_interface.utils import scenario_representation_traversal, HomeRepresentationUtils
+
+from d3a_interface.constants_limits import FLOATING_POINT_TOLERANCE
+from d3a_interface.utils import scenario_representation_traversal, HomeRepresentationUtils, \
+    sort_list_of_dicts_by_attribute, perform_pay_as_bid_match
```
fievelk, line 5: Nit:
```suggestion
from d3a_interface.utils import (
    scenario_representation_traversal, HomeRepresentationUtils, sort_list_of_dicts_by_attribute,
    perform_pay_as_bid_match)
```

d3a_interface/utils.py
Hassan754: The d3a has a perform_pay_as_bid function but with a slightly different interface, this is why I ported this here
In the context of my next task I may create an `Adapter` in the d3a and remove the function from there

d3a_interface/dataclasses.py
```diff
@@ -0,0 +1,19 @@
+from dataclasses import dataclass
+
+
+@dataclass
```
spyrostz, line 4: I like the dataclass usage here, however I think it is worth it to try it with Pypy 3.7 to validate that no issues occur (because the dataclasses were introduced in Python 3.7, would be great to check the Pypy implementation to rule out any incompatibilities)

spyrostz (approved):
LGTM

d3a_interface/dataclasses.py
Hassan754: AFAIU it works (couldn't validate this as pyenv is failing)

Hassan754: > This backport assumes that dict objects retain their insertion order. This is true in the language spec for Python 3.7 and greater. Since this is a backport to Python 3.6, it raises an interesting question: does that guarantee apply to 3.6? For CPython 3.6 it does. As of the time of this writing, it’s also true for all other Python implementations that claim to be 3.6 compatible, of which there are none. Any new 3.6 implementations are expected to have ordered dicts
This is from the pypi dataclasses documentation [link](https://pypi.org/project/dataclasses/)

spyrostz: This link is valid only for the backport, therefore it does not really say much about Pypy because we do not use the backport anyways (we do not run `pip install dataclass`). I strongly recommend creating a local d3a virtualenv with Pypy in order to test such cases (even without Pyenv you can install pypy3 directly in your machine and use it instead). For now feel free to merge this, since I am also 99% sure that this will work, I am mentioning the above in order to be prepared for any future relevant issues.

spyrostz (approved):
LGTM

fievelk (approved):

faizan2590 (approved):

PR #223, opened 2021-06-15
Feature/d3 asim 3464: Savings KPI

faizan2590:


d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +145,50 @@ def update_area_kpi(self, area_dict, core_stats):
...
+            if is_producer_node_type(child):
+                if_not_in_list_append(self.producer_list, child['uuid'])
+            elif is_load_node_type(child):
+                if_not_in_list_append(self.consumer_list, child['uuid'])
+            elif is_prosumer_node_type(child):
+                if_not_in_list_append(self.ess_list, child['uuid'])
+
+        fir_excl_gf_alp = \
+            core_stats.get(area_dict['uuid'], {}).get('feed_in_tariff', 0.) - gf_alp
+        mmr_incl_gf_alp = \
```
Hassan754, line 171: Name of variables should either be verbose or at least an inline comment needed

d3a_interface/sim_results/__init__.py
```diff
@@ -50,6 +50,17 @@ def is_buffer_node_type(area):
     return area['type'] == "InfiniteBusStrategy"
 
 
+def is_house_node_type(area):
+    return get_unified_area_type(area) == "Area"
+
+
+def has_no_grand_child(area):
```
Hassan754, line 57:
```suggestion
def has_no_grand_children(area):
```

d3a_interface/sim_results/kpi.py
```diff
@@ -221,6 +272,11 @@ def _kpi_ratio_to_percentage(self, area_name):
     def update(self, area_dict, core_stats, current_market_slot):
         if not self._has_update_parameters(area_dict, core_stats, current_market_slot):
             return
+
+        self.area_uuid_to_cum_fee_path[area_dict['uuid']] = \
+            self.area_uuid_to_cum_fee_path.get(area_dict['parent_uuid'], 0.) + \
+            core_stats.get(area_dict['uuid'], {}).get('const_fee_rate', 0.)
```
Hassan754, line 278: Minor, use parenthesis instead of backslashes for escaping new lines

d3a_interface/sim_results/kpi.py
```diff
@@ -162,6 +201,15 @@ def __repr__(self):
     def area_performance_indices(self, area_dict, core_stats):
         if area_dict['uuid'] not in self.state:
             self.state[area_dict['uuid']] = KPIState()
+
+        # initialization of house saving state
+        if area_dict['uuid'] not in self.saving_state and \
+                has_no_grand_child(area_dict):
+            self.saving_state[area_dict['uuid']] = SavingKPI()
+        if area_dict['uuid'] in self.saving_state:
```
Hassan754, line 209:
```suggestion
        if area_dict['uuid'] in self.saving_state:
          ...
        elif  has_no_grand_child(area_dict):
          ...
```

Hassan754 (commented):
Left some comments, will review once addressed

d3a_interface/sim_results/kpi.py
fievelk: I agree with Hassan about this :) Also, it would be good if you could use parentheses over backslashes in order to follow PEP-8 recommendations.

faizan2590: Correct me if I'm wrong, isn't that proposal is rejected: https://www.python.org/dev/peps/pep-3125/

d3a_interface/sim_results/__init__.py
faizan2590: Sorry again english grammar(not my mother tongue) but whats wrong with singular noun here? 
I'm checking not even a single grand kid for this target area.

Hassan754: If the noun following the no is countable, it is preferred to use it in plural 
https://englishlessonsbrighton.co.uk/followed-singular-plural-noun/

faizan2590: Thanks for the link, will adapt it accordingly.

d3a_interface/sim_results/kpi.py
fievelk: The PEP you linked was to completely eliminate `\` as a marker for line continuation from python. It's been rejected, so the `\` still works, but python guidelines recommend avoiding it and use parentheses instead:
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

Some exceptions are `assert` statements.

d3a_interface/sim_results/kpi.py
faizan2590: Can you please explain it, sorry haven'T actually understood it

Hassan754: This is to save up the first condition `area_dict['uuid'] not in self.saving_state`

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +146,59 @@ def update_area_kpi(self, area_dict, core_stats):
...
+                if_not_in_list_append(self.consumer_list, child['uuid'])
+            elif is_prosumer_node_type(child):
+                if_not_in_list_append(self.ess_list, child['uuid'])
+
+        # fir_excl_gf_alp: feed-in tariff rate excluding grid fee along path
+        fir_excl_gf_alp = (core_stats.get(area_dict['uuid'], {}).get('feed_in_tariff', 0.) -
+                           gf_alp)
+        # mmr_incl_gf_alp: market maker rate include grid fee along path
+        mmr_incl_gf_alp = (core_stats.get(area_dict['uuid'], {}).get('market_maker_rate', 0.) +
+                           gf_alp)
```
spyrostz, line 181: I think that, instead of using acronyms, a much easier approach for the reader would be to use separate small functions, that are self-documented by the function name. That way you will not need the acronyms and just name the variables as: 
`feed_in_tariff = self.get_feed_in_tariff_rate_excluding_path_grid_fees(core_stats, area_dict["uuid"], path_grid_fee)`

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +146,59 @@ def update_area_kpi(self, area_dict, core_stats):
...
+        self.saving_percentage = 0.
+
+    def calculate_saving_kpi(self, area_dict, core_stats, gf_alp):
+        """
+        :param area_dict: contain nested area info
+        :param core_stats: contain area's raw/key statistics
+        :param gf_alp: grid_fee_along_the_path - cumulative grid fee from root to target area
+        :return:
+        """
+        for child in area_dict['children']:
```
spyrostz, line 168: Trivial, but since this is an open-source repo, if you changed the ' to " throughout the file to adhere to the team standard would be great, thanks in advance!

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +146,59 @@ def update_area_kpi(self, area_dict, core_stats):
...
+        # fir_excl_gf_alp: feed-in tariff rate excluding grid fee along path
+        fir_excl_gf_alp = (core_stats.get(area_dict['uuid'], {}).get('feed_in_tariff', 0.) -
+                           gf_alp)
+        # mmr_incl_gf_alp: market maker rate include grid fee along path
+        mmr_incl_gf_alp = (core_stats.get(area_dict['uuid'], {}).get('market_maker_rate', 0.) +
+                           gf_alp)
+        for trade in core_stats.get(area_dict['uuid'], {}).get('trades', []):
+            if trade['seller_origin_id'] in [self.producer_list, self.ess_list]:
+                self.fit_revenue += fir_excl_gf_alp * trade['energy']
+                self.d3a_revenue += trade['rate'] * trade['energy']
```
spyrostz, line 185: Very minor, but in order to avoid the multiplication I think that you can use just `trade["offer"]["price"]`

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +146,59 @@ def update_area_kpi(self, area_dict, core_stats):
         self._accumulate_energy_trace(core_stats)
 
 
+class SavingKPI:
+    def __init__(self):
+        self.producer_list = list()
+        self.consumer_list = list()
+        self.ess_list = list()
+        self.house_list = list()
```
spyrostz, line 154: Again minor, but converting these to sets will help you avoid the `if_not_in_list_append` calls. Furthermore, judging from the ways that these lists are used, I would recommend not having 4 distinct lists, but 2: `producer_ess_list/set` , `consumer_ess_list/set`

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +146,59 @@ def update_area_kpi(self, area_dict, core_stats):
...
+                if_not_in_list_append(self.ess_list, child['uuid'])
+
+        # fir_excl_gf_alp: feed-in tariff rate excluding grid fee along path
+        fir_excl_gf_alp = (core_stats.get(area_dict['uuid'], {}).get('feed_in_tariff', 0.) -
+                           gf_alp)
+        # mmr_incl_gf_alp: market maker rate include grid fee along path
+        mmr_incl_gf_alp = (core_stats.get(area_dict['uuid'], {}).get('market_maker_rate', 0.) +
+                           gf_alp)
+        for trade in core_stats.get(area_dict['uuid'], {}).get('trades', []):
+            if trade['seller_origin_id'] in [self.producer_list, self.ess_list]:
```
spyrostz, line 183: This will not be needed due to another comment (the one that condenses the 4 lists to 2), but I think that the lists should be unpacked for this to work: `[*self.producer_list, *self._ess_list]`

spyrostz (commented):
Some comments, LGTM once these are resolved though. Also, would be great if you added unit tests for the SavingKPI class.

d3a_interface/sim_results/__init__.py
```diff
@@ -50,6 +50,17 @@ def is_buffer_node_type(area):
...
 
 
+def is_house_node_type(area):
+    return get_unified_area_type(area) == "Area"
+
+
+def has_no_grand_children(area):
+    for child in area.get('children', []):
+        if child.get('children', []):
+            return False
```
fievelk, line 60: Please use `"` over `'`

d3a_interface/sim_results/kpi.py
```diff
@@ -1,3 +1,4 @@
+
```
fievelk, line 1: Is this a typo?

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +146,59 @@ def update_area_kpi(self, area_dict, core_stats):
         self._accumulate_energy_trace(core_stats)
 
 
+class SavingKPI:
```
fievelk, line 149: Nit: is it supposed to be `Savings` or `Saving`?

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +146,59 @@ def update_area_kpi(self, area_dict, core_stats):
...
+        self.d3a_revenue = 0.
+        self.saving_absolute = 0.
+        self.saving_percentage = 0.
+
+    def calculate_saving_kpi(self, area_dict, core_stats, gf_alp):
+        """
+        :param area_dict: contain nested area info
+        :param core_stats: contain area's raw/key statistics
+        :param gf_alp: grid_fee_along_the_path - cumulative grid fee from root to target area
+        :return:
```
fievelk, line 166: As discussed on slack, please use Google-style docstrings rather than reStructured Text.

You can change the PyCharm settings from:
`Settings > Python integrated tools > Docstrings > Docstring format`

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +146,59 @@ def update_area_kpi(self, area_dict, core_stats):
...
+        """
+        for child in area_dict['children']:
+            if is_producer_node_type(child):
+                if_not_in_list_append(self.producer_list, child['uuid'])
+            elif is_load_node_type(child):
+                if_not_in_list_append(self.consumer_list, child['uuid'])
+            elif is_prosumer_node_type(child):
+                if_not_in_list_append(self.ess_list, child['uuid'])
```
fievelk, line 174: Nit: I think that these multiple `if` statements point out the need for refactoring, maybe using a separate class (possibly a mapper) to decouple this logic.
This would also make unittests much simpler to write.

fievelk (commented):
I would suggest adding unittests for the `SavingsKPI` class. In that context, I believe that refactoring the class (splitting it into different classes following the single-responsibility principle) would make things easier.

Nit: as suggested by spyros, please use `"` over `'`.

d3a_interface/sim_results/kpi.py
faizan2590: Target of saving up first condition will make loose saving kpi to be calculated for the first slot.

faizan2590: > In our case I see that the odds of area_dict['uuid'] in self.saving_state returning True is high + this is a O(1) statement that will be executed very quickly unlike the has_no_grand_children(area_dict) which is a recursive call and this is why this should be placed second

So to summarize, you proposal is dependent on probability, meaning you are willing to skip calculation for the first time result service has started/rebooted?

How about creating a nested if clause to avoid calculation of the expensive  `has_no_grand_children(area_dict)`

```
    if area_dict['uuid'] not in self.saving_state:
        if has_no_grand_children(area_dict):
            self.saving_state[area_dict['uuid']] = SavingsKPI()
```

d3a_interface/sim_results/kpi.py
faizan2590: Thanks, have used `trade["price"]`

d3a_interface/schemas.py
```diff
@@ -26,6 +26,9 @@ class ScenarioSchemas:
                     'type': {'anyOf': [{'enum': ['Area']}, {'enum': ['null']}]},
                     'name': {'type': 'string'},
                     'number_of_clones': {'type': 'number'},
+                    'const_fee_rate': {'type': 'number'},
+                    'feed_in_tariff': {'type': 'number'},
+                    'market_maker_rate': {'type': 'number'},
```
fievelk, line 31: Please use `"` in place of `'`.

d3a_interface/sim_results/__init__.py
```diff
@@ -50,6 +50,17 @@ def is_buffer_node_type(area):
     return area['type'] == "InfiniteBusStrategy"
 
 
+def is_house_node_type(area):
+    return get_unified_area_type(area) == "Area"
```
fievelk, line 54: Where is this method used? I can't seem to find it.

d3a_interface/sim_results/__init__.py
```diff
@@ -50,6 +50,17 @@ def is_buffer_node_type(area):
...
+
+
+def has_no_grand_children(area):
+    for child in area.get("children", []):
+        if child.get("children", []):
+            return False
+    return True
```
fievelk, line 61: Nit: this negative method is a bit "twisted". Maybe it would be better to have `has_grandchildren` and use it like: `if not has_grandchildren(area)`. What do you think?

d3a_interface/sim_results/__init__.py
```diff
@@ -50,6 +50,17 @@ def is_buffer_node_type(area):
     return area['type'] == "InfiniteBusStrategy"
 
 
+def is_house_node_type(area):
+    return get_unified_area_type(area) == "Area"
+
+
+def has_no_grand_children(area):
```
fievelk, line 57:
```suggestion
def has_no_grandchildren(area):
```
(or see comment below)

d3a_interface/sim_results/kpi.py
```diff
@@ -18,7 +18,7 @@
 from typing import Dict
 from d3a_interface.utils import if_not_in_list_append
 from d3a_interface.sim_results import is_load_node_type, is_buffer_node_type, \
-    is_prosumer_node_type, is_producer_node_type
+    is_prosumer_node_type, is_producer_node_type, has_no_grand_children
```
fievelk, line 21: Nit: please optimize the imports and use brackets over `\` for line continuation.

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +145,74 @@ def update_area_kpi(self, area_dict, core_stats):
...
+        self.house_list = list()
+        self.fit_revenue = 0.
+        self.utility_bill = 0.
+        self.base_case_revenue = 0.
+        self.d3a_revenue = 0.
+        self.saving_absolute = 0.
+        self.saving_percentage = 0.
+
+    def calculate_saving_kpi(self, area_dict, core_stats, gf_alp):
+        """
```
fievelk, line 161: Nit: the first line of a docstring should be a short summary that describes the method. Here it is missing, so it might be good to add it. Lines below could give some more information; for example: what's the "saving kpi"?

fievelk: Besides this, it's great that you're documenting the arguments for this function 👍

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +145,74 @@ def update_area_kpi(self, area_dict, core_stats):
...
+        Args:
+            area_dict: contain nested area info
+            core_stats: contain area's raw/key statistics
+            gf_alp: grid_fee_along_the_path - cumulative grid fee from root to target area
+
+        Returns:
```
fievelk, line 167: This can be omitted (or it should be filled with something)

tests/test_savings_kpi.py
```diff
@@ -0,0 +1,30 @@
...
+from d3a_interface.sim_results.kpi import SavingsKPI
+
+
+class TestSavingsKPI(TestCase):
+
+    def setUp(self) -> None:
+        self.savings_kpi = SavingsKPI()
+
+    def tearDown(self) -> None:
+        pass
```
fievelk, line 13:
```suggestion
```

tests/test_savings_kpi.py
```diff
@@ -0,0 +1,30 @@
+from unittest import TestCase
+from uuid import uuid4
+
+from d3a_interface.sim_results.kpi import SavingsKPI
+
+
+class TestSavingsKPI(TestCase):
```
fievelk, line 7: Since this is a new test, please avoid using `unittest.TestCase` and just rely on pytest's features. This is something that came up in the last two days during a conversation on slack.

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +145,74 @@ def update_area_kpi(self, area_dict, core_stats):
         self._accumulate_energy_trace(core_stats)
 
 
+class SavingsKPI:
+    def __init__(self):
+        self.producer_ess_set = set()
+        self.consumer_ess_set = set()
```
hannesdiedrich, line 151: Why is `ess` part of the name? I got confused and thought that it is a set of storages. But it is not, right?

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +145,74 @@ def update_area_kpi(self, area_dict, core_stats):
...
+            if trade["buyer_origin_id"] in self.consumer_ess_set and \
+                    trade["seller_origin_id"] not in self.producer_ess_set:
+                self.utility_bill += mmr_incl_gf_alp * trade["energy"]
+                self.d3a_revenue -= trade["price"]
+        self.base_case_revenue = self.fit_revenue - self.utility_bill
+        self.saving_absolute = self.d3a_revenue - self.base_case_revenue
+        self.saving_percentage = ((self.saving_absolute / self.base_case_revenue) * 100
+                                  if self.base_case_revenue else 0.)
+
+    def populate_area_type(self, area_dict):
```
hannesdiedrich, line 190: How about that?
```suggestion
    def populate_consumer_producer_sets(self, area_dict):
```

tests/test_savings_kpi.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def setUp(self) -> None:
+        self.savings_kpi = SavingsKPI()
+
+    def tearDown(self) -> None:
+        pass
+
+    def test_correct_classification_of_area_type(self):
```
fievelk, line 15: I would suggest naming unit tests including the name of the method you are testing. For example:
```suggestion
    def test_populate_area_type(self):
```
or
```suggestion
    def test_populate_area_type_is_correct(self):
```
So the format would be something like **<method_name>_<state_tested>_<expected_outcome>** or similar.

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +145,74 @@ def update_area_kpi(self, area_dict, core_stats):
...
+        self.consumer_ess_set = set()
+        self.house_list = list()
+        self.fit_revenue = 0.
+        self.utility_bill = 0.
+        self.base_case_revenue = 0.
+        self.d3a_revenue = 0.
+        self.saving_absolute = 0.
+        self.saving_percentage = 0.
+
+    def calculate_saving_kpi(self, area_dict, core_stats, gf_alp):
```
fievelk, line 160:
```suggestion
    def calculate_savings_kpi(self, area_dict, core_stats, gf_alp):
```

fievelk: This method is the main feature of this PR but it's not being unit-tested as far as I see. I suggest adding tests for it and for the individual smaller methods (e.g. `get_feed_in_tariff_rate_excluding_path_grid_fees` and the others).

d3a_interface/sim_results/kpi.py
```diff
@@ -162,45 +225,62 @@ def __repr__(self):
...
+                                self.state[area_dict['uuid']].total_energy_produced_wh)
         return {
             "self_sufficiency": self_sufficiency, "self_consumption": self_consumption,
             "total_energy_demanded_wh": self.state[area_dict['uuid']].total_demand,
             "demanded_buffer_wh": self.state[area_dict['uuid']].demanded_buffer_wh,
             "self_consumption_buffer_wh": self.state[area_dict['uuid']].self_consumption_buffer_wh,
             "total_energy_produced_wh": self.state[area_dict['uuid']].total_energy_produced_wh,
             "total_self_consumption_wh":
-                self.state[area_dict['uuid']].total_self_consumption_wh}
+                self.state[area_dict['uuid']].total_self_consumption_wh,
```
fievelk, line 269: Nit: please use `"` in place of `"` (also applies to the rest of the PR).

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +145,74 @@ def update_area_kpi(self, area_dict, core_stats):
...
+    def get_market_maker_rate_including_path_grid_fees(area_core_stat, path_grid_fee):
+        return area_core_stat.get("market_maker_rate", 0.) + path_grid_fee
+
+
 class KPI(ResultsBaseClass):
     def __init__(self):
         self.performance_indices = dict()
         self.performance_indices_redis = dict()
         self.state = {}
+        self.saving_state = {}
```
fievelk, line 214: Why is this a dict? If I'm not wrong, you're only filling it with one single element:
```python
self.saving_state[area_dict['uuid']] = SavingsKPI()
```
So why don't we instantiate it as `None` here, and do this later?
```python
self.saving_state = SavingsKPI()
```

fievelk: Moreover, the naming of the variable is still inconsistent: it should probably be `savings_state` (with an `s`)

d3a_interface/sim_results/kpi.py
```diff
@@ -162,45 +225,62 @@ def __repr__(self):
...
-                self.state[area_dict['uuid']].total_self_consumption_wh}
+                self.state[area_dict['uuid']].total_self_consumption_wh,
+            "base_case_revenue": getattr(self.saving_state.get(area_dict['uuid'], None),
+                                         'base_case_revenue', None),
+            "d3a_revenue": getattr(self.saving_state.get(area_dict['uuid'], None),
+                                   'd3a_revenue', None),
+            "saving_absolute": getattr(self.saving_state.get(area_dict['uuid'], None),
+                                       'saving_absolute', None),
+            "saving_percentage": getattr(self.saving_state.get(area_dict['uuid'], None),
+                                         'saving_percentage', None)
```
fievelk, line 277: I'm not sure I understand this part. Why are you using `getattr`? We know that all instances of `SavingsKPI` will always have these variables, since you've defined them in the `__init__()`: https://github.com/gridsingularity/d3a-interface/blob/2f2e624df987b329ece69544d13539f0712c452e/d3a_interface/sim_results/kpi.py#L26-L36.

What do you think about something like the following:
```suggestion
area_savings_kpi = self.saving_state.get(area_dict['uuid'])
return {
    ...
    "base_case_revenue": getattr(area_savings_kpi, 'base_case_revenue', None),
    "d3a_revenue": getattr(area_savings_kpi, 'd3a_revenue', None),
    "saving_absolute": getattr(area_savings_kpi, 'saving_absolute', None),
    "saving_percentage": getattr(area_savings_kpi, 'saving_percentage', None)
}
```
What do you think?

fievelk: Another thing that could be nice. You could add a `to_dict()` method in the `SavingsKPI` class, so that you would just have this:
```python
area_savings_kpi = self.saving_state.get(area_dict['uuid'])
area_savings_output = area_savings_kpi.to_dict()
return {
	    ...
	}.update(area_savings_output)
```

d3a_interface/sim_results/kpi.py
```diff
@@ -221,6 +301,11 @@ def _kpi_ratio_to_percentage(self, area_name):
     def update(self, area_dict, core_stats, current_market_slot):
         if not self._has_update_parameters(area_dict, core_stats, current_market_slot):
             return
+
+        parent_fee = self.area_uuid_to_cum_fee_path.get(area_dict.get('parent_uuid', ''), 0.)
+        area_fee = core_stats.get(area_dict.get('uuid', ''), {}).get('const_fee_rate', 0.)
+        self.area_uuid_to_cum_fee_path[area_dict['uuid']] = parent_fee + area_fee
+
```
fievelk, line 308: I'm having some issues trying to understand what object is expected from all these chained `.get()` calls. I think it would be good to refactor this to improve clarity.
Also, why is the default value sometimes `''` and some other times `None` (not here, but in the whole PR)?

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +145,74 @@ def update_area_kpi(self, area_dict, core_stats):
...
+            if trade["seller_origin_id"] in self.producer_ess_set:
+                self.fit_revenue += fir_excl_gf_alp * trade["energy"]
+                self.d3a_revenue += trade["price"]
+            if trade["buyer_origin_id"] in self.consumer_ess_set and \
+                    trade["seller_origin_id"] not in self.producer_ess_set:
+                self.utility_bill += mmr_incl_gf_alp * trade["energy"]
+                self.d3a_revenue -= trade["price"]
+        self.base_case_revenue = self.fit_revenue - self.utility_bill
+        self.saving_absolute = self.d3a_revenue - self.base_case_revenue
+        self.saving_percentage = ((self.saving_absolute / self.base_case_revenue) * 100
```
hannesdiedrich, line 187: Although I got a hold of what these variables represent, it would be great if you could either add a comment in `__init__` for each of them or add a description in the docstring. Thanks a lot!

fievelk:
I left some comments. I think some refactoring would help to make the code more maintainable.

d3a_interface/sim_results/__init__.py
faizan2590: Added and skipped its usage down the road while implementing. Thanks for catching will remove it now.

d3a_interface/sim_results/__init__.py
faizan2590: To me it looks getting twisted after explicitly adding `not` in `has_grandchildren `, nevertheless I'm indifferent and will adapt as you like here.

d3a_interface/sim_results/kpi.py
faizan2590: https://github.com/gridsingularity/d3a-interface/pull/223#discussion_r654221854

d3a_interface/sim_results/kpi.py
faizan2590: > Why is this a dict?

I need to calculate savings kpi for individual areas and all areas calculations are independent of each other.

d3a_interface/sim_results/kpi.py
faizan2590: > I'm not sure I understand this part. Why are you using `getattr`?
Purpose to use `getattr` is `saving_kpi` is only instantiated for house areas. Therefore it will raise the following exception for non-house areas.
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'abc'
```

fievelk: > Purpose to use getattr is saving_kpi is only instantiated for house areas. Therefore it will raise the following exception for non-house areas.

Thanks for the explanation :) I think this should be made explicit, and the two cases (house vs non-house areas) should be separated.
What is happening at the moment is basically this, for cases in which `SavingsKPI` hasn't been instantiated:
```python
getattr(None, "some_attribute", None)
```
This code is in my opinion too "broad", because it doesn't say much about the objects we're dealing with and it might lead to unexpected effects in the future. In theory, one could replace the values if `self.saving_state` with any possible object, and the code wouldn't raise any useful exception (making it impossible to debug).

If the area we're dealing with is not supposed to have a `savings_kpi` attribute (e.g. it's a house area), we should not try to query that attribute. For the same reason, the `KPI` instance of that area should not have a `savings_kpi` attribute, since the area doesn't support it.
One option would be to create separate classes (respecting the single-responsibility principle): one that deals with KPIs of house areas and another one that deals with other area types.

fievelk: Update: as you said, the `KPI` instance contains kpis for all the areas (not just one) so my suggestion of using two separate KPI classes doesn't probably apply. What I see is that the `area_performance_indices` method is doing "too much", so I have the feeling that it could be refactored. In any case, what I'm suggesting is to create more specialized code based on the type of area we're dealing with, and avoid the generic `getattr`.

faizan2590:
```
        kpi_parm_dict = {
            "self_sufficiency": self_sufficiency, "self_consumption": self_consumption,
            "total_energy_demanded_wh": self.state[area_dict["uuid"]].total_demand,
            "demanded_buffer_wh": self.state[area_dict["uuid"]].demanded_buffer_wh,
            "self_consumption_buffer_wh": self.state[area_dict["uuid"]].self_consumption_buffer_wh,
            "total_energy_produced_wh": self.state[area_dict["uuid"]].total_energy_produced_wh,
            "total_self_consumption_wh": self.state[area_dict["uuid"]].total_self_consumption_wh,
        }
        if self.savings_state.get(area_dict["uuid"]) is not None:
            kpi_parm_dict.update(self.savings_state[area_dict["uuid"]].to_dict())
        return kpi_parm_dict
```

@fievelk How about this for now?

d3a_interface/sim_results/kpi.py
```diff
@@ -147,26 +148,25 @@ def update_area_kpi(self, area_dict, core_stats):
...
+        self.d3a_revenue = 0.  # standard revenue of a house participating in D3A
+        self.saving_absolute = 0.  # savings achieved by a house via participating in D3A
+        self.saving_percentage = 0.  # savings in percentage wrt FIT via participating in D3A
+
+    def calculate_savings_kpi(self, area_dict, core_stats, gf_alp):
+        """Calculates the referenced saving from feed-in tariff based participation vs D3A
         Args:
             area_dict: contain nested area info
-            core_stats: contain area's raw/key statistics
+            core_stats: contain area"s raw/key statistics
```
fievelk, line 164: Nit: a side-effect of the global replacement of `'` :)

d3a_interface/sim_results/kpi.py
fievelk: So if `savings_state[area_dict["uuid"]]` is not there, it means that the area is a house (and it doesn't have KPIs for savings). Am I correct? If so, adding a comment to explain this would be great. In any case, this looks like a good way to go in my opinion 👍 Thanks for dealing with it!

faizan2590: > So if savings_state[area_dict["uuid"]] is not there, it means that the area is`n't` a house (and it doesn't have KPIs for savings).

Minor correction(hope that was a typo)

fievelk: Yes: `savings_kpi` is only instantiated for house areas, so if it's not there it means that the area is not a house. Thanks for the correction! 👍

tests/test_savings_kpi.py
faizan2590: https://github.com/gridsingularity/d3a-interface/pull/223/commits/cb925bc2368e0d411295da61bf6b9e52467c9ae3

d3a_interface/sim_results/kpi.py
faizan2590: Added: https://github.com/gridsingularity/d3a-interface/pull/223/files#diff-4fe53216bd5f83d932a02c99f5062b7714e165902d3aad6274ef33e009b5b2cbR151-R158

d3a_interface/sim_results/kpi.py
faizan2590: `savings_kpi` with grid fees being tested via integration_tests, so not sure if still the need arise to write unittests

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +146,78 @@ def update_area_kpi(self, area_dict, core_stats):
...
+                "saving_percentage": self.saving_percentage}
+
+
 class KPI(ResultsBaseClass):
     def __init__(self):
         self.performance_indices = dict()
         self.performance_indices_redis = dict()
         self.state = {}
+        self.savings_state = {}  # keeps initialized savings kpi object for individual area
+        self.area_uuid_to_cum_fee_path = {}  # keeps track of grid fee from root to target area
```
hannesdiedrich, line 220: Minor:
Sorry, i just saw this variable now. I found the name and the description a bid confusing. Hier a proposal for change (or you have a better idea).
```suggestion
        self.area_uuid_cum_grid_fee_mapping = {}  # mapping of area_uuid and the corresponding cumulative grid-free along the path from root-area to the area 
```

d3a_interface/sim_results/kpi.py
```diff
@@ -212,43 +292,52 @@ def _kpi_ratio_to_percentage(self, area_name):
...
-        self.performance_indices[area_dict['uuid']] = \
-            self.area_performance_indices(area_dict, core_stats)
-        self.performance_indices_redis[area_dict['uuid']] = \
-            self._kpi_ratio_to_percentage(area_dict['uuid'])
 
-        for child in area_dict['children']:
-            if len(child['children']) > 0:
+        parent_fee = self.area_uuid_to_cum_fee_path.get(area_dict.get("parent_uuid", None), 0.)
+        area_fee = core_stats.get(area_dict.get("uuid", None), {}).get("const_fee_rate", 0.)
+        self.area_uuid_to_cum_fee_path[area_dict["uuid"]] = parent_fee + area_fee
```
hannesdiedrich, line 311: Also minor, but this part could live inside a separate function.

hannesdiedrich: Advantage of that could be to unit test this tracking of the cumulative fee

hannesdiedrich (commented):
I would be in favour of more unittests of the SavingsKPI.
Also added some minor comments, once addressed, LGTM

hannesdiedrich (approved):
Some more unittest could make the codecov warnings go away and potentially save us some bugs in the future. Also some more docstrings at least for the new code would be helpful. 
LGTM nevertheless

faizan2590:
> Some more unittest could make the codecov warnings go away

https://github.com/gridsingularity/d3a-interface/pull/223/commits/db327b50ac41b9ac3a35188ca416703b91ae99bc

> Also some more docstrings at least for the new code would be helpful.

https://github.com/gridsingularity/d3a-interface/pull/223/commits/2f2a483a07c25e45313e67feec621311299a2060

d3a_interface/sim_results/kpi.py
```diff
@@ -145,11 +146,104 @@ def update_area_kpi(self, area_dict, core_stats):
...
+                self.utility_bill += mmr_incl_gf_alp * trade["energy"]
+                self.d3a_cost += trade["price"]
+        self.base_case_cost = self.utility_bill - self.fit_revenue
+        self.saving_absolute = self.base_case_cost - self.d3a_cost
+        self.saving_percentage = ((self.saving_absolute / self.base_case_cost) * 100
+                                  if self.base_case_cost else 0.)
+
+    def populate_consumer_producer_sets(self, area_dict: dict):
+        """
+        Responsible for devices' classification
```
hannesdiedrich, line 192: Docstring should be written in imperative and should be ending with a dot.
```suggestion
        Populate sets of device classes.
```

d3a_interface/sim_results/kpi.py
hannesdiedrich: reminder

hannesdiedrich (approved):
LGTM

spyrostz (dismissed):
LGTM

PR #224, opened 2021-06-16
D3ASIM-2540: Update read_and_convert_identity_profile_to_float to not…

Hassan754:
… parse string as it is already parsed in read_arbitrary_profile


faizan2590 (approved):
LGTM

PR #228, opened 2021-06-29
Refactor device validation tests using pytest

fievelk:
- `unittest.subTest` can't be used with `pytest`, unless additional packages are employed (see [pytest docs](https://docs.pytest.org/en/6.2.x/unittest.html?highlight=subtest));
- `@pytest.mark.parametrize` and fixtures can't be used if the test class inherits from `unittest.TestCase`.

For these reasons, this PR refactors the unit tests for device validation to make a broader use of pytest, its assertions and its parametrized parameters.

### DIFF
--- a/tests/test_device_validator.py
+++ b/tests/test_device_validator.py
@@ -15,9 +15,10 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-import unittest
 from unittest.mock import patch
 from uuid import uuid4
+
+import pytest
 
 from d3a_interface.constants_limits import ConstSettings
 from d3a_interface.device_validator import (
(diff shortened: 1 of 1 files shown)


tests/test_device_validator.py
```diff
@@ -32,262 +33,230 @@
...
+        {"hrs_per_day": 1, "daily_load_profile": VALID_LOAD_PROFILE},
+        {"hrs_of_day": [1, 2, 3, 4, 5, 6], "daily_load_profile": VALID_LOAD_PROFILE}
+    ])
+    def test_load_device_setting_fails(failing_arguments):
+        """The PV device validation fails when incompatible arguments are provided."""
+        with pytest.raises(D3ADeviceException):
+            validate_load_device(**failing_arguments)
+
+    @staticmethod
+    @pytest.mark.parametrize("valid_arguments", [
```
spyrostz, line 81: This is a good opportunity to talk about it, up until now we used this package for parameterized tests https://pypi.org/project/parameterized/ (due to lack of the parametrize parameter in pytest.mark some years ago IIRC). Since such functionality is available in pytest now I think it is worth it to use it for all new code and gradually deprecate the use of the parameterized package (I am not sure whether pytest supports parameterized classes, but we do not use this feature anyways). Nothing to do in the context of this PR, just something to keep in mind for future PRs.

spyrostz (approved):
LGTM

hannesdiedrich (approved):
Great refactoring! LGTM

faizan2590 (approved):
LGTM

tests/test_device_validator.py
fievelk: I agree, I think we can gradually refactor existing code when we deal with it, until we can completely remove that dependency.

PR #229, opened 2021-06-30
D3ASIM-3356: Add abstract matching algorithm class | refactor pay_as_…

Hassan754:
…bid_matcher


Hassan754:
@spyrostz @hannesdiedrich @faizan2590 @fievelk I was wondering if we should allow some matching algorithms to reside in the Myco api client repo, for instance is it okay for the in progress attributed matcher algorithm to reside there ?

hannesdiedrich:
> @spyrostz @hannesdiedrich @faizan2590 @fievelk I was wondering if we should allow some matching algorithms to reside in the Myco api client repo, for instance is it okay for the in progress attributed matcher algorithm to reside there ?

As discussed on slack I would vote for moving all matching algorithms into the d3a-interface in order to be able to imported from the d3a and from the myco-api-client. Each new matching algorithm we provide should live here too, IMO.

d3a_interface/matching_algorithms/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .abstract_matching_algorithm import AbstractMatchingAlgorithm
+from .pay_as_bid_algorithm import PayAsBidMatchingAlgorithm
+from .pay_as_clear_algorithm import PayAsClearMatchingAlgorithm
+
+__all__ = [
+    AbstractMatchingAlgorithm,
+    PayAsBidMatchingAlgorithm,
+    PayAsClearMatchingAlgorithm
+]
```
fievelk, line 9: I think this should be a list of strings. From the [docs](https://docs.python.org/3.8/reference/simple_stmts.html):
> The public names defined by a module are determined by checking the module’s namespace for a variable named __all__; if defined, it must be a sequence of strings which are names defined or imported by that module.

d3a_interface/matching_algorithms/abstract_matching_algorithm.py
```diff
@@ -0,0 +1,16 @@
+from abc import ABC, abstractmethod
+from typing import Dict, List
+
+
+class AbstractMatchingAlgorithm(ABC):
```
hannesdiedrich, line 5: Minor and not sure, but I would not put "abstract" into the name.
How about `BaseMatchingAlgorithm`

hannesdiedrich (approved):
LGTM

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
One comment, once addressed/adapted, LGTM from my side

spyrostz (approved):
LGTM

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
```diff
@@ -0,0 +1,212 @@
...
+from typing import List, Dict, Tuple
+
+from d3a_interface.constants_limits import ConstSettings
+from d3a_interface.dataclasses import MarketClearingState, Clearing, BidOfferMatch
+from d3a_interface.matching_algorithms import BaseMatchingAlgorithm
+from d3a_interface.utils import sort_list_of_dicts_by_attribute, add_or_create_key
+
+log = getLogger(__name__)
+
+MATCH_FLOATING_POINT_TOLERANCE = 1e-8
```
faizan2590, line 32: Why not using [FLOATING_POINT_TOLERANCE](https://github.com/gridsingularity/d3a-interface/blob/master/d3a_interface/constants_limits.py#L246) and possibly update the precision level ?

Hassan754: I did not touch the code here, however these two have different value AFAIS

faizan2590: Leave it as it is for now, but do look in your next enhancement ticket if these two closely related constants could be merged

PR #230, opened 2021-06-30
D3ASIM-3505: Validate fit_to_limit False

fievelk:
If `fit_to_limit` is `False`, the rate increase/decrease must be explicitly specified when creating a device.

This PR adds device validation for the case in which `fit_to_limit == False` and `energy_rate_increase_per_update == None` or `energy_rate_decrease_per_update == None`.


d3a_interface/device_validator.py
```diff
@@ -60,9 +61,34 @@ def validate_energy(cls, **kwargs):
     @classmethod
     def validate_rate(cls, **kwargs):
         """Validate rates of a Home Meter device."""
+
+        cls._validate_fit_to_limit(
+            fit_to_limit=kwargs.get("fit_to_limit"),
+            energy_rate_increase_per_update=kwargs.get("energy_rate_increase_per_update"),
+            energy_rate_decrease_per_update=kwargs.get("energy_rate_decrease_per_update"))
```
Hassan754, line 68: it would be better if we stick to a way of calling the validators, either call them and pass the kwargs arguments at once and let them handle unpacking them, or explicitly passing arguments for each validator

d3a_interface/device_validator.py
```diff
@@ -60,9 +61,34 @@ def validate_energy(cls, **kwargs):
...
         cls._validate_home_meter_consumption_rates(**kwargs)
         cls._validate_home_meter_production_rates(**kwargs)
 
+    @staticmethod
+    def _validate_fit_to_limit(
+            fit_to_limit: Optional[bool], energy_rate_increase_per_update: Optional[bool],
+            energy_rate_decrease_per_update: Optional[bool]):
+        if fit_to_limit is True and (
```
Hassan754, line 76: What other value can `fit_to_limit` hold besides True, False or None? if no other values can be received remove the is True operator

Hassan754 (approved):
LGTM

d3a_interface/device_validator.py
fievelk: I think using `**kwargs` to later unpack lots of parameters should be avoided as much as possible. In some cases it can help, but most of the time it creates confusion and can lead to unexpected behavior.
I much prefer, when possible, having named parameters.

In this case, I think it's fine to pass `**kwargs` to the main `validate()` method. Internally, however, the other methods (like `_validate_fit_to_limit`) should make it clear that they are specialized in handling specific parameters.

fievelk: Actually, I believe that `_validate_fit_to_limit` and all other internal validations should not be methods, but rather separate classes (e.g. `FitToLimitValidator`, `EnergyRateValidator` and so on). This would allow combining them together more flexibly. But this might be part of a separate refactoring PR.

d3a_interface/device_validator.py
fievelk: In general I agree with you. In this specific case I didn't know whether we should have been more strict (in case someone decided to pass `fit_to_limit="yeah!"`) but I just remembered that the schema should fail in that case: https://github.com/gridsingularity/d3a-interface/blob/348a9ecb626e23cbeff2694ebc84a587b1f730dd/d3a_interface/schemas.py#L62

So I can change it as you suggested. Thanks!

d3a_interface/device_validator.py
Hassan754: I agree, it's just that I see in the same scope functions with different arguments style, for example the `_validate_home_meter_consumption_rates` is being passed kwargs arguments

d3a_interface/device_validator.py
hannesdiedrich: Please let me remind you on the frontend. They there are a lot of places where they send "null" instead of a False, so please prepare for that too.

hannesdiedrich: Sorry, only got it now. So all good.

d3a_interface/device_validator.py
```diff
@@ -60,9 +61,34 @@ def validate_energy(cls, **kwargs):
...
         """Validate rates of a Home Meter device."""
+
+        cls._validate_fit_to_limit(
+            fit_to_limit=kwargs.get("fit_to_limit"),
+            energy_rate_increase_per_update=kwargs.get("energy_rate_increase_per_update"),
+            energy_rate_decrease_per_update=kwargs.get("energy_rate_decrease_per_update"))
         cls._validate_home_meter_consumption_rates(**kwargs)
         cls._validate_home_meter_production_rates(**kwargs)
 
+    @staticmethod
```
spyrostz, line 72: I assume that refactoring this in order to use this method and not the duplicated code that checks fit to limit-related parameters for each device is part of a following PR , right ?

spyrostz (approved):
If the removal of the duplicate code is part of another PR, LGTM

d3a_interface/device_validator.py
fievelk: Unfortunately, I'm not sure how to proceed with that. This function deals with two energy rates:
- `energy_rate_increase_per_update`
- `energy_rate_decrease_per_update`
Most devices only have one of the two, which means that the other one will always be `None`. For those devices, this would then always raise the exception, which is not what we want.

Maybe I'm missing something though. Do you have any suggestions to re-use this?

Hassan754 (approved):
LGTM

fievelk:
One of the tests is failing, but only when run through `coverage` (it works when using `pytest` directly). I'll investigate.

d3a_interface/validators/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+from d3a_interface.validators.finite_diesel_generator_validator import (
+    FiniteDieselGeneratorValidator)
+from d3a_interface.validators.home_meter_validator import HomeMeterValidator
+from d3a_interface.validators.infinite_bus_validator import InfiniteBusValidator
+from d3a_interface.validators.load_validator import LoadValidator
+from d3a_interface.validators.market_maker_validator import MarketMakerValidator
+from d3a_interface.validators.pv_validator import PVValidator
+from d3a_interface.validators.storage_validator import StorageValidator
+
+# Explicitly declare the names of the module's public API
```
hannesdiedrich, line 26: Sorry if this has been discussed in one of the other PRs. Can you maybe point me to the discussion and explain why this is needed?

d3a_interface/validators/finite_diesel_generator_validator.py
```diff
@@ -0,0 +1,54 @@
...
+                                         f"{CepSettings.MAX_POWER_KW_LIMIT.max}."]}
+                for value in kwargs["max_available_power_kW"].values():
+                    utils.validate_range_limit(
+                        CepSettings.MAX_POWER_KW_LIMIT.min, value,
+                        CepSettings.MAX_POWER_KW_LIMIT.max, error_message)
+            else:
+                raise D3ADeviceException({
+                    "misconfiguration": ["max_available_power_kW has an invalid type."]})
+
+        super().validate(**kwargs)
```
hannesdiedrich, line 54: Now I see, understand, agree and congratulate for the reusage of code. :-D

d3a_interface/validators/home_meter_validator.py
```diff
@@ -0,0 +1,159 @@
...
+                    "together."]})
+
+        if fit_to_limit is False and (
+                energy_rate_increase_per_update is None
+                or energy_rate_decrease_per_update is None):
+            raise D3ADeviceException(
+                {"misconfiguration": [
+                    "energy_rate_increase/decrease_per_update must be set if fit_to_limit is "
+                    "False."]})
+
```
hannesdiedrich, line 64: Hmm, this could be reused in the storage strategy, right?

d3a_interface/validators/__init__.py
Hassan754: btw pep states that the dunder names should be placed before imports https://www.python.org/dev/peps/pep-0008/#id24

hannesdiedrich (dismissed):
One question and one refactoring proposal, LGTM

d3a_interface/validators/__init__.py
fievelk: @hannesdiedrich Adding the imports in `__init__.py` allows to import the validators directly from the `validators` package, without the specific module:
```python
from d3a_interface.validators import CommercialProducerValidator
# Instead of this
from d3a_interface.validators.cep_validator import CommercialProducerValidator
```
However, flake8 fails if we just write the imports:
> `./d3a_interface/validators/__init__.py:17:1`: F401 'd3a_interface.validators.cep_validator.CommercialProducerValidator' imported but unused

One option would be to ignore this using `#noqa` or similar. Another approach is to use `__all__`. From PEP8:
> To better support introspection, modules should explicitly declare the names in their public API using the `__all__` attribute. Setting `__all__` to an empty list indicates that the module has no public API.

This is especially useful to limit what gets imported when you write `from d3a_interface.validators import *` (see [Importing from a package](https://docs.python.org/3.8/tutorial/modules.html#importing-from-a-package)).

@Hassan754 interesting, I didn't know that! Thanks, I'm moving the imports below then.

d3a_interface/validators/home_meter_validator.py
fievelk: I think you're right! I can move it to the `utils` module for the moment and share it between the two validators. I have the feeling that these functions can become their own classes (e.g. `FitToLimitValidator`) but this might be for a different PR :)

d3a_interface/validators/__init__.py
hannesdiedrich: Thanks a lot! Makes sense and I will consider this way for my next implementations.

d3a_interface/validators/home_meter_validator.py
hannesdiedrich: :-D Yes, leave the different class for now and only move this to a utility function. But why not making it a static function in the Baseclass? Or is this not a common practice?

Hassan754 (approved):
LGTM

d3a_interface/validators/utils.py
```diff
@@ -14,13 +14,39 @@
...
 
 CepSettings = ConstSettings.CommercialProducerSettings
 
 
+def validate_fit_to_limit(
+        fit_to_limit: Optional[bool], energy_rate_increase_per_update: Optional[bool],
+        energy_rate_decrease_per_update: Optional[bool]):
+    """Check that fit_to_limit is correct when both energy rates changes are allowed.
+
+    Important: don't use this method to validate devices that only accept one rate (e.g. Load and
```
hannesdiedrich, line 30: 👍

hannesdiedrich (approved):
LGTM

d3a_interface/validators/home_meter_validator.py
hannesdiedrich: Ok, yes, convinced. It is really a util. Thanks a lot!

PR #233, opened 2021-07-01
Refactor validators to implement a common interface

fievelk:
This PR unifies the validators interface. This potentially allows to remove code duplication and to simplify the creation of new validators in the future.

### DIFF
--- a/d3a_interface/validators/finite_diesel_generator_validator.py
+++ b/d3a_interface/validators/finite_diesel_generator_validator.py
@@ -27,28 +27,30 @@ class FiniteDieselGeneratorValidator(CommercialProducerValidator):
     @classmethod
     def validate(cls, **kwargs):
         """Validate the parameters of the device."""
-        if kwargs.get("max_available_power_kW") is not None:
-            if isinstance(kwargs["max_available_power_kW"], (int, float)):
-                error_message = {
-                    "misconfiguration": ["max_available_power_kW should be in between "
-                                         f"{CepSettings.MAX_POWER_KW_LIMIT.min} & "
-                                         f"{CepSettings.MAX_POWER_KW_LIMIT.max}."]}
+        if kwargs.get("max_available_power_kW") is None:
+            return
+
+        if isinstance(kwargs["max_available_power_kW"], (int, float)):
+            error_message = {
+                "misconfiguration": ["max_available_power_kW should be in between "
+                                     f"{CepSettings.MAX_POWER_KW_LIMIT.min} & "
+                                     f"{CepSettings.MAX_POWER_KW_LIMIT.max}."]}
+            utils.validate_range_limit(
+                CepSettings.MAX_POWER_KW_LIMIT.min,
+                kwargs["max_available_power_kW"],
+                CepSettings.MAX_POWER_KW_LIMIT.max,
+                error_message)
+        elif isinstance(kwargs["max_available_power_kW"], dict):
+            error_message = {
+                "misconfiguration": ["max_available_power_kW should be in between "
+                                     f"{CepSettings.MAX_POWER_KW_LIMIT.min} & "
+                                     f"{CepSettings.MAX_POWER_KW_LIMIT.max}."]}
+            for value in kwargs["max_available_power_kW"].values():
                 utils.validate_range_limit(
-                    CepSettings.MAX_POWER_KW_LIMIT.min,
-                    kwargs["max_available_power_kW"],
-                    CepSettings.MAX_POWER_KW_LIMIT.max,
-                    error_message)
-            elif isinstance(kwargs["max_available_power_kW"], dict):
-                error_message = {
-                    "misconfiguration": ["max_available_power_kW should be in between "
-                                         f"{CepSettings.MAX_POWER_KW_LIMIT.min} & "
-                                         f"{CepSettings.MAX_POWER_KW_LIMIT.max}."]}
-                for value in kwargs["max_available_power_kW"].values():
-                    utils.validate_range_limit(
-                        CepSettings.MAX_POWER_KW_LIMIT.min, value,
-                        CepSettings.MAX_POWER_KW_LIMIT.max, error_message)
-            else:
-                raise D3ADeviceException({
-                    "misconfiguration": ["max_available_power_kW has an invalid type."]})
+                    CepSettings.MAX_POWER_KW_LIMIT.min, value,
+                    CepSettings.MAX_POWER_KW_LIMIT.max, error_message)
+        else:
+            raise D3ADeviceException({
+                "misconfiguration": ["max_available_power_kW has an invalid type."]})
 
         super().validate(**kwargs)
--- a/d3a_interface/validators/finite_diesel_generator_validator.py
+++ b/d3a_interface/validators/finite_diesel_generator_validator.py
@@ -15,9 +15,9 @@ not, see <http://www.gnu.org/licenses/>.
 """
 from d3a_interface.constants_limits import ConstSettings
 from d3a_interface.exceptions import D3ADeviceException
-from d3a_interface.validators import BaseValidator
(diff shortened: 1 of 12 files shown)


d3a_interface/validators/finite_diesel_generator_validator.py
```diff
@@ -16,34 +16,39 @@
...
-                                      f"{CepSettings.MAX_POWER_KW_LIMIT.min} & "
-                                      f"{CepSettings.MAX_POWER_KW_LIMIT.max}."]}
-            for _, value in kwargs["max_available_power_kW"].items():
+class FiniteDieselGeneratorValidator(CommercialProducerValidator):
+    """Validator class for Finite Diesel Generators."""
+
+    @classmethod
+    def validate(cls, **kwargs):
+        """Validate the parameters of the device."""
+        if kwargs.get("max_available_power_kW") is not None:
```
Hassan754, line 30: You can save up an indentation level
```suggestion
        if kwargs.get("max_available_power_kW") is None:
               super().validate(**kwargs)
        if isinstance(kwargs["max_available_power_kW"], (int, float)):
```

d3a_interface/validators/utils.py
```diff
@@ -21,39 +21,42 @@
...
 
 
 def _validate_rate_profile(energy_rate_profile):
     for date, value in energy_rate_profile.items():
-        value = float(value) if type(value) == str else value
-        error_message = \
-            {"misconfiguration": [f"energy_rate should at time: {date} be in between "
-                                  f"{CepSettings.ENERGY_RATE_LIMIT.min} & "
-                                  f"{CepSettings.ENERGY_RATE_LIMIT.max}."]}
+        value = float(value) if isinstance(value, str) else value
```
Hassan754, line 57: This should work in case the value is a str or float
```suggestion
        value = float(value)
```

d3a_interface/validators/pv_validator.py
```diff
@@ -15,85 +15,93 @@
...
-        if (kwargs["cloud_coverage"] != 4) and \
-           ("power_profile" in kwargs and kwargs["power_profile"] is not None):
-            raise D3ADeviceException(
-                {"misconfiguration": [f"cloud_coverage (if values 0-3) & "
-                                      f"power_profile can't be set together."]})
-
+    @classmethod
+    def validate_energy(cls, **kwargs):
+        """Validate energy values of the device."""
+        if "panel_count" in kwargs and kwargs["panel_count"] is not None:
```
Hassan754, line 37:
```suggestion
        if kwargs.get("panel_count") is not None:
```

d3a_interface/validators/load_validator.py
```diff
@@ -15,104 +15,113 @@
...
-        error_message = {"misconfiguration": [f"avg_power_W should be in between "
-                                              f"{LoadSettings.AVG_POWER_LIMIT.min} & "
-                                              f"{LoadSettings.AVG_POWER_LIMIT.max}."]}
-        validate_range_limit(LoadSettings.AVG_POWER_LIMIT.min,
-                             kwargs["avg_power_W"],
-                             LoadSettings.AVG_POWER_LIMIT.max, error_message)
+    @classmethod
+    def validate_energy(cls, **kwargs):
+        """Validate energy values of the device."""
+        if "avg_power_W" in kwargs and kwargs["avg_power_W"] is not None:
```
Hassan754, line 37:
```suggestion
        if kwargs.get("avg_power_W") is not None:
```

d3a_interface/validators/load_validator.py
```diff
@@ -15,104 +15,113 @@
...
-            {"misconfiguration": [f"hrs_of_day should be less between "
-                                  f"{LoadSettings.HOURS_LIMIT.min} & "
-                                  f"{LoadSettings.HOURS_LIMIT.max}."]})
-    if ("hrs_of_day" in kwargs and kwargs["hrs_of_day"] is not None) and \
-            ("hrs_per_day" in kwargs and kwargs["hrs_per_day"] is not None) and \
-            (len(kwargs["hrs_of_day"]) < kwargs["hrs_per_day"]):
-        raise D3ADeviceException(
-            {"misconfiguration": [f"length of hrs_of_day list should be "
-                                  f"greater than or equal hrs_per_day."]})
+        if (("avg_power_W" in kwargs and kwargs["avg_power_W"] is not None) or
```
Hassan754, line 45: Same applies to the rest of similar checks
```suggestion
        if ((kwargs.get("avg_power_W") is not None) or
```

d3a_interface/validators/load_validator.py
```diff
@@ -15,104 +15,113 @@
...
+            ("hrs_per_day" in kwargs and kwargs["hrs_per_day"] is not None) or
+            ("hrs_of_day" in kwargs and kwargs["hrs_of_day"] is not None)) \
```
Hassan754, line 47: If possible get rid of the backslashes for line escaping in context of the refactoring done

Hassan754 (commented):
Left some styling comments, LGTM

Hassan754:
Can you also run d3a tests against this branch? the validators could be used there too

fievelk:
@Hassan754 Thanks for the comments! Most of the code is just copied from the previous implementation, but I can apply your requested changes to improve the style 👍 

> Can you also run d3a tests against this branch? the validators could be used there too

Both d3a and d3a-web have their own PRs for these changes (since they use validators, as you said).

hannesdiedrich (approved):
LGTM, but can you please explain, why the move from functions to classes is a good idea here? I dont see any function being reused, all validators inherit from the BaseValidator which has no function.

fievelk:
> LGTM, but can you please explain, why the move from functions to classes is a good idea here? I dont see any function being reused, all validators inherit from the BaseValidator which has no function.

Sure! First of all, these classes can (and should) be further improved so that they can share code. This means that in the future we could create individual validator classes for the parameters and make them composable. E.g.:
```python
class BaseValidator:
    def __init__(self, validators):
        self.validators = validators

    def validate(self, **kwargs):
        for validator in self.validators:
            validator.validate(**kwargs)

# Then you could create new device validator classes very easily
PVValidator = BaseValidator(validators=[InitialSellingRateValidator, MaxPanelPowerValidator, ...])
HomeMeterValidator = BaseValidator(validators=[InitialSellingRateValidator, InitialBuyingRateValidator, ...])
```
I'm not saying this is what we'll do, but it might be an option.

Moreover: when using individual functions (e.g. `validate_load_device()`, `validate_pv_device()` etc.) you have to explicitly import them and use them, for example inside of their relative strategy classes in d3a. As a consequence, you'll have many repeated lines of code that do the same thing (validating the strategy/device) but can't be deleted, because the name of the function is different every time.

If we turn them into classes and each class shares the same interface, you can refactor the code to have something like:
```python
class HomeMeterStrategy(BaseStrategy):
    def __init__(self, ...):
        self.validator = HomeMeterValidator  # Only this changes depending on the strategy class
```
At this point, you can just use `self.validator.validate(...)` throughout the Strategy. And since this is the same for every strategy, you can refactor code and put it (for example) in the `BaseStrategy` class and avoid code repetition.

If you bring this to the extremes, you could pass the validator as one of the `__init__()` parameters. This is not necessary right now for us, but if you make enough refactoring you could define new strategies simply by passing different parameters to the base strategy. For example:
```python
FridgeStrategy = BaseStrategy(validator=FridgeValidator, on_tick_behaviour=FridgeOnTickBehaviour(), ...)
```
You could in theory do the same with functions, but they're much less flexible and you have no assurance that they implement the same interface.
Of course this is just an example, but the general concept is to re-use code as much as possible and make it combinable.

Hassan754 (approved):
LGTM as the context of this PR is dealing with the architecture of validators, in the future we could probably revisit and refactor the units of the validators line by line

PR #243, opened 2021-07-29
Bug/d3 asim 3544 Fixed bug in SimulationAssetsInfo when accumulating total_energy_demand_kwh

hannesdiedrich:


Hassan754 (dismissed):
Beside back slashes, LGTM

d3a_interface/sim_results/simulation_assets_info.py
```diff
@@ -21,50 +24,52 @@ def merge_results_to_global(market_results: Dict, global_results: Dict, slot_lis
...
 
     def update(self, area_result_dict: Dict, core_stats: Dict, current_market_slot: str):
+        # total_energy_demanded_wh is already accumulated value and should only be replaced
         updated_results_dict = {
-            'number_of_load_type': 0,
-            'total_energy_demand_kwh': self.assets_info['total_energy_demand_kwh'],
-            'number_of_pv_type': 0,
-            'total_energy_generated_kwh': self.assets_info['total_energy_generated_kwh']
+            "number_of_load_type": 0,
+            "total_energy_demand_kwh": 0,
```
fievelk, line 30: Could you explain why/how this change fixes the reported bug?

Moreover, in the ticket I read:
> Secondly, my PV has sold 12.29 kWh, so I would expect its PV, Total Generated to be 12.29 kWh instead of 0 kWh

so I would have expected also `total_energy_generated_kwh` to be modified, but that's not the case. That's why I'd like to know more about the reason why the solution works :)

tests/test_assets_info.py
```diff
@@ -31,11 +31,13 @@ def test_update(self):
             }
         }
         self.assets_info.update({}, core_stats, "")
+        self.assets_info.update({}, core_stats, "")
```
fievelk, line 34: I suppose this line was duplicated by mistake
```suggestion
```

fievelk: Unless it's part of the reason why I didn't quite understand the fix :D

fievelk: It looks like you're calling `.update()` twice, so that the total generated energy is accumulated (as you say in the comment). If this is the case, I would suggest to actually check the value twice. For example:
```python
self.assets_info.update({}, core_stats, "")
expected_res = {
                "number_of_load_type": 3,
                "number_of_pv_type": 2,
                "total_energy_demand_kwh": 0.06,
                "total_energy_generated_kwh": 25
                }
actual_res = {key: self.assets_info.raw_results[key] for key in expected_res.keys()}
assert expected_res == actual_res

# total_energy_demanded_wh is already accumulated value
expected_res = {
                "number_of_load_type": 3,
                "number_of_pv_type": 2,
                "total_energy_demand_kwh": 0.06,
                "total_energy_generated_kwh": 50
                }
actual_res = {key: self.assets_info.raw_results[key] for key in expected_res.keys()}
assert expected_res == actual_res
```

hannesdiedrich: Yes, I call `.update()` twice on purpose. You are right, let me check the values twice, too.

d3a_interface/sim_results/simulation_assets_info.py
hannesdiedrich: I tried to explain it in my comment in Line 27 : `total_energy_demanded_wh is already accumulated value and should only be replaced` , whereas `total_energy_generated_kwh` only applies to the last market slot for the PVs. 
This mismatch is not ideal and if you agree I could add a refactoring story where we homogenise the sent values on the d3a side.

fievelk: Ahh ok, so `total_energy_demanded_wh` is already accumulated (somewhere else in the code), so we just need to set its values as we receive it. Yes, if possible I would vote for creating a ticket for homogenizing it!

d3a_interface/sim_results/simulation_assets_info.py
```diff
@@ -21,50 +24,52 @@ def merge_results_to_global(market_results: Dict, global_results: Dict, slot_lis
             "merge not supported.")
 
     def update(self, area_result_dict: Dict, core_stats: Dict, current_market_slot: str):
+        # total_energy_demanded_wh is already accumulated value and should only be replaced
```
fievelk, line 27: Nit:
```suggestion
        # total_energy_demand_kwh is already accumulated value and should only be replaced
```

fievelk (approved):

Hassan754 (approved):
LGTM

d3a_interface/sim_results/simulation_assets_info.py
hannesdiedrich: Done https://gridsingularity.atlassian.net/browse/D3ASIM-3546

PR #244, opened 2021-08-03
Feature/D3ASIM-3542: Add Validation for BidOfferMatch

Hassan754:


hannesdiedrich (approved):
LGTM

d3a_interface/dataclasses.py
```diff
@@ -1,16 +1,36 @@
...
+"""
+
 from dataclasses import dataclass, field
-from typing import Dict, List
+from typing import List, Dict, Optional
 
 
 @dataclass
 class BidOfferMatch:
+    """Representation of a market match."""
```
fievelk, line 25: Nit: I think it would be good to describe what a "market match" is.

d3a_interface/dataclasses.py
```diff
@@ -1,16 +1,36 @@
...
     market_id: str
     bids: List[Dict]
     selected_energy: float
     offers: List[Dict]
     trade_rate: float
 
-    def serializable_dict(self):
+    def serializable_dict(self) -> Dict:
```
fievelk, line 32: Nit: since we have `from_dict()`, I think this method might be called `to_dict()`.

d3a_interface/dataclasses.py
```diff
@@ -19,6 +39,34 @@ def serializable_dict(self):
...
         }
 
+    @classmethod
+    def from_dict(cls, bid_offer_match: Dict) -> Optional["BidOfferMatch"]:
+        """Receive a serializable dict of BidOfferMatch and return a BidOfferMatch object."""
+        if cls.is_valid_dict(bid_offer_match):
+            return BidOfferMatch(**bid_offer_match)
+
+    @classmethod
+    def is_valid_dict(cls, bid_offer_match: Dict) -> bool:
```
fievelk, line 49: I highly suggest creating a class for this, to respect the single responsibility principle and avoiding code explosion in the future.

fievelk (commented):
One major comment, I'll review again

d3a_interface/dataclasses.py
fievelk: If you want to make the validation even more useful, you could find a way to provide a meaningful error message for each validation error (for example using exceptions). For example it could be used like this:
```python
try:
    BidOfferMatch.from_dict(dictionary)
except BidOfferMatchValidationError as ex:
    logger.error(ex)
```

Hassan754: I don't think creating a validator class for each dataclass makes sense, we will can lots of dataclasses in the future.
Actually I was trying to abstract the validation in the beginning but typing's types made things more complicated.
I suggest either

1. Creating a class responsible for all dataclasses validation (Can't work on none dataclasses)
2. Creating a schema for each class-dataclass and use `jsonschema.validators` `d3a_interface/scenario_validators.py:24`; If The jsonschema cannot cover all cases I may also introduce [Cerberus](https://docs.python-cerberus.org/en/stable/) in our codebase

What do you think ?

fievelk: I don't think we should worry about future dataclasses. We don't know if/how many of them we will need to have. Also, not all of them might need validation. Even if we had 1000 dataclasses, I would personally suggest to create 1000 validator classes: they are not the same thing and they should be kept separate.

> typing's types made things more complicated

What do you mean by this? Do you have an example?

Using schemas might be a good idea: it really depends on what kind of validation you want to achieve. If you want to have more complex validation, I'm not sure that's the best approach. If you just want to validate dictionaries it might be good (especially if these dictionaries are being passed from one service to another).

Hassan754: > What do you mean by this? Do you have an example?
For example we can have List[Dict, str, int] and Dict[Dict[List[somethings]], List], this is just a basic example, I faced several challenges while I was trying to implement this abstract validation
Anw I suppose we can agree on the dedicated validation class for the dataclass, I'll try to figure out the best and cleanest way to do it

Hassan754: @fievelk trying to implement a validator class, the benefit over this current approach is that we would need to add a validate function for each field in the dataclass, each new field will introduce a new function.
This indeed follows the Single responsibility principle and will help saving the current function from linearly growing.
however I feel like in any case this dataclass won't scale up more often, and keeping the validation inside the class itself in this case is reasonable and easier to use instead of decoupling and over-engineer the design
What do you think ?

d3a_interface/dataclasses.py
Hassan754: I agree, however I kept the naming because we have it in all our data classes/classes (Offer, Bid, Trade, TradeInfo ....)

d3a_interface/dataclasses.py
fievelk: If you think this won't scale up, that's fine for me. I just wouldn't want to have a massive validation function once we start adding new information to the `BidOfferMatch` objects, because that would become hard to test and maintain/refactor.

fievelk (approved):

d3a_interface/dataclasses.py
Hassan754: I totally agree, If this happen in the future we can create a design for this

PR #245, opened 2021-08-11
Fix PAC multi markets matching

Hassan754:
Fix a bug when sending multiple markets keys to the matcher
There is a test that should've checked this but there was a wrong expected data values


hannesdiedrich (approved):
LGTM

fievelk (approved):

PR #246, opened 2021-08-12
D3ASIM-3534: Fetching clearing state at the time of get_matches_recom…

faizan2590:
…mendations.


Hassan754:
I'm afraid this will break the interface, the `get_matches_recommendations` of matching algorithms should only return a list of recommendations , Otherwise the caller code will break or we'll have to add conditional statements which isn't clean approach imo
Do you think there is some other way you can tackle this ?

faizan2590:
> I'm afraid this will break the interface, the `get_matches_recommendations` of matching algorithms should only return a list of recommendations , Otherwise the caller code will break or we'll have to add conditional statements which isn't clean approach imo
> Do you think there is some other way you can tackle this ?

https://github.com/gridsingularity/d3a-interface/pull/246/commits/7dad840155947de8623e605bc68840ff90400052

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
```diff
@@ -119,48 +118,46 @@ def _clearing_point_from_supply_demand_curve(bids_rate_energy: Dict, offers_rate
...
                 ascending_rate_bids, cumulative_offers)
         elif ConstSettings.IAASettings.PAY_AS_CLEAR_AGGREGATION_ALGORITHM == 2:
             cumulative_bids = self._discrete_point_curve(self.sorted_bids, math.floor)
             cumulative_offers = self._discrete_point_curve(self.sorted_offers, math.ceil)
-            max_rate = self._populate_market_cumulative_offer_and_bid(cumulative_bids,
-                                                                      cumulative_offers)
-            clearing = self._get_clearing_point(max_rate)
+            max_rate, cumulative_bids, cumulative_offers = \
+                self._populate_market_cumulative_offer_and_bid(cumulative_bids, cumulative_offers)
+            clearing = self._get_clearing_point(max_rate, cumulative_bids, cumulative_offers)
```
hannesdiedrich, line 140: `_get_clearing_point` returns two values, but I dont see it is being used here. I think you can remove it (`cumulative_bids[rate]`) then. Right?

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
```diff
@@ -119,48 +118,46 @@ def _clearing_point_from_supply_demand_curve(bids_rate_energy: Dict, offers_rate
...
+        clearing, cumulative_bids, cumulative_offers = None, None, None
 
         if len(self.sorted_bids) == 0 or len(self.sorted_offers) == 0:
             return
 
         if ConstSettings.IAASettings.PAY_AS_CLEAR_AGGREGATION_ALGORITHM == 1:
             cumulative_bids = self._accumulated_energy_per_rate(self.sorted_bids)
             cumulative_offers = self._accumulated_energy_per_rate(self.sorted_offers)
-            self.state.cumulative_bids = cumulative_bids
-            self.state.cumulative_offers = cumulative_offers
```
hannesdiedrich, line 134: To repeat the offline discussion we had yesterday: Why not using the `self.state` as it was used before. With this you can omit passing the  `cumulative_bids` and `cumulative_offers` to `_get_clearing_point` and access them via `self.state` in there.
Or am I missing something here? If this needs to be done the way you implemented, please remove the self.state from the class because I dont see it being used anywhere else, right?

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
faizan2590: Its saved to `clearing ` as tuple i.e first as clearing rate and second as clearing energy. Its been like that before this PR too.

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
faizan2590: `_get_clearing_point ` would be triggered multiple times within a slot, and not everytime there would be a clearing but still saving uncleared `cumulative_bids /cumulative_offers ` to state would lead to increase in memory which is also not really needed for the supply/demand curve. Regarding removing self.state from the class, is simply an organisation of markets clearing related data like, clearing rate, clearing energy, supply curve and demand cruve; and these data used for plotting the backend supply/demand curve.

hannesdiedrich: Ok, I guess I was confused by the fact that the `cumulative_bids`, `cumulative_offers` where passed around in the methods in this class while there should be also in the `state` member. I would have reused this one, but this is probably a matter of taste.

hannesdiedrich (commented):
As always, please add docstrings and preferably typing to the changed methods. Once this is done, LGTM

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
```diff
@@ -119,48 +122,48 @@ def _clearing_point_from_supply_demand_curve(bids_rate_energy: Dict, offers_rate
...
-                                                                      cumulative_offers)
-            clearing = self._get_clearing_point(max_rate)
+            max_rate, cumulative_bids, cumulative_offers = \
+                self._populate_market_cumulative_offer_and_bid(cumulative_bids, cumulative_offers)
+            clearing = self._get_clearing_point(max_rate, cumulative_bids, cumulative_offers)
         if clearing is not None:
-            self.state.clearing[current_time] = clearing[0]
+            self.state.cumulative_bids[market_id] = {current_time: cumulative_bids}
+            self.state.cumulative_offers[market_id] = {current_time: cumulative_offers}
+            self.state.clearing[market_id] = {current_time: clearing}
```
fievelk, line 150: The `market_id` and these lines do not seem to belong to this method, as they don't have any effect on returning the `clearing_point` that you're returning. Also, they break the [Command-Query separation principle](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation).
I would recommend moving them into a new method and call it from `get_matches_recommendations()`.

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
```diff
@@ -53,7 +54,7 @@ def get_matches_recommendations(self, matching_data):
             bids = data.get("bids")
             offers = data.get("offers")
             current_time = data.get("current_time")
-            clearing = self.get_clearing_point(bids, offers, current_time)
+            clearing = self.get_clearing_point(bids, offers, current_time, market_id)
```
fievelk, line 57: What is `market_id`? Is it a market slot ID, an asset ID, or something else?

faizan2590: Its an area market slot id

fievelk: I'm more confused than before :D What's an "area market slot"? How's it different from a normal market slot?

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
```diff
@@ -83,13 +84,15 @@ def _smooth_discrete_point_curve(obj, limit, asc_order=True):
                 obj[i] = obj.get(i, 0) + obj.get(i + 1, 0)
         return obj
 
-    def _get_clearing_point(self, max_rate):
+    @staticmethod
+    def _get_clearing_point(max_rate: int, cumulative_bids: OrderedDict,
+                            cumulative_offers: OrderedDict) -> (int, float):
+        """Gets the energy rate and cumulative energy at the point of equilibrium."""
```
fievelk, line 90: Nit:
```suggestion
        """Get the energy rate and cumulative energy at the point of equilibrium."""
```

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
```diff
@@ -83,13 +84,15 @@ def _smooth_discrete_point_curve(obj, limit, asc_order=True):
...
         for rate in range(1, max_rate + 1):
-            if self.state.cumulative_offers[rate] >= self.state.cumulative_bids[rate]:
-                if self.state.cumulative_bids[rate] == 0:
-                    return rate-1, self.state.cumulative_offers[rate-1]
-                else:
-                    return rate, self.state.cumulative_bids[rate]
+            if cumulative_offers[rate] >= cumulative_bids[rate]:
+                if cumulative_bids[rate] == 0:
+                    return rate-1, cumulative_offers[rate-1]
```
fievelk, line 94: Could you explain what this does? Why is it `rate - 1`?

fievelk:
General:
- It would be good to unit-test some methods like `get_clearing_point`;
- please, add lots of documentation in this module. It takes a while to understand what's going on even in functions that sound easier to understand (e.g. `_accumulated_energy_per_rate`).

Nits:
- imports are not optimized;
- please replace `\` with `()` and hanging indentation.

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
faizan2590: It's a `market slot ID`, I tried to be explicit in my last comment but apparently confuses you even more. What I meant was its area's market's slot ID

hannesdiedrich: We need to decide on this first IMO in order to not fall into the same pit as with the single and double quotes. I dont remember to rename `market` to `market_slot` . Until we decided on this, I vote vor keeping it consistently the old way where the `market_id` is the id of the multiple markets the area has (in the time dimension).

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
hannesdiedrich: Also, why do we need to save the `current_time` in there. Isn't the `market_id` a sufficient unique identifier for the market?
Sorry that I only ask this now.

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
fievelk: @hannesdiedrich ok, let's postpone the conversation to the moment when we start the refactoring then.

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
spyrostz: This needs familiarity with the 2 sided PAC discrete point curve algorithm to understand. Bids and offers are sorted according to their energy rate, each creating a series of lines in a 2-dimensional graph. This method traverses this graph by increments of 1 cents / kWh , and tries to find the intersection of the offer and bid lines (at the point when the offer becomes greater than the bid line, that's when we know that the lines intersect). This is a special case that, in reality, will appear only if bids and offers are both zero at one energy rate, meaning that the energy needs to be sold for zero rate. In this case we select the energy rate just before the last one, in order to not trade energy for free.

spyrostz:
> General:
> 
>     * It would be good to unit-test some methods like `get_clearing_point`;
> 
>     * please, add lots of documentation in this module. It takes a while to understand what's going on even in functions that sound easier to understand (e.g. `_accumulated_energy_per_rate`).
> 
> 
> Nits:
> 
>     * imports are not optimized;
> 
>     * please replace `\` with `()` and hanging indentation.

I think that documentation/unit tests are not really necessary for these methods, because the algorithm that these methods implement is more or less deprecated (have a look at PAY_AS_CLEAR_AGGREGATION_ALGORITHM constant) and it really is not worth it for code that will be deprecated at some point anyways. For the rest I agree

spyrostz (approved):
LGTM once the rest of the comments are dealt with

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
faizan2590: > Also, why do we need to save the current_time in there. Isn't the market_id a sufficient unique identifier for the market?
Sorry that I only ask this now.

There can be multiple clearings in a single market. Hope it answers your question.

faizan2590: > The market_id and these lines do not seem to belong to this method, as they don't have any effect on returning the clearing_point that you're returning (and they also break the Command-Query separation principle).
I would recommend moving them into a new method and call it from get_matches_recommendations()

Moving it to a new method and calling it from `get_matches_recommendations` requires `get_clearing_point` to return additional values like `cumulative_offers` & `cumulative_bids` which violates the method name from which it expects only to return a clearing point. As target of this *bug ticket* was to solve a bug, not refactoring, I would rather create a ticket to think of better refactoring.

fievelk (approved):

PR #251, opened 2021-08-30
D3ASIM-3492: Adding convert_datetime_to_ui_str_format utility method.

faizan2590:


d3a_interface/utils.py
```diff
@@ -482,3 +484,8 @@ def sort_list_of_dicts_by_attribute(input_list: List[Dict],
         return sorted(
             input_list,
             key=lambda obj: obj.get(attribute))
+
+
+def convert_datetime_to_ui_str_format(data_time):
```
fievelk, line 489: What's the reason to implement this method on the backend side, and not on the frontend side?

faizan2590: I followed the approach we have been using so far that is to pre-convert datetime to ui formatted string, for reference, have a look at similar method `convert_datetime_to_str_in_list`

d3a_interface/utils.py
```diff
@@ -482,3 +484,7 @@ def sort_list_of_dicts_by_attribute(input_list: List[Dict],
...
+    return data_time.strftime("%m-%d-%Y, %H:%M:%S")
```
spyrostz, line 490: I assume that this format was something that product / frontend teams requested, right ? In this case I would create a constant value that exactly refers to where this will be used (e.g. DATE_TIME_UI_SAVINGS_KPI_FORMAT). Reason that I insist is the existence of this variable DATE_TIME_UI_FORMAT , I do not mind that much however having one hardcoded format and another format in a variable, both for the UI, will cause confusion.

spyrostz (approved):
One comment, LGTM once resolved

d3a_interface/utils.py
fievelk: Nit: just as an idea: why don't we just decide on one unique format for the entire backend stack? The frontend team (and even us, for logging) could then reformat it as needed.

faizan2590: > I assume that this format was something that product / frontend teams requested, right ?

They haven't explicitly requested any format, I was using the same format as being used  for UI conversion  and as you referenced via `DATE_TIME_UI_FORMAT="MMMM DD YYYY, HH:mm [h]"` using a utility function but since `'datetime.datetime' object has no attribute 'format'` , therefore couldn'T used the old utility method(`format_datetime`) so I created another utility function to have similar datetime_str for UI.

fievelk: pendulum's `DateTime` objects inherit from `datetime.datetime`, so you can use `datetime.datetime`'s methods on them. This means that we can just use the default datetime format also with pendulum's dates:
```python
DATE_TIME_UI_FORMAT = "%B %d %Y, %H:%M h"
pendulum.datetime(2021, 12, 21).strftime(DATE_TIME_UI_FORMAT)
# 'December 21 2021, 00:00 h'
```
Or you can convert datetime objects into pendulum `DateTime` objects, and then format them as before:
```python
date = datetime.datetime.now()
DATE_TIME_UI_FORMAT = "MMMM DD YYYY, HH:mm [h]"
pendulum.instance(date).format(DATE_TIME_UI_FORMAT)
# 'August 31 2021, 10:42 h'
```

tests/test_utils.py
```diff
@@ -45,3 +46,8 @@ def test_sort_list_of_dicts_by_attribute(self):
         assert [4, 2, 1, 3] == [data["id"] for data in output_list]
         output_list = sort_list_of_dicts_by_attribute(input_list, "price", reverse_order=True)
         assert [3, 1, 2, 4] == [data["id"] for data in output_list]
+
+    def test_convert_datetime_to_ui_str_format(self):
+        current_time = datetime.now()
+        current_time_str = convert_datetime_to_ui_str_format(current_time)
+        assert type(current_time_str) is str
```
fievelk, line 53: Sorry for noticing this now: why don't we test the actual output? I think it would be good to pass a specific date and test the full output, not just the type.

d3a_interface/utils.py
spyrostz: IIRC we have decided a long time ago to use pendulum exclusively for datetime formats, and I think that one of the major reasons was the fact that the datetime format of the standard library is much more difficult to read on first glance compared to pendulum's (@hannesdiedrich correct me if I am wrong and I do not recall our discussion correctly). Thus I would vote for the second approach, since the implementation of `instance` is only some type checks and a casting to the pendulum type.

fievelk: Ok, if pendulum is given for granted, then the second solution sounds good 😄

spyrostz (approved):
LGTM

hannesdiedrich (approved):
LGTM

PR #253, opened 2021-08-31
D3ASIM-3556: Export market structures to d3a-interface

Hassan754:


tests/test_dataclasses.py
```diff
@@ -77,3 +93,476 @@ def test_from_dict():
         assert bid_offer_match.offers == expected_dict["offers"]
         assert bid_offer_match.selected_energy == expected_dict["selected_energy"]
         assert bid_offer_match.trade_rate == expected_dict["trade_rate"]
+
+
+class TestBaseBidOffer:
+
+    def setup_method(self):
+        self.initial_data = {
+            "id": uuid.uuid4(),
```
faizan2590, line 102:
```suggestion
            "id": str(uuid.uuid4()),
```

d3a_interface/data_classes.py
```diff
@@ -0,0 +1,442 @@
...
+
+    @classmethod
+    def from_json(cls, offer_or_bid, current_time=None) -> Union["Offer", "Bid"]:
+        offer_bid_dict = json.loads(offer_or_bid)
+        object_type = offer_bid_dict.pop("type")
+
+        offer_bid_dict.pop("energy_rate", None)
+
+        if object_type == "Offer":
+            offer_bid_dict["time"] = current_time
```
faizan2590, line 92: Why time only needs to be injected for Offer and not for Bid ?

d3a_interface/data_classes.py
```diff
@@ -0,0 +1,442 @@
...
+            return Offer(**offer_bid_dict)
+        if object_type == "Bid":
+            return Bid(**offer_bid_dict)
+
+
+class Offer(BaseBidOffer):
+    def __init__(self, id: str, time: datetime, price: float,
+                 energy: float, seller: str, original_price: Optional[float] = None,
+                 seller_origin: str = None, seller_origin_id: str = None,
+                 seller_id: str = None,
```
faizan2590, line 102: These parameters should not have a default None and must always be provided

d3a_interface/data_classes.py
```diff
@@ -0,0 +1,442 @@
...
+                 ):
+        super().__init__(id=id, time=time, price=price, energy=energy,
+                         original_price=original_price,
+                         attributes=attributes, requirements=requirements)
+        self.buyer = buyer
+        self.buyer_origin = buyer_origin
+        self.buyer_origin_id = buyer_origin_id
+        self.buyer_id = buyer_id
+
+    def __hash__(self) -> int:
```
faizan2590, line 176: Where do you need it?

tests/test_dataclasses.py
```diff
@@ -77,3 +93,476 @@ def test_from_dict():
...
+        assert offer.price == self.initial_data["price"]
+        assert offer.energy == self.initial_data["energy"]
+        assert offer.original_price == self.initial_data["original_price"]
+        assert offer.energy_rate == self.initial_data["price"] / self.initial_data["energy"]
+        assert offer.attributes == self.initial_data["attributes"]
+        assert offer.requirements == self.initial_data["requirements"]
+        assert offer.seller == "seller"
+        assert offer.seller_id is None
+        assert offer.seller_origin is None
+        assert offer.seller_origin_id is None
```
faizan2590, line 217: These should be a non None

d3a_interface/data_classes.py
```diff
@@ -0,0 +1,442 @@
...
+                self.price == other.price and
+                self.original_price == other.original_price and
+                self.energy == other.energy and
+                self.seller == other.seller and
+                self.seller_origin_id == other.seller_origin_id and
+                self.attributes == other.attributes and
+                self.requirements == other.requirements)
+
+    def csv_values(self) -> Tuple:
+        rate = round(self.energy_rate, 4)
```
faizan2590, line 144: rounding is done in many places with same decimal places, better to replace it with a common rounding constant variable

d3a_interface/data_classes.py
```diff
@@ -0,0 +1,442 @@
...
+        self.seller_origin = seller_origin
+        self.seller_origin_id = seller_origin_id
+        self.seller_id = seller_id
+
+    def __hash__(self) -> int:
+        return hash(self.id)
+
+    def __repr__(self) -> str:
+        return ("<Offer('{s.id!s:.6s}', '{s.energy} kWh@{s.price}', '{s.seller} {rate}'>"
+                .format(s=self, rate=self.energy_rate))
```
faizan2590, line 118: Although its more like a migration from d3a to d3a-interface BUT still how about using f-string now

faizan2590 (commented):
Left some comments, will give another look once addressed/answered.

d3a_interface/data_classes.py
```diff
@@ -0,0 +1,442 @@
...
+
+from copy import deepcopy
+from dataclasses import dataclass, field, asdict
+from typing import List, Dict, Optional, Tuple, Union
+from pendulum import DateTime, parse
+
+from d3a_interface.utils import datetime_to_string_incl_seconds, key_in_dict_and_not_none
+
+
+def my_converter(date_obj: DateTime):
```
fievelk, line 29: Nit: I believe this should be renamed.

d3a_interface/data_classes.py
Hassan754: I'm not sure, there are dozens of calls to Bid/ Offer instantiation (especially in tests) and they were always omitted
I have checked and it looks like in the production code, these are always sent
Basically I agree that it makes sense to not allow these to be None, however this will require changes to all unit tests, do you think these can be postponed to another time?

d3a_interface/data_classes.py
```diff
@@ -0,0 +1,442 @@
...
+    @staticmethod
+    def from_json(trade_bid_offer_info: str) -> "TradeBidOfferInfo":
+        info_dict = json.loads(trade_bid_offer_info)
+        return TradeBidOfferInfo(**info_dict)
+
+
+class Trade:
+    def __init__(self, id: str, time: datetime, offer_bid: Union[Offer, Bid],
+                 seller: str, buyer: str, residual: Optional[Union[Offer, Bid]] = None,
+                 already_tracked: bool = False,
```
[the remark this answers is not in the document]
Hassan754: From the official documentation of Optional:

> Optional[X] is equivalent to Union[X, None].
So this is not relevant here

d3a_interface/data_classes.py
Hassan754: This can be useful when trying to set an offer/bid or any other market structure as a dictionary key, it will know how to hash the object

d3a_interface/data_classes.py
Hassan754: They are used 3 times with a value of 4 for the energy_rate, however I do not think that we should add a constant for them (at least not a global constant ) we may add a constant at the top of the file but The way I see it, this is not critical and would restrict us from having different decimal places for the existing different round calls
Anyway I could add the constant if you believe it's worth it

d3a_interface/data_classes.py
spyrostz: Should be for both the same. However keep in mind that this is probably the time in string format, json.loads is not able to deserialize datetimes.

d3a_interface/data_classes.py
spyrostz: Please move it to the base class, this applies to all child classes.

spyrostz (dismissed):
LGTM once comments are resolved. I was thinking that it could be worth it to make the base class an ABC in order to know which methods need to be implemented if we plan to add another similar class in the future, however please do not do it because 1. I do not foresee it  2. will unnecessary complicate the existing classes' implementation  3. this PR is urgent

d3a_interface/data_classes.py
Hassan754: @spyrostz for some reason when I implement __hash__() in the parent class, the children are not inheriting it, I'm still investigating why 
But it's probably due to the fact the that we're overriding the __eq__ in children classes https://stackoverflow.com/a/2134487

spyrostz: I see, in this case LGTM

Hassan754:
> LGTM once comments are resolved. I was thinking that it could be worth it to make the base class an ABC in order to know which methods need to be implemented if we plan to add another similar class in the future, however please do not do it because 1. I do not foresee it 2. will unnecessary complicate the existing classes' implementation 3. this PR is urgent

Thanks a lot!
There is always extra room for improvements 😫

d3a_interface/data_classes.py
Hassan754: @faizan2590 seems great, done

spyrostz (approved):
LGTM

d3a_interface/data_classes.py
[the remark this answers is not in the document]
faizan2590: Sorry, didn'T quite get you, here `already_tracked ` isn't a union with None. Can you please explain a bit more.

d3a_interface/data_classes.py
faizan2590: Okay, please make a ticket to be in radar that this is something to be dealt with.

d3a_interface/data_classes.py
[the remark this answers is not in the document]
Hassan754: the already_tracked's default value is False, so it's either True or False -> bool
Optional can be added if it can also be None, but it doesn't

d3a_interface/data_classes.py
Hassan754: I'll keep the same ticket and move it to SFD after this is merged, and I'll update its description to include more refactorings in the future (I already have some in mind)

d3a_interface/data_classes.py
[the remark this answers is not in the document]
fievelk: @faizan2590 maybe the misunderstanding is related to the name "Optional". In type hints, "Optional" doesn't mean that the parameter can be omitted when calling the function. It means that the value of the parameter can be `None` in some cases. In this situation (as Hassan explained) the `already_tracked` parameter can't be `None`, so using `Optional` would not be correct.

d3a_interface/data_classes.py
faizan2590: We have a similar [round_floats_for_ui](https://github.com/gridsingularity/d3a-interface/blob/master/d3a_interface/utils.py#L302). So, could use it here too, off-course its rounding upto 3 decimal points.

Anyways non-blocking

tests/test_dataclasses.py
faizan2590: https://github.com/gridsingularity/d3a-interface/pull/253/files?file-filters%5B%5D=.py&hide-deleted-files=true#r708031409

d3a_interface/data_classes.py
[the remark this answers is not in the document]
faizan2590: Thank You both for explanation, got it now

d3a_interface/data_classes.py
faizan2590: > I'll keep the same ticket and move it to SFD after this is merged, I'll update its description to include more refactorings in the future (I already have some in mind)

If the description of ticket needs to be updated at the time of review, it means thing weren't planned thought of before starting and its something additional, so I would vote to move this ticket to done once this PR is merged, and create a follow-up ticket targeting further enhancements.

Tagging @fievelk @spyrostz in case they don't agree with me .

faizan2590 (approved):
LGTM

d3a_interface/data_classes.py
Hassan754: I'm okay with creating a new ticket, but it will also be a broad : refactor market structures
THe original tickets did not explicitly describe the needed changes indeed, it's a `refactoring` ticket so I resort to my instincts on the go while refactoring, but the majority of changes were mentioned in an older ticket (mostly comments from Pierpaolo) + I added some more changes and left some for the future

fievelk: Since you want to first merge this PR, and then create a new one to address this specific comment, I would say that we can safely open a new ticket. It will also help to track the changes with the commit messages' prefixes.

PR #256, opened 2021-09-02
D3ASIM-3567: Added more supported datetime formats to the str_to_pend…

spyrostz:
…ulum_datetime method.


hannesdiedrich (approved):
LGTM

d3a_interface/utils.py
```diff
@@ -185,7 +185,14 @@ def str_to_pendulum_datetime(input_str):
...
             pendulum_time = from_format(input_str, DATE_TIME_FORMAT)
         except ValueError:
-            raise Exception(f"Format is not one of ('{TIME_FORMAT}', '{DATE_TIME_FORMAT}')")
+            try:
+                pendulum_time = from_format(input_str, DATE_TIME_FORMAT_SECONDS)
+            except ValueError:
+                try:
+                    pendulum_time = from_format(input_str, TIME_FORMAT_SECONDS)
+                except ValueError:
+                    raise Exception(f"Format is not one of ('{TIME_FORMAT}', '{DATE_TIME_FORMAT}'"
```
Hassan754, line 194: This is linearly scaling up for each new format we add, I would suggest using a different approach like the following:
```python
supported_formats = iter([TIME_FORMAT, DATE_TIME_FORMAT_SECONDS, TIME_FORMAT_SECONDS])
pendulum_time = None
for format in supported_formats:
    try:
        pendulum_time = from_format(input_str, format)
        break
    except ValueError:
        continue
if not  pendulum_time:
    raise Exception(f"Format is not ....")
```

Hassan754 (approved):
Left 1 non blocking comment

d3a_interface/utils.py
spyrostz: Agreed, it is much better this way! Fixing this

Hassan754 (approved):

tests/test_utils.py
```diff
@@ -51,3 +51,16 @@ def test_convert_datetime_to_ui_str_format(self):
...
+
+    def test_str_to_pendulum_datetime(self):
+        datetime_obj = datetime(year=2021, month=4, day=5, hour=12, minute=30)
+        assert str_to_pendulum_datetime("2021-04-05T12:30") == datetime_obj
+        assert str_to_pendulum_datetime("2021-04-05T12:30:00") == datetime_obj
+        datetime_obj = today(tz="UTC")
+        datetime_obj = datetime_obj.set(hour=12, minute=30)
+        assert str_to_pendulum_datetime("12:30") == datetime_obj
+        assert str_to_pendulum_datetime("12:30:00") == datetime_obj
+        with pytest.raises(Exception):
```
Hassan754, line 63: Nit, it's better if we explicitly check the `ValueError` exception instead of a broad Exception

spyrostz: The code (both original and updated) raise an Exception, not a ValueError. ValueError is the exception that is emitted by from_format and caught in the method

Hassan754: Ah you're right, LGTM then

Hassan754 (approved):

faizan2590 (approved):
LGTM

PR #259, opened 2021-09-14
D3ASIM-3533: Add from_dict method to Offer and Bid

Hassan754:


spyrostz (dismissed):
During the move of the files from d3a, and the original source code was not modified, right (I mean apart from the separate commits which I was able to review) ? From a brief glance I think that the code is the same that's why I LGTMed it already, otherwise ping me to give a closer look at the changes in the requirements_validators.

Hassan754:
> During the move of the files from d3a, and the original source code was not modified, right (I mean apart from the separate commits which I was able to review) ? From a brief glance I think that the code is the same that's why I LGTMed it already, otherwise ping me to give a closer look at the changes in the requirements_validators.

Yep it's almost the same, In general I added two new method `is_offer_requirement_satisfied` `is_bid_requirement_satisfied` that check a single requirement is satisfied + I added unit tests

spyrostz:
> > During the move of the files from d3a, and the original source code was not modified, right (I mean apart from the separate commits which I was able to review) ? From a brief glance I think that the code is the same that's why I LGTMed it already, otherwise ping me to give a closer look at the changes in the requirements_validators.
> 
> Yep it's almost the same, In general I added two new method `is_offer_requirement_satisfied` `is_bid_requirement_satisfied` that check a single requirement is satisfied + I added unit tests



> > During the move of the files from d3a, and the original source code was not modified, right (I mean apart from the separate commits which I was able to review) ? From a brief glance I think that the code is the same that's why I LGTMed it already, otherwise ping me to give a closer look at the changes in the requirements_validators.
> 
> Yep it's almost the same, In general I added two new method `is_offer_requirement_satisfied` `is_bid_requirement_satisfied` that check a single requirement is satisfied + I added unit tests

Thanks, reviewed these 2 methods again, LGTM still stands !

d3a_interface/data_classes.py
```diff
@@ -131,6 +132,24 @@ def serializable_dict(self) -> Dict:
                 "seller_id": self.seller_id,
                 }
 
+    @staticmethod
+    def from_dict(offer: Dict) -> "Offer":
+        """Deserialize an offer dict."""
+
+        return Offer(
+            **{"id": offer["id"],
```
hannesdiedrich, line 140: Nit: Why not `Offer(id=offer["id"], ...)` ?

hannesdiedrich (dismissed):
One minor question, once answered, LGTM

d3a_interface/data_classes.py
Hassan754: Right, adapted

d3a_interface/data_classes.py
```diff
@@ -197,6 +215,24 @@ def serializable_dict(self) -> Dict:
...
+
+        return Bid(
+            id=bid["id"],
+            time=str_to_pendulum_datetime(bid["time"]),
+            energy=bid["energy"],
+            price=bid["energy"] * bid["energy_rate"],
+            original_price=bid.get("original_price"),
+            buyer=bid.get("buyer"),
+            buyer_origin=bid.get("buyer_origin"),
+            buyer_origin_id=bid.get("buyer_origin_id"),
```
faizan2590, line 230:
```suggestion
            buyer_origin=bid["buyer_origin"],
            buyer_origin_id=bid["buyer_origin_id"],
```
Can you be a bit strict here, to guarantee its there and not being overlooked.

Hassan754: @faizan2590 I'm afraid this is not possible at the moment, this will require updating all references (mostly in tests) to pass this variable, which is out of the context of this PR imo

d3a_interface/data_classes.py
```diff
@@ -131,6 +132,23 @@ def serializable_dict(self) -> Dict:
...
+            time=str_to_pendulum_datetime(offer["time"]),
+            energy=offer["energy"],
+            price=offer["energy"] * offer["energy_rate"],
+            original_price=offer.get("original_price"),
+            seller=offer.get("seller"),
+            seller_origin=offer.get("seller_origin"),
+            seller_origin_id=offer.get("seller_origin_id"),
+            seller_id=offer.get("seller_id"),
+            attributes=offer.get("attributes"),
+            requirements=offer.get("requirements"))
```
fievelk, line 150: In case the `offer` dictionary isn't correct, this would raise a `KeyError` exception.
I'd suggest catching possible `KeyError`s and raise a new exception (e.g. `OfferDeserializationError`) with a more informative message.
Also, I would really encourage you to modify the `str_to_pendulum_datetime` function to raise a specific exception (it's currently raising the bare `Exception`). This would allow us to also catch possible errors during the datetime conversion and handle them properly.

fievelk: The same comment applies to the `Bid.from_dict()` below.

d3a_interface/matching_algorithms/requirements_validators.py
```diff
@@ -0,0 +1,205 @@
...
+"""
+
+
+from abc import ABC, abstractmethod
+from typing import Dict, Optional, Union
+
+from d3a_interface.data_classes import Offer, Bid
+
+
+class Requirement(ABC):
```
fievelk, line 23: Nit: since "requirement" is a term that can be used in many different contexts, what do you think of renaming it? My first suggestion would be `OrderRequirement`, since this would refer to both bids and offers as discussed on Slack.

d3a_interface/matching_algorithms/requirements_validators.py
```diff
@@ -0,0 +1,205 @@
...
+        """Check whether a requirement is satisfied."""
+
+
+class TradingPartnersRequirement(Requirement):
+    """Check if trading_partners requirement is satisfied for both bid and offer."""
+
+    @classmethod
+    def is_satisfied(cls, offer: Offer, bid: Bid, requirement: Dict,
+                     clearing_rate: Optional[float] = None,
+                     selected_energy: Optional[float] = None) -> bool:
```
fievelk, line 41: I'm a bit confused by this structure. This class is a `Requirement`, but this method has an argument named `requirement`. I get that they're different objects, but the terminology is a bit confusing and it's hiding this difference, making it difficult to understand what's going on.

Besides that: pylint complains that this method has "too many arguments". It might be a hint for a possible refactoring. Also: please make sure to add docstrings to all public methods (we discussed this in Lisbon and then on slack) :)

d3a_interface/matching_algorithms/requirements_validators.py
```diff
@@ -0,0 +1,205 @@
...
+
+SUPPORTED_BID_REQUIREMENTS = {
+    "trading_partners": TradingPartnersRequirement,
+    "energy_type": EnergyTypeRequirement,
+    "energy": SelectedEnergyRequirement,
+    "price": PriceRequirement
+}
+
+
+class RequirementsSatisfiedChecker:
```
fievelk, line 119: Where is this class used? I currently only see it in the myco-api-client, so I think it should be possible to move it there (without keeping it in d3a-interface). What do you think?

d3a_interface/matching_algorithms/requirements_validators.py
Hassan754: I also recall that we agreed not to make a big deal of doctsrings that only show `what is done`
Anyway, I did add a docstring for it in the base class interface method `is_satisfied` so I don't think that I should add docstring each time I implement this method :)

Furthermore, it's true that the class is `Requirement` but the method is abstract, so we do need to pass a requirement to check whether it's satisfied or not

d3a_interface/matching_algorithms/requirements_validators.py
Hassan754: It's also used in the d3a, this is why I kept it here

Hassan754: https://github.com/gridsingularity/d3a/pull/1227 
I have updated the references in this PR

d3a_interface/matching_algorithms/requirements_validators.py
fievelk: In fact, `clearing_rate` and `selected_energy` are not used for this method, nor for any of the other `Requirement` classes. The only one that uses it is `PriceRequirement`, but I think that adding these parameters to all other classes just for this reason isn't the best option. I have the feeling that we could still improve this interface.

d3a_interface/matching_algorithms/requirements_validators.py
```diff
@@ -0,0 +1,205 @@
...
+    @classmethod
+    def is_satisfied(cls, offer: Offer, bid: Bid, requirement: Dict,
+                     clearing_rate: Optional[float] = None,
+                     selected_energy: Optional[float] = None) -> bool:
+        bid_required_energy = requirement.get("energy")
+        assert isinstance(bid_required_energy, (int, float)), \
+            f"Invalid data type for energy {requirement}"
+        assert isinstance(selected_energy, (int, float)), \
+            f"Invalid data type for selected_energy {selected_energy}"
+        # bid_required_energy is None or it is defined & equals selected_energy -> true
```
fievelk, line 82: If `bid_required_energy` is `None` the method will raise an `AssertionError`. Should this comment be updated?

d3a_interface/matching_algorithms/requirements_validators.py
```diff
@@ -0,0 +1,205 @@
...
+        """Check whether at least 1 offer and 1 bid requirement are satisfied."""
+
+        offer = Offer.from_dict(offer) if isinstance(offer, dict) else offer
+        bid = Bid.from_dict(bid) if isinstance(bid, dict) else bid
+
+        if offer.requirements:
+            offer_requirement_satisfied = cls.are_offer_requirements_satisfied(
+                offer, bid, clearing_rate, selected_energy)
+        else:
+            # If concerned offer have no requirements, consider it as satisfied.
```
fievelk, line 137: Nit:
```suggestion
            # If the concerned offer has no requirements, consider them as satisfied.
```

d3a_interface/matching_algorithms/requirements_validators.py
```diff
@@ -0,0 +1,205 @@
...
+            offer_requirement_satisfied = True
+
+        if bid.requirements:
+            bid_requirement_satisfied = cls.are_bid_requirements_satisfied(
+                offer, bid, clearing_rate, selected_energy)
+        else:
+            # If concerned bid have no requirements, consider it as satisfied.
+            bid_requirement_satisfied = True
+
+        return offer_requirement_satisfied and bid_requirement_satisfied
```
fievelk, line 147: Nit, but since in this code we're differentiating between individual and total requirements:
```suggestion
        return offer_requirements_satisfied and bid_requirements_satisfied
```

tests/test_dataclasses.py
```diff
@@ -259,6 +259,12 @@ def test_serializable_dict(self):
             "seller_id": offer.seller_id,
         }
 
+    def test_from_dict(self):
```
fievelk, line 262: Please also add tests of failing examples (e.g. when you pass a dict with wrong keys/values).

tests/test_dataclasses.py
```diff
@@ -361,6 +367,12 @@ def test_serializable_dict(self):
             "buyer_id": bid.buyer_id,
         }
 
+    def test_from_dict(self):
```
fievelk, line 370: Same as https://github.com/gridsingularity/d3a-interface/pull/259/files#r723053938

d3a_interface/matching_algorithms/requirements_validators.py
Hassan754: Thanks for spotting that, a None value should raise an assertion error so I will adapt it accordingly

tests/test_matching_algorithms/test_requirements_validators.py
```diff
@@ -0,0 +1,120 @@
...
+
+@pytest.fixture
+def bid():
+    return Bid("bid_id", now(), 9, 10, "B", 9)
+
+
+class TestRequirementsValidator:
+    """Test all bid/offer requirements validators."""
+
+    def test_trading_partners_requirement(self, offer, bid):
```
fievelk, line 26: It would be good to split all these tests into separate test cases. At the moment, you're grouping tests for failures with tests for success. Ideally, each scenario should be described by just one test.
I would suggest to create a class for each individual requirement class (e.g. `TestTradingPartnersRequirement`) and separate individual methods (e.g. `test_is_satisfied_succeeds`, `test_is_satisfied_<condition>_fails`).

Even without going so much into the granular level, it would still be better to split these tests a bit.

tests/test_matching_algorithms/test_requirements_validators.py
```diff
@@ -0,0 +1,120 @@
...
+            offer, bid, requirement, clearing_rate=8, selected_energy=1) is False
+        requirement = {"price": 8}
+        bid.requirements = [requirement]
+        assert PriceRequirement.is_satisfied(
+            offer, bid, requirement, clearing_rate=8, selected_energy=1) is True
+
+    @patch(
+        "d3a_interface.matching_algorithms.requirements_validators."
+        "TradingPartnersRequirement.is_satisfied", MagicMock())
+    def test_requirements_validator(self, offer, bid):
```
fievelk, line 113: I'm not sure about this method: what is this testing?

d3a_interface/matching_algorithms/requirements_validators.py
Hassan754: A requirement needs extra context which consists of both the offer / bid, the selected energy between them and the decided on clearing rate
I made the interface accept them all because I suspect these are the only stakeholders at the moment
In general I'm okay with further refactoring the interface

tests/test_matching_algorithms/test_requirements_validators.py
fievelk: Since unit tests focus on a single function at a time, I usually recommend following a specific structure for naming and testing. Each method/function should be tested, so the name of the test could be something like: `test_<function-name>_<condition>_<expected-output>`. Following this makes it easier to create one (or more) tests for each individual function.

d3a_interface/matching_algorithms/requirements_validators.py
```diff
@@ -0,0 +1,205 @@
...
+    @staticmethod
+    def is_offer_requirement_satisfied(
+            offer: Union[Offer, Offer.serializable_dict],
+            bid: Union[Bid, Bid.serializable_dict],
+            offer_requirement: Dict,
+            clearing_rate: Optional[float] = None,
+            selected_energy: Optional[float] = None):
+        """Check whether an offer requirement is satisfied."""
+        offer = Offer.from_dict(offer) if isinstance(offer, dict) else offer
+        bid = Bid.from_dict(bid) if isinstance(bid, dict) else bid
```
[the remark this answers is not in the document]
Hassan754: Okay so we have several use cases,
sometimes we want to test whether 1 specific requirement is satisfied, and in other times we want to test whether at least 1 requirement is satisfied (to mark offer and bid as compatible)
So in the Attributed matching algorithm you can check how I'm using them both

Hassan754:
> Left some comments. In general, I think the flexibility of the architecture is a bit too "expensive" and generic, since it's adding many unused parameters. It would be great to refactor tests a bit, but this can be done after the previous comments are possibly addressed.
> 
> Nits:
> 
> * docstrings for public methods and some pylint warnings.

@fievelk I have adapted to some comments
However, kindly note that this code was already in our codebase, this PR only deals with moving it from d3a to d3a-interface
In general I believe I have left docstrings to all public methods (besides the ones that are overriden, but still I added docstrings to the base method)
I will leave the rest of comments (class renaming/refactoring/testing) as they need more time and discussion
I will add a reference to the comments to this [ticket](https://gridsingularity.atlassian.net/browse/D3ASIM-3637) anyway

Hope this is okay for you

d3a_interface/matching_algorithms/requirements_validators.py
fievelk: > the selected energy between them and the decided on clearing rate

But as far as I see, the classes that implement `Requirement` don't use all of the arguments. That's why I think the interface should be changed.

Hassan754: @fievelk you are right, It's worth giving the design another look

Hassan754: It may not be the `perfect` design yet, however the main rational behind it was to make sure it's scalable to other requirements that could be added in the future

fievelk: > It may not be the perfect design yet, however the main rational behind it was to make sure it's scalable to other requirements that could be added in the future

The problem is exactly that every time we add a new requirement (with its sets of arguments) we need to modify all existing methods to also accept those parameters. That's why I think this is not very scalable. Anyway, this can be dealt with it in the other ticket.

PR #261, opened 2021-09-28
D3ASIM-3596: Export market progression logger to d3a-interface

Hassan754:


requirements/base.txt
```diff
@@ -1,66 +1,35 @@
...
-    # via tox
-tox==3.20.1
-    # via -r requirements/base.ini
-urllib3==1.26.4
-    # via requests
-virtualenv==20.2.2
-    # via tox
-websockets==8.1
-    # via -r requirements/base.ini
+appdirs==1.4.4            # via virtualenv
```
spyrostz, line 7: I think we should at some point settle on the version of pip-tools when running `fab compile` , in order to get rid of huge diffs like this which are mostly caused by a change in the format of the pip-compile tool. How about updating pip-tools to a version greater than 6.2.0 and try compiling the requirements again ? If it does not do the trick then please do not spend more time on it, it is minor.

spyrostz (dismissed):
One comment, LGTM once resolved or discussed.

requirements/base.txt
fievelk: I agree with Spyros; updating `pip` and `pip-tools` to the latest version should be part of the installations process, but also of the requirements update process. Unfortunately, adding `pip` or `pip-tools` to the requirements files doesn't work and it's also discouraged, so this must be done manually. We can add a paragraph in our docs if it's not already there.

requirements/base.ini
```diff
@@ -5,3 +5,4 @@ jsonschema
 websockets
 requests==2.25.1
 kafka-python
+tabulate
```
faizan2590, line 8: Nice to have one blank new line

faizan2590 (approved):
LGTM

requirements/base.txt
Hassan754: @spyrostz @fievelk updated the pip-tools and yes I agree that we should all use the same version

PR #265, opened 2021-10-04
D3ASIM-3593: Add time_slot parameter to the BidOfferMatch and matching algorithms

Hassan754:


hannesdiedrich (approved):
LGTM

tests/test_matching_algorithms/test_pay_as_bid_matching_algorithm.py
```diff
@@ -7,44 +7,53 @@ class TestPayAsBidMatchingAlgorithm:
...
+                        {"id": 12, "seller": "C", "energy_rate": 1, "energy": 65},
+                    ],
+                }
+
             }
         }
         trades = PayAsBidMatchingAlgorithm.get_matches_recommendations(data)
         expected_trades = [
             {"market_id": "market1",
+             "time_slot": "20211006T12:00",
```
spyrostz, line 44: Same as in the d3a PR, for consistency reasons it would be great to change the format of these strings to DATE_TIME_FORMAT.

d3a_interface/matching_algorithms/pay_as_clear_matching_algorithm.py
```diff
@@ -52,19 +52,20 @@ def __init__(self):
...
-            matches.extend(self._create_bid_offer_matches(
-                self.sorted_offers, self.sorted_bids, market_id, current_time))
+        for market_id, time_slot_data in matching_data.items():
+            for time_slot, data in time_slot_data.items():
+                bids = data.get("bids")
+                offers = data.get("offers")
+                current_time = data.get("current_time")
+                clearing = self.get_clearing_point(bids, offers, current_time, market_id)
+                if clearing is None:
+                    return []
```
faizan2590, line 62:
```suggestion
                    return
```

I know it was there before your code changes, but just caught my attention, better to return None as `clearing ` is a dataclass of `Clearing`, so rather than return an empty list, which itself is a different data format then `Clearing`

Hassan754: @Faizan this method should return a list as in the type hints, the `clearing` variable doesn't have anything to do with the return value (We're returning the `matches` list)

PR #282, opened 2021-11-01
Updated Dependencies

faizan2590:


Hassan754 (dismissed):
LGTM

requirements/base.ini
```diff
@@ -1,9 +1,7 @@
 redis
-tox
 pendulum==2.1.2
 jsonschema
 websockets
 requests==2.25.1
 kafka-python
 tabulate
-flake8
```
hannesdiedrich, line 9: Hmm, seems to me like a quick fix. The pre-commit hock will only work if falke is installed, so the user has to install the test environment, right?
IMO We should be consistent and either put flake in the base requirement or in the test environment in both d3a-interface and d3a. What do other people think?

hannesdiedrich: Same applies for `tox` BTW.

hannesdiedrich (commented):
I think you need to recompile the other requirement files too, before pushing.

hannesdiedrich (commented):
Left two important comments. Will review again once discussed (also the connected PRs)

requirements/base.ini
Hassan754: I agree with Hannes

faizan2590: I have moved `tox` to  test and `falke8` was already part of `test`. Give it a look here: https://github.com/gridsingularity/d3a-interface/pull/282/files#diff-f68ab380059c49f1cb9b2ba59eab5058f2aa0b7063701077063bdf9e8157aca4

I moved it to test after a recommendation from Spyros on slack

faizan2590:
> I think you need to recompile the other requirement files too, before pushing.

Sorry, didn't quite get you here. Requirements have `base.in` and `test.in`; And both have been compiled.

requirements/base.ini
Hassan754: I'm not sure about this, but shouldn't these be installed already for the hook to work?

fievelk: The pre-commit hook doesn't need to have flake installed, as far as I know. `pre-commit` downloads and installs flake8 from the github repo specified here: https://github.com/gridsingularity/d3a-interface/blob/edbbad42240edba47d854ad593db79564a0a750f/.pre-commit-config.yaml#L2

requirements/tests.ini
```diff
@@ -1,5 +1,6 @@
 -r base.txt
 
+tox
```
fievelk, line 3: This means that we can run `tox` only if we install the requirements with `pip install -e tests.txt`. Is this expected?

hannesdiedrich (dismissed):
LGTM

spyrostz (dismissed):
LGTM

spyrostz (approved):
LGTM

requirements/base.ini
faizan2590: @fievelk I have tried to address your comment in this commit: https://github.com/gridsingularity/d3a-interface/pull/282/commits/ca9ad04909bfed7b05d6528a46d0e37c6e265b77

I will merge this PR to unblock dependencies update of other repos now. In case its not what you meant, can create a followup pr later.

Thanks!

fievelk: @faizan2590 that's not what I meant :) The original message from Hannes refers to the line in which you removed flake from the requirements. What I wrote meant that flake doesn't need to be added to the requirements files in order to be used by the pre-commit hook, because the instructions provided in `.pre-commit-config.yaml` are telling `pre-commit` to download flake and then use it. The change you linked is disabling this feature, which I believe will cause failures.

fievelk: To be more precise: it will only work if the user has flake8 already installed. But the previous version was more stable because it would have downloaded flake if it wasn't available.

PR #284, opened 2021-11-01
 Fix price requirement validator

Hassan754:


d3a_interface/matching_algorithms/requirements_validators.py
```diff
@@ -98,7 +98,7 @@ def is_satisfied(cls, offer: Offer, bid: Bid, requirement: Dict,
         assert isinstance(clearing_rate, (int, float)), \
             f"Invalid data type for clearing_rate {clearing_rate}"
         # bid_required_price <= price -> true
-        return bid_required_price <= selected_energy * clearing_rate
+        return bid_required_price >= selected_energy * clearing_rate
```
Hassan754, line 101: This is a remnant of when this class was responsible for the rate requirement, but as it is now responsible for the price one, this operator had to be flipped

andreabertolini1995 (approved):
LGTM

spyrostz (approved):
LGTM

PR #294, opened 2021-11-11
Feature/d3 asim 3633

spyrostz:
- Removed deprecated parameter SELL_ON_MOST_EXPENSIVE_MARKET
- Adapted generate_market_slot_list method to include the future market time slots, in order to correctly expand the profiles to take into account future markets too.

### DIFF
--- a/tests/test_user_profile.py
+++ b/tests/test_user_profile.py
@@ -96,7 +96,7 @@ class TestReadUserProfile(unittest.TestCase):
             datetime(2021, 2, 12, 0, 30, 0): 0.1
         })
 
-    def test_correct_time_expansion_read_arbitrary_profile(self):
+    def test_read_arbitrary_profile_returns_profile_with_correct_time_expansion(self):
         market_maker_rate = 30
         if GlobalConfig.IS_CANARY_NETWORK:
             GlobalConfig.sim_duration = duration(hours=3)
@@ -121,4 +121,6 @@ class TestReadUserProfile(unittest.TestCase):
             GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 24
             GlobalConfig.sim_duration = duration(hours=1)
             mmr = read_arbitrary_profile(InputProfileTypes.IDENTITY, market_maker_rate)
-            assert (list(mmr.keys())[-1] - today(tz=TIME_ZONE)).days == 1
+            time_diff = list(mmr.keys())[-1] - today(tz=TIME_ZONE)
+            assert time_diff.hours == 0
+            assert time_diff.days == 1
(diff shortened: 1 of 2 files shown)


Hassan754 (approved):
LGTM

Hassan754 (approved):

tests/test_user_profile.py
```diff
@@ -91,3 +95,30 @@ def test_read_profile_for_player(self):
             datetime(2021, 2, 12, 0, 15, 0): 0.5,
             datetime(2021, 2, 12, 0, 30, 0): 0.1
         })
+
+    def test_correct_time_expansion_read_arbitrary_profile(self):
```
hannesdiedrich, line 99: Following our style convention i would propose:
```suggestion
    def test_read_arbitrary_profile_returns_pofile_with_correct_time_expansion(self):
```

tests/test_user_profile.py
```diff
@@ -91,3 +95,30 @@ def test_read_profile_for_player(self):
...
+            mmr = read_arbitrary_profile(InputProfileTypes.IDENTITY, market_maker_rate)
+            assert list(mmr.keys())[-1] == expected_last_time_slot
+        else:
+            GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 0
+            GlobalConfig.sim_duration = duration(hours=3)
+            mmr = read_arbitrary_profile(InputProfileTypes.IDENTITY, market_maker_rate)
+            assert (list(mmr.keys())[-1] - today(tz=TIME_ZONE)).days == 0
+            GlobalConfig.sim_duration = duration(hours=36)
+            mmr = read_arbitrary_profile(InputProfileTypes.IDENTITY, market_maker_rate)
+            assert (list(mmr.keys())[-1] - today(tz=TIME_ZONE)).days == 1
```
hannesdiedrich, line 119: How about being even more and check for the hours in order to catch (unlikely) errors that only add a few future markets.
```suggestion
            assert (list(mmr.keys())[-1] - today(tz=TIME_ZONE)).hours == 24
```
Also applies to the other checks.

hannesdiedrich (approved):
Two minor comments. LGTM in general

faizan2590 (dismissed):
LGTM

tests/test_user_profile.py
spyrostz: It does not work like that. Pendulum duration(days=1) has the following members:
duration.hours = 0
duration.days = 1
In order to satisfy it I will save it in a variable and compare both

Hassan754 (approved):
LGTM

tests/test_user_profile.py
spyrostz: Forgot to leave a comment here, should be fixed by now !

tests/test_user_profile.py
hannesdiedrich: Darn, I always confuse this. I actually meant:
```suggestion
            assert (list(mmr.keys())[-1] - today(tz=TIME_ZONE)).total_hours == 24
```

PR #300, opened 2021-11-23
D3ASIM-3690: Added limit_float_precision when calculating the energy_…

hannesdiedrich:
…rate in the BaseBidOffer.

### DIFF
--- a/gsy_framework/data_classes.py
+++ b/gsy_framework/data_classes.py
@@ -174,9 +174,6 @@ class Offer(BaseBidOffer):
 
     def __eq__(self, other) -> bool:
         return (self.id == other.id and
-                self.price == other.price and
-                self.original_price == other.original_price and
-                self.energy == other.energy and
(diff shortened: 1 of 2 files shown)


Hassan754 (dismissed):
LGTM

gsy_framework/data_classes.py
```diff
@@ -173,9 +174,6 @@ def from_dict(offer: Dict) -> "Offer":
 
     def __eq__(self, other) -> bool:
         return (self.id == other.id and
-                self.price == other.price and
-                self.original_price == other.original_price and
-                self.energy == other.energy and
```
hannesdiedrich, line 178: I removed these in order to not fall into the trap of comparing `0.99999999` and `10` e.g. for the price. The comparison of the `id` should suffice in general, but I am OK to keep the other checks here too.

fievelk (approved):

faizan2590 (approved):

PR #306, opened 2021-12-02
Adapted the merge_last_market_results_to_global to not recreate the m…

spyrostz:
…arket slot datetime list on every call, but to receive it from teh caller code. Adapted generate_market_slot_list_from_config to ignore the duration check specifically from context where there is no GlobalCofig set.

### DIFF
--- a/gsy_framework/utils.py
+++ b/gsy_framework/utils.py
@@ -68,19 +68,26 @@ def convert_datetime_to_str_in_list(in_list: List, ui_format: bool = False):
 
 
 def generate_market_slot_list_from_config(sim_duration: duration, start_timestamp: DateTime,
-                                          slot_length: duration):
+                                          slot_length: duration, ignore_duration_check=False):
     """
-    Returns a list of all slot times in Datetime format
-    @param sim_duration: Total simulation duration
-    @param start_timestamp: Start datetime of the simulation
-    @param slot_length: Market slot length
-    @return: List with market slot datetimes
+    Returns a list of all slot times in Datetime format.
+
+    Args:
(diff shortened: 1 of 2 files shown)


gsy_framework/utils.py
```diff
@@ -68,19 +68,26 @@ def convert_datetime_to_str_in_list(in_list: List, ui_format: bool = False):
...
-    @return: List with market slot datetimes
+    Returns a list of all slot times in Datetime format.
+
+    Args:
+        sim_duration: Total simulation duration
+        start_timestamp: Start datetime of the simulation
+        slot_length: Market slot length
+        ignore_duration_check: Ignores the check if each timestamp is in the simulation duration.
+                               Useful for calling the method from context where no config has been
+                               set
```
hannesdiedrich, line 81: Sorry, I still don't get why the lack of a config is the reason for this. AFAIS, the slot list is only created from imput variables. But probably I oversaw it. Can you help me please?
(BTW; also a reason fo renaming this method and omit the `_from_config` suffix IMO).

spyrostz: Unfortunately the duration check needs either GlobalConfig, or a constructed SimulationConfig object (in the context of this function, GlobalConfig is used). Therefore, if the GlobalConfig is not set correctly (which is not possible in gsy-web) then such a parameter is needed to circumvent the duration check.

hannesdiedrich: 🤦  darn, overlooked this. Indeed.

hannesdiedrich (approved):
LGTM

Hassan754 (approved):

PR #308, opened 2021-12-15
Feature/gsye 59

spyrostz:
Changed the pay as bid algorithm to be able to reuse the same offer and the same bid in multiple matches, taking into account the residual energy for each of the previous matches.

### DIFF
--- a/gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
+++ b/gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
@@ -16,7 +16,7 @@ class PayAsBidMatchingAlgorithm(BaseMatchingAlgorithm):
     """
 
     @staticmethod
-    def _match_one_bid_one_offer(offer: Dict, bid: Dict, already_selected_order_energy: Dict,
+    def _match_one_bid_one_offer(offer: Dict, bid: Dict, available_order_energy: Dict,
                                  market_id: str, time_slot: str) -> Optional[BidOfferMatch]:
         """
         Try to match one bid with one offer, and at the same time update the dict with the
@@ -26,21 +26,21 @@ class PayAsBidMatchingAlgorithm(BaseMatchingAlgorithm):
         if (offer.get("energy_rate") - bid.get("energy_rate")) > FLOATING_POINT_TOLERANCE:
             return None
 
-        if bid["id"] not in already_selected_order_energy:
-            already_selected_order_energy[bid["id"]] = bid["energy"]
-        if offer["id"] not in already_selected_order_energy:
-            already_selected_order_energy[offer["id"]] = offer["energy"]
-        offer_energy = already_selected_order_energy[offer["id"]]
-        bid_energy = already_selected_order_energy[bid["id"]]
+        if bid["id"] not in available_order_energy:
+            available_order_energy[bid["id"]] = bid["energy"]
+        if offer["id"] not in available_order_energy:
+            available_order_energy[offer["id"]] = offer["energy"]
+        offer_energy = available_order_energy[offer["id"]]
+        bid_energy = available_order_energy[bid["id"]]
 
         selected_energy = min(offer_energy, bid_energy)
         if selected_energy <= FLOATING_POINT_TOLERANCE:
             return None
 
-        already_selected_order_energy[bid["id"]] -= selected_energy
-        already_selected_order_energy[offer["id"]] -= selected_energy
+        available_order_energy[bid["id"]] -= selected_energy
+        available_order_energy[offer["id"]] -= selected_energy
         assert all(v >= -FLOATING_POINT_TOLERANCE
-                   for v in already_selected_order_energy.values())
+                   for v in available_order_energy.values())
 
         return BidOfferMatch(
             market_id=market_id,
@@ -62,20 +62,20 @@ class PayAsBidMatchingAlgorithm(BaseMatchingAlgorithm):
         sorted_bids = sort_list_of_dicts_by_attribute(bids, "energy_rate", True)
         # Sorted offers in descending order
         sorted_offers = sort_list_of_dicts_by_attribute(offers, "energy_rate", True)
-        already_selected_order_energy = {}
+        available_order_energy = {}
         for offer in sorted_offers:
             for bid in sorted_bids:
                 if offer.get("seller") == bid.get("buyer"):
                     continue
 
                 possible_match = cls._match_one_bid_one_offer(
-                    offer, bid, already_selected_order_energy, market_id, time_slot
+                    offer, bid, available_order_energy, market_id, time_slot
                 )
                 if possible_match:
                     bid_offer_matches.append(possible_match)
 
-                if (offer["id"] in already_selected_order_energy and
-                        already_selected_order_energy[offer["id"]] <= FLOATING_POINT_TOLERANCE):
+                if (offer["id"] in available_order_energy and
+                        available_order_energy[offer["id"]] <= FLOATING_POINT_TOLERANCE):
                     break
         return bid_offer_matches
 
--- a/gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
+++ b/gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
@@ -1,4 +1,4 @@
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from gsy_framework.constants_limits import FLOATING_POINT_TOLERANCE
 from gsy_framework.data_classes import BidOfferMatch
@@ -16,10 +16,14 @@ class PayAsBidMatchingAlgorithm(BaseMatchingAlgorithm):
     """
 
     @staticmethod
-    def _match_one_bid_one_offer(offer, bid, already_selected_order_energy,
-                                 market_id, time_slot):
-        if (offer.get("energy_rate") - bid.get(
-                "energy_rate")) > FLOATING_POINT_TOLERANCE:
+    def _match_one_bid_one_offer(offer: Dict, bid: Dict, already_selected_order_energy: Dict,
+                                 market_id: str, time_slot: str) -> Optional[BidOfferMatch]:
+        """
+        Try to match one bid with one offer, and at the same time update the dict with the
+        already selected order energy in order to be able to reuse the same order in future
+        matches.
+        """
+        if (offer.get("energy_rate") - bid.get("energy_rate")) > FLOATING_POINT_TOLERANCE:
             return None
 
         if bid["id"] not in already_selected_order_energy:
@@ -36,7 +40,7 @@ class PayAsBidMatchingAlgorithm(BaseMatchingAlgorithm):
         already_selected_order_energy[bid["id"]] -= selected_energy
         already_selected_order_energy[offer["id"]] -= selected_energy
         assert all(v >= -FLOATING_POINT_TOLERANCE
-                   for v in already_selected_order_energy.values()), f"{already_selected_order_energy}"
+                   for v in already_selected_order_energy.values())
 
         return BidOfferMatch(
             market_id=market_id,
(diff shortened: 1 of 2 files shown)


gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
```diff
@@ -15,33 +15,78 @@ class PayAsBidMatchingAlgorithm(BaseMatchingAlgorithm):
            since the most affordable offers will be allocated for the most aggressive buyers.
     """
 
+    @staticmethod
+    def _match_one_bid_one_offer(offer: Dict, bid: Dict, already_selected_order_energy: Dict,
```
Hassan754, line 19: AFAIS this name is misleading `already_selected_order_energy`, it gives the impression that it's holding track of the amount of energy that was already consumed for each order, whereas at the end of the iteration it will in fact be holding the available energy per each order

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
```diff
@@ -15,33 +15,78 @@ class PayAsBidMatchingAlgorithm(BaseMatchingAlgorithm):
...
+        """
+        if (offer.get("energy_rate") - bid.get("energy_rate")) > FLOATING_POINT_TOLERANCE:
+            return None
+
+        if bid["id"] not in already_selected_order_energy:
+            already_selected_order_energy[bid["id"]] = bid["energy"]
+        if offer["id"] not in already_selected_order_energy:
+            already_selected_order_energy[offer["id"]] = offer["energy"]
+        offer_energy = already_selected_order_energy[offer["id"]]
+        bid_energy = already_selected_order_energy[bid["id"]]
```
hannesdiedrich, line 34: Nit.
```suggestion
        offer_energy = offer["energy"]
        bid_energy = bid["energy"]
        if bid["id"] not in already_selected_order_energy:
            already_selected_order_energy[bid["id"]] = bid_energy
        if offer["id"] not in already_selected_order_energy:
            already_selected_order_energy[offer["id"]] = offer_energy

```

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
Hassan754: I was also wondering whether we may end up with 2 orders with the same id and how we should handle it

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
```diff
@@ -15,33 +15,78 @@ class PayAsBidMatchingAlgorithm(BaseMatchingAlgorithm):
...
+
+        selected_energy = min(offer_energy, bid_energy)
+        if selected_energy <= FLOATING_POINT_TOLERANCE:
+            return None
```
hannesdiedrich, line 38: In this case I would expect that we need to remove the entries from the `already_selected_order_energy` again, before we return, right?

tests/test_matching_algorithms/test_pay_as_bid_matching_algorithm.py
```diff
@@ -46,15 +56,129 @@ def test_perform_pay_as_bid_match(self):
...
+             "bids": [{"id": 1, "buyer": "B", "energy_rate": 3, "energy": 70}],
+             "offers": [{"id": 12, "seller": "C", "energy_rate": 1, "energy": 10}],
+             "selected_energy": 10, "trade_rate": 3},
+            {"market_id": "market1",
+             "time_slot": "2021-10-06T12:00",
+             "bids": [{"id": 1, "buyer": "B", "energy_rate": 3, "energy": 70}],
+             "offers": [{"id": 10, "seller": "A", "energy_rate": 1, "energy": 20}],
+             "selected_energy": 20, "trade_rate": 3},
+        ]
+        print(trades)
```
Hassan754, line 183: to be removed

Hassan754 (dismissed):
General comment, I see that the algorithm is designed to send 1 offer vs 1 bid matches, this is okay indeed but we could've also designed it to send 1 offer with multiple bids in the same match, however both are fine I suppose
left some minor comments, LGTM though

tests/test_matching_algorithms/test_pay_as_bid_matching_algorithm.py
```diff
@@ -46,15 +56,129 @@ def test_perform_pay_as_bid_match(self):
...
+             "bids": [{"id": 2, "buyer": "C", "energy_rate": 3, "energy": 20}],
+             "offers": [{"id": 4, "seller": "A", "energy_rate": 1, "energy": 70}],
+             "selected_energy": 20, "trade_rate": 3},
+            {"market_id": "market1",
+             "time_slot": "2021-10-06T12:00",
+             "bids": [{"id": 1, "buyer": "B", "energy_rate": 3, "energy": 20}],
+             "offers": [{"id": 4, "seller": "A", "energy_rate": 1, "energy": 70}],
+             "selected_energy": 20, "trade_rate": 3},
+        ]
+        assert trades == expected_trades
```
hannesdiedrich, line 150: I was surprised to see trades here, so I propose to rename in the whole module:
```suggestion
        recommendations = PayAsBidMatchingAlgorithm.get_matches_recommendations(data)
        # The bid-id 3 will not match because it has the same buyer as the seller
        expected_recommendations = [
            {"market_id": "market1",
             "time_slot": "2021-10-06T12:00",
             "bids": [{"id": 2, "buyer": "C", "energy_rate": 3, "energy": 20}],
             "offers": [{"id": 4, "seller": "A", "energy_rate": 1, "energy": 70}],
             "selected_energy": 20, "trade_rate": 3},
            {"market_id": "market1",
             "time_slot": "2021-10-06T12:00",
             "bids": [{"id": 1, "buyer": "B", "energy_rate": 3, "energy": 20}],
             "offers": [{"id": 4, "seller": "A", "energy_rate": 1, "energy": 70}],
             "selected_energy": 20, "trade_rate": 3},
        ]
        assert recommendations == expected_recommendations
```
Thanks!

hannesdiedrich (commented):
One question and some nit comments. LGTM in general

spyrostz:
> General comment, I see that the algorithm is designed to send 1 offer vs 1 bid matches, this is okay indeed but we could've also designed it to send 1 offer with multiple bids in the same match, however both are fine I suppose left some minor comments, LGTM though

This is generally not possible, because of the `selected_energy` and `energy_rate` parameter on each match / pair. Even when the offer / bid is reused in multiple pairs, this is usually with different energy rate and selected energy, and if they were all embedded in the same M/N match we would need to save the energy / rate of each bid and offer in the match, which would then be equivalent to having separate matches for each. Have a look for instance at the 1/N M/1 unit tests, the matches should produce different selected_energy values.

Hassan754:
> > General comment, I see that the algorithm is designed to send 1 offer vs 1 bid matches, this is okay indeed but we could've also designed it to send 1 offer with multiple bids in the same match, however both are fine I suppose left some minor comments, LGTM though
> 
> This is generally not possible, because of the `selected_energy` and `energy_rate` parameter on each match / pair. Even when the offer / bid is reused in multiple pairs, this is usually with different energy rate and selected energy, and if they were all embedded in the same M/N match we would need to save the energy / rate of each bid and offer in the match, which would then be equivalent to having separate matches for each. Have a look for instance at the 1/N M/1 unit tests, the matches should produce different selected_energy values.

I'm aware that we'll have some other difficulties, I actually implemented such logic in the preferred trading partner algorithm (which is quite much more complicated due to orders requirements), this is why I was wondering whether we should adapt one of our algorithms to follow the same pattern as the other one

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
spyrostz: Not really, because absence from the already_selected_order_energy implies that the offer has not been seen yet. This dict works the following way: if there are no entries in it, then we assume that the offer / bid has not iterated over yet. First time that we encounter an offer / bid, we add its energy to this dict, and we gradually decrease this energy while we include this order to matches, until it reaches 0. From then on, we leave it to the dict in order for the loop to know that this order has already depleted its energy.

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
spyrostz: These are not equivalent actually. `offer["energy"]` and `bid["energy"]` are the energy from the original offer and bid respectively. However, we might have already included this energy in other matches, therefore we really need to use the already_selected_order_energy values (if they are there of course). If the dict does not contain the orders yet, then we anyways need to populate it, that's why we add the corresponding entries to the dict.

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
Hassan754: Looking at it again I suppose that it can carry two different meanings

1. already_selected  /  order_energy  # the energy selected per each order (this is what my first impression was)
2. already_selected_order  /  energy  # The orders selected with the associated energy available

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
spyrostz: Regarding the first comment, I agree that this is misleading, and I will rename it as you suggested to `available_order_energy`.

Regarding the latter, I think we have discussed it in the past that hash collisions with uuid4() are very very rare, therefore I think that we can safely ignore them. If it would make you feel safer, I think that this will need to be handled by either the Myco or the market validator, which will essentially loop over all the orders and validate that there are no hash collisions. I find it an overkill, if you would like it though I can open a PR that deals with this.

tests/test_matching_algorithms/test_pay_as_bid_matching_algorithm.py
spyrostz: Thanks!

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
hannesdiedrich: 🤦  Indeed, I forgot that the `already_selected_order_energy` is a refference to an instance outside of this function and it might be updated there. Thanks for explaining!

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
Hassan754: It basically can be interpreted as either a mapper of the selected energy (energy that was already consumed) or the absolute opposite (the meaning you already explained above)
Anw I can now see your point, if it can be further improved it would be great, otherwise LGTM

spyrostz:
> > > General comment, I see that the algorithm is designed to send 1 offer vs 1 bid matches, this is okay indeed but we could've also designed it to send 1 offer with multiple bids in the same match, however both are fine I suppose left some minor comments, LGTM though
> > 
> > 
> > This is generally not possible, because of the `selected_energy` and `energy_rate` parameter on each match / pair. Even when the offer / bid is reused in multiple pairs, this is usually with different energy rate and selected energy, and if they were all embedded in the same M/N match we would need to save the energy / rate of each bid and offer in the match, which would then be equivalent to having separate matches for each. Have a look for instance at the 1/N M/1 unit tests, the matches should produce different selected_energy values.
> 
> I'm aware that we'll have some other difficulties, I actually implemented such logic in the preferred trading partner algorithm (which is quite much more complicated due to orders requirements), this is why I was wondering whether we should adapt one of our algorithms to follow the same pattern as the other one

Actually it is not difficulties, it is really impossible algorithmicalle in this case, because of the nature of pay-as-bid it is obligatory that all matches should devolve to 1-1 (because the trades are produced with that assumption, that there is a specific seller/buyer, with specific energy and rate). Therefore, even if one tries to aggregate multiple offers and bids into one BidOfferMatch object, this object will still need to maintain the information about the energy and the rate of each associated pair. TLDR, pay as bid markets are typically peer-to-peer, and not aggregated, thus we need to generate information for each match. 
For the preferred trading partner this might be possible (however algorithmically even degrees of freedom are peer-to-peer, but I need to check the exact implementation)

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
Hassan754: Yes I always knew that this hash collision is a rare situation and yet I'm always anxious about it 😃 , don't bother with it then, Thanks!

gsy_framework/matching_algorithms/pay_as_bid_matching_algorithm.py
spyrostz: No worries at all, I intended for the second meaning during implementation, however it was evident from the review that the name was not clear enough, therefore I will rename it to the one you suggested in the other thread, thanks!

hannesdiedrich (approved):
LGTM

Hassan754 (approved):

PR #310, opened 2021-12-16
PH-932: Fix PVValidator to account for new cloud_coverage setting

hannesdiedrich:


spyrostz (approved):
LGTM

Hassan754 (approved):
LGTM, we should use enums instead of hardcoded values

hannesdiedrich:
> LGTM, we should use enums instead of hardcoded values

This would not have solved the issue unfortunately, I would have had to add the enum instead.

PR #314, opened 2021-12-17
State restore bug

spyrostz:


hannesdiedrich:
@spyrostz I think we can close this PR in favour of my latest https://github.com/gridsingularity/gsy-framework/pull/347 because it it outdated. What do you think?

spyrostz:
> @spyrostz I think we can close this PR in favour of my latest #347 because it it outdated. What do you think?

@hannesdiedrich thanks, I forgot that this was still open! Closing it, thanks again

PR #319, opened 2021-12-27
GSYE-111: Add selected_requirement attribute to the orders

Hassan754:
Orders can have multiple requirements, and the trade should depend on the **selected** requirements (like `energy` or `price`) to take priority over the `energy` or `price` of the actual order.
For instance, let's suppose that we have an offer `offer("seller", energy=10, price=10, requirements=[{"preferred_partners": ["buyerA",] "price": 5]}])`, and a bid `bid("buyerA", energy=10, price=10}])`.
The requirement of the offer is satisfied when creating the match, so the price should be set to 5 instead of 10 because in the validation process of the recommendations we check these values (check [here](https://github.com/gridsingularity/gsy-e/blob/9e4cbeee3f202f50286dbc98039d47e4ce7c631f/src/gsy_e/models/market/two_sided.py#L373)).
In the validation function I posted, we're checking the `price` and `energy` members of the orders (which may be not used), so I saw that we have two possible patches for this:

1. Overwrite the order's members when creating the recommendation (I believe this is a not a good approach for a set of reasons I'm ready to list upon request)
2. pass the `selected_requirement`, it's the requirement that was picked when creating the recommendation so it should be checked before any other member (it's quite important to know which requirement was selected imo)

In this PR, I adopted the second solution, added an extra `selected_requirement`, [populated it when matching based on recommendations](https://github.com/gridsingularity/gsy-myco-sdk/pull/36) and used it in the [validation](https://github.com/gridsingularity/gsy-e/pull/1349)

### DIFF
--- a/gsy_framework/data_classes.py
+++ b/gsy_framework/data_classes.py
@@ -45,7 +45,8 @@ class BaseBidOffer:
     """Base class defining shared functionality of Bid and Offer market structures."""
     def __init__(self, id: str, creation_time: DateTime, price: float, energy: float,
                  original_price: Optional[float] = None, time_slot: DateTime = None,
-                 attributes: Dict = None, requirements: List[Dict] = None):
+                 attributes: Dict = None, requirements: List[Dict] = None,
+                 selected_requirement: Optional[Dict] = None):
         self.id = str(id)
         self.creation_time = creation_time
         self.time_slot = time_slot  # market slot of creation
@@ -55,6 +56,7 @@ class BaseBidOffer:
         self.energy_rate = limit_float_precision(self.price / self.energy)
         self.attributes = attributes
         self.requirements = requirements
+        self.selected_requirement = selected_requirement
 
     def update_price(self, price: float) -> None:
         """Update price member."""
@@ -91,7 +93,8 @@ class BaseBidOffer:
             "creation_time": datetime_to_string_incl_seconds(self.creation_time),
             "time_slot": datetime_to_string_incl_seconds(self.time_slot),
             "attributes": self.attributes,
-            "requirements": self.requirements
+            "requirements": self.requirements,
+            "selected_requirement": self.selected_requirement
         }
 
     @classmethod
@@ -124,10 +127,12 @@ class Offer(BaseBidOffer):
                  seller_id: str = None,
                  attributes: Dict = None,
                  requirements: List[Dict] = None,
-                 time_slot: DateTime = None):
+                 time_slot: DateTime = None,
+                 selected_requirement: Optional[Dict] = None):
         super().__init__(id=id, creation_time=creation_time, price=price, energy=energy,
                          original_price=original_price,
-                         attributes=attributes, requirements=requirements, time_slot=time_slot)
+                         attributes=attributes, requirements=requirements, time_slot=time_slot,
+                         selected_requirement=selected_requirement)
         self.seller = seller
         self.seller_origin = seller_origin
         self.seller_origin_id = seller_origin_id
@@ -170,7 +175,8 @@ class Offer(BaseBidOffer):
             seller_origin_id=offer.get("seller_origin_id"),
             seller_id=offer.get("seller_id"),
             attributes=offer.get("attributes"),
-            requirements=offer.get("requirements"))
+            requirements=offer.get("requirements"),
+            selected_requirement=offer.get("selected_requirement"))
 
     def __eq__(self, other) -> bool:
         return (self.id == other.id and
@@ -197,7 +203,7 @@ class Offer(BaseBidOffer):
         return Offer(offer.id, offer.creation_time, offer.price, offer.energy, offer.seller,
                      offer.original_price, offer.seller_origin, offer.seller_origin_id,
                      offer.seller_id, attributes=offer.attributes, requirements=offer.requirements,
-                     time_slot=offer.time_slot)
+                     time_slot=offer.time_slot, selected_requirement=offer.selected_requirement)
 
 
 class Bid(BaseBidOffer):
@@ -210,11 +216,12 @@ class Bid(BaseBidOffer):
                  buyer_id: str = None,
                  attributes: Dict = None,
                  requirements: List[Dict] = None,
-                 time_slot: Optional[DateTime] = None
-                 ):
+                 time_slot: Optional[DateTime] = None,
+                 selected_requirement: Optional[Dict] = None):
         super().__init__(id=id, creation_time=creation_time, price=price, energy=energy,
                          original_price=original_price,
-                         attributes=attributes, requirements=requirements, time_slot=time_slot)
+                         attributes=attributes, requirements=requirements, time_slot=time_slot,
+                         selected_requirement=selected_requirement)
         self.buyer = buyer
         self.buyer_origin = buyer_origin
         self.buyer_origin_id = buyer_origin_id
@@ -260,8 +267,8 @@ class Bid(BaseBidOffer):
             buyer_id=bid.get("buyer_id"),
             attributes=bid.get("attributes"),
             requirements=bid.get("requirements"),
-            time_slot=str_to_pendulum_datetime(bid.get("time_slot"))
-        )
+            time_slot=str_to_pendulum_datetime(bid.get("time_slot"),),
+            selected_requirement=bid.get("selected_requirement"))
 
     def csv_values(self) -> Tuple:
         """Return values of class members that are needed for creation of CSV export."""
--- a/tests/test_dataclasses.py
+++ b/tests/test_dataclasses.py
@@ -190,7 +190,8 @@ class TestBaseBidOffer:
             "creation_time": datetime_to_string_incl_seconds(bid_offer.creation_time),
             "attributes": bid_offer.attributes,
             "requirements": bid_offer.requirements,
-            "time_slot": datetime_to_string_incl_seconds(bid_offer.time_slot)
+            "time_slot": datetime_to_string_incl_seconds(bid_offer.time_slot),
+            "selected_requirement": bid_offer.selected_requirement
         }
 
     def test_from_json(self):
@@ -305,7 +306,8 @@ class TestOffer:
             "seller_origin": offer.seller_origin,
             "seller_origin_id": offer.seller_origin_id,
             "seller_id": offer.seller_id,
-            "time_slot": datetime_to_string_incl_seconds(offer.time_slot)
+            "time_slot": datetime_to_string_incl_seconds(offer.time_slot),
+            "selected_requirement": offer.selected_requirement
         }
 
     def test_from_dict(self):
@@ -419,7 +421,8 @@ class TestBid:
             "buyer_origin": bid.buyer_origin,
             "buyer_origin_id": bid.buyer_origin_id,
             "buyer_id": bid.buyer_id,
-            "time_slot": datetime_to_string_incl_seconds(bid.time_slot)
+            "time_slot": datetime_to_string_incl_seconds(bid.time_slot),
+            "selected_requirement": bid.selected_requirement
         }
 
     def test_from_dict(self):
--- a/tox.ini
+++ b/tox.ini
@@ -29,7 +29,7 @@ exclude = .env/* , .tox/*
 basepython = python3.8
 commands =
     python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
-    pip-sync requirements/tests.txt
+    pip-sync requirements/tests.txt requirements/pandapower.txt
     pip install -e .
     flake8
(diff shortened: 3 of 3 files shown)


hannesdiedrich:
In order to elaborate further. I understand, that the selected requirement is of interest somehow. However, the PAB rules should always apply and should be used as hard limits. The requirements the user can add to an order are only soft limits. Or did I miss a change of direction in this?

Hassan754:
> > The requirement of the offer is satisfied when creating the match, so the price should be set to 5 instead of 10
> 
> Do you mean the traded_rate with `the price should be set to 5 instead of 10` ? So do you mean that the trade_rate of the match should be 5 instead of 10? If yes, I do not agree fir the pay as bid market.

Nope, the trade_rate will be equal to energy/price where price is the requirement's member "price", have a look [here](https://github.com/gridsingularity/gsy-e/blob/faddd803c73751ab497e8db10a6cca0ea1c920a2/src/gsy_e/models/market/two_sided.py#L390).
Basically we do not have a rate requirement, we have a price one instead

spyrostz:
> > > The requirement of the offer is satisfied when creating the match, so the price should be set to 5 instead of 10
> > 
> > 
> > Do you mean the traded_rate with `the price should be set to 5 instead of 10` ? So do you mean that the trade_rate of the match should be 5 instead of 10? If yes, I do not agree fir the pay as bid market.
> 
> Nope, the trade_rate will be equal to energy/price where price is the requirement's member "price", have a look [here](https://github.com/gridsingularity/gsy-e/blob/faddd803c73751ab497e8db10a6cca0ea1c920a2/src/gsy_e/models/market/two_sided.py#L390). Basically we do not have a rate requirement, we have a price one instead

I think that the latter is not correct. Price on its own does not really tell much in this context, unless it is associated with energy (for instance, in your example ```offer("seller", energy=10, price=10, requirements=[{"preferred_partners": ["buyerA",] "price": 5]}])``` the requirement `price: 5` needs to be associated with an amount of energy. Unless we are talking about the price for the whole 10 kWh, which is okay, however renders this comment incorrect ```Basically we do not have a rate requirement, we have a price one instead```).

In addition to that, on Hannes' comment, it depends on the design. If the DoF are completely decoupled by design from the matching algorithm, then it can very well be that they follow their own algorithm and not pay-as-bid. However, it is very unlikely that this is the intent of this feature (I assume that pay-as-bid should still continue to apply and not pay-as-offer as implied in this example) , I'll check the ticket description again but on first glance I did not see that this is something that Andrea wanted.

Hassan754:
Sorry for not paying greater attention to the example I listed, we indeed should only have the price requirement in the bids, not the offers and this is how it's implemented at the moment, it's just a bad example.
Please consider the requirement to be in the bid's side, not the offer's.

spyrostz:
> Sorry for not paying greater attention to the example I listed, we indeed should only have the price requirement in the bids, not the offers and this is how it's implemented at the moment, it's just a bad example. Please consider the requirement to be in the bid's side, not the offer's.

No worries at all! If we convert the example to the bid side then we indeed have a pay-as-bid algorithm, therefore the behavior makes sense!

hannesdiedrich:
OK, thanks. I had another thorough look into the three PRs. Makes sense now. One question, before I leave my LGTMs:
The user really wants to add a `price` requirement to the orders, not a `rate` requirement?

spyrostz (commented):
I have a general comment here, the selected_requirement is not a property of the bid, but a property of the matching, because it is only needed in the context of the match. Therefore I think that this whole PR can be avoided and the orders do not really need the selected requirement.

Hassan754:
> I have a general comment here, the selected_requirement is not a property of the bid, but a property of the matching, because it is only needed in the context of the match. Therefore I think that this whole PR can be avoided and the orders do not really need the selected requirement.

@spyrostz I originally thought that such member belongs to the BidOfferMatch, however the fact that we may have counter requirement for the offer too (Basically one requirement for offer is satisfied and another compatible one for bid is satisfied).
Furthermore, a BidOfferMatch can contain a list of bids and list of offers, all of them can have a certain selected requirement, this is why I thought it could be a field in the orders

Hassan754:
> OK, thanks. I had another thorough look into the three PRs. Makes sense now. One question, before I leave my LGTMs: The user really wants to add a `price` requirement to the orders, not a `rate` requirement?

@hannesdiedrich can you check this answer from Spyros https://gridsingularity.slack.com/archives/C024VLM6W4E/p1633086974032100?thread_ts=1633082052.019400&cid=C024VLM6W4E

spyrostz:
> > I have a general comment here, the selected_requirement is not a property of the bid, but a property of the matching, because it is only needed in the context of the match. Therefore I think that this whole PR can be avoided and the orders do not really need the selected requirement.
> 
> @spyrostz I originally thought that such member belongs to the BidOfferMatch, however the fact that we may have counter requirement for the offer too (Basically one requirement for offer is satisfied and another compatible one for bid is satisfied). Furthermore, a BidOfferMatch can contain a list of bids and list of offers, all of them can have a certain selected requirement, this is why I thought it could be a field in the orders

Regarding your first point, even if we have multiple requirements from offer and bid, the outcome (in other words, input to the clearing process) is always the same though: select an x amount of energy with y price for bid a and offer b. Therefore I do not think that it makes sense to differentiate selected requirements for offer and bid, if the price / energy combination fulfills the requirements of both then we do not need the information if it is the offer or the bid or both that matched. This is from the perspective of the clearing process, if there is another place where this might be used please point it to me.

The second point is indeed a major concern though, and I think that it might be worth it to rethink the many-to-many matches, because even in pay-as-bid, we have more or less replaced this approach with multiple 1-1 matches. I would vote to take this discussion offline and counterweight the pros and cons of each approach

Hassan754:
> Regarding your first point, even if we have multiple requirements from offer and bid, the outcome (in other words, input to the clearing process) is always the same though: select an x amount of energy with y price for bid a and offer b. Therefore I do not think that it makes sense to differentiate selected requirements for offer and bid, if the price / energy combination fulfills the requirements of both then we do not need the information if it is the offer or the bid or both that matched. This is from the perspective of the clearing process, if there is another place where this might be used please point it to me.
> 
> The second point is indeed a major concern though, and I think that it might be worth it to rethink the many-to-many matches, because even in pay-as-bid, we have more or less replaced this approach with multiple 1-1 matches. I would vote to take this discussion offline and counterweight the pros and cons of each approach

@spyrostz If we're ONLY interested about knowing the energy/price from the requirement, the first point makes sense, however I was thinking that we still need to know from which requirement these values came from (we may also have additional requirements in the future and it could be to our benefit to keep track of the one that got selected). However I don't have a strong opinion on this, so if you like I can adopt your proposal.
Regarding the multiple matching, I once implemented it as per the requirements, but I'll raise this topic again so that we agree on one of the two approaches

spyrostz:
> > Regarding your first point, even if we have multiple requirements from offer and bid, the outcome (in other words, input to the clearing process) is always the same though: select an x amount of energy with y price for bid a and offer b. Therefore I do not think that it makes sense to differentiate selected requirements for offer and bid, if the price / energy combination fulfills the requirements of both then we do not need the information if it is the offer or the bid or both that matched. This is from the perspective of the clearing process, if there is another place where this might be used please point it to me.
> > The second point is indeed a major concern though, and I think that it might be worth it to rethink the many-to-many matches, because even in pay-as-bid, we have more or less replaced this approach with multiple 1-1 matches. I would vote to take this discussion offline and counterweight the pros and cons of each approach
> 
> @spyrostz If we're ONLY interested about knowing the energy/price from the requirement, the first point makes sense, however I was thinking that we still need to know from which requirement these values came from (we may also have additional requirements in the future and it could be to our benefit to keep track of the one that got selected). However I don't have a strong opinion on this, so if you like I can adopt your proposal. Regarding the multiple matching, I once implemented it as per the requirements, but I'll raise this topic again so that we agree on one of the two approaches

@Hassan754 regarding the first point, I think even if we want to track the requirement where these values came from, we can do it implicitly in the BidOfferMatch dataclass by having an extra enum attribute called "requirement_origin" (bad name, but it was the first that came to mind unfortunately) that will track the origin of the selected requirement. That way even if in the future we want to track if the selected requirements were price or attribute (or any other future requirement) related, it will be possible to do so. 
The reason why I insist so much in this, is because conceptually orders should ideally not contain any information related to the matching and clearing processes, because these are 2 separate operations and coupling these will probably lead to problems later on.

Hassan754:
> @Hassan754 regarding the first point, I think even if we want to track the requirement where these values came from, we can do it implicitly in the BidOfferMatch dataclass by having an extra enum attribute called "requirement_origin" (bad name, but it was the first that came to mind unfortunately) that will track the origin of the selected requirement. That way even if in the future we want to track if the selected requirements were price or attribute (or any other future requirement) related, it will be possible to do so. The reason why I insist so much in this, is because conceptually orders should ideally not contain any information related to the matching and clearing processes, because populating the orderbook is a separate operation from these 2 and coupling them will probably lead to problems later on.

@spyrostz even if we create this requirement origin (AFAIU it could be equal to order or requirement) and it's equal to "requirement", we may have several requirements and each one of them can have different value for energy and price.
I may be missing something here so I'm not sure

spyrostz:
> > @Hassan754 regarding the first point, I think even if we want to track the requirement where these values came from, we can do it implicitly in the BidOfferMatch dataclass by having an extra enum attribute called "requirement_origin" (bad name, but it was the first that came to mind unfortunately) that will track the origin of the selected requirement. That way even if in the future we want to track if the selected requirements were price or attribute (or any other future requirement) related, it will be possible to do so. The reason why I insist so much in this, is because conceptually orders should ideally not contain any information related to the matching and clearing processes, because populating the orderbook is a separate operation from these 2 and coupling them will probably lead to problems later on.
> 
> @spyrostz even if we create this requirement origin (AFAIU it could be equal to order or requirement) and it's equal to "requirement", we may have several requirements and each one of them can have different value for energy and price. I may be missing something here so I'm not sure .
> 

Correct me if I am wrong, but as far as I understand this PR was created in order to track only the requirement that was selected. Therefore the "selected" requirement should be only one, not multiple. If the intent was to track all requirements that were met, then the selected requirement should be a list with all the requirements that were met from the bid and the offer, but the same idea still applies.

Hassan754:
> Correct me if I am wrong, but as far as I understand this PR was created in order to track only the requirement that was selected. Therefore the "selected" requirement should be only one, not multiple. If the intent was to track all requirements that were met, then the selected requirement should be a list with all the requirements that were met from the bid and the offer, but the same idea still applies.

When creating matches, we only care for the first requirement dictionary that was met, so we don't really care about the others.
One extra note, we can have requirements for both bids and offers, so when creating a match we compare the compatibility of each requirement of the bid to the requirements of the offer, and if both were met in the same condition, we create the match.
AFAIS you suggested creating an enum field in the BidOfferMatch that will tell whether we got the exchange values from the orders or from their requirements, this will tell us whether we should get the values from orders or from the requirements, but this won't tell us WHICH requirement dict was selected.

spyrostz:
> > Correct me if I am wrong, but as far as I understand this PR was created in order to track only the requirement that was selected. Therefore the "selected" requirement should be only one, not multiple. If the intent was to track all requirements that were met, then the selected requirement should be a list with all the requirements that were met from the bid and the offer, but the same idea still applies.
> 
> When creating matches, we only care for the first requirement dictionary that was met, so we don't really care about the others. One extra note, we can have requirements for both bids and offers, so when creating a match we compare the compatibility of each requirement of the bid to the requirements of the offer, and if both were met in the same condition, we create the match. AFAIS you suggested creating an enum field in the BidOfferMatch that will tell whether we got the exchange values from the orders or from their requirements, this will tell us whether we should get the values from orders or from the requirements, but this won't tell us WHICH requirement dict was selected.

Not necessarily an enum. It can be a dict with as many elements as necessary in order to distinguish which requirement dict was selected, whether it was from the bid or the offer and any other information that might be needed in the future

Hassan754:
> Not necessarily an enum. It can be a dict with as many elements as necessary in order to distinguish which requirement dict was selected, whether it was from the bid or the offer and any other information that might be needed in the future

This seems promising, we may create a dict with `bid_requirement` and `offer_requirement` keys for now, and it could be scalable to include more keys in the future.
If this is what you meant, seems good to me

spyrostz:
> > Not necessarily an enum. It can be a dict with as many elements as necessary in order to distinguish which requirement dict was selected, whether it was from the bid or the offer and any other information that might be needed in the future
> 
> This seems promising, we may create a dict with `bid_requirement` and `offer_requirement` keys for now, and it could be scalable to include more keys in the future. If this is what you meant, seems good to me

Glad that we came to a consensus here! One last point, keep in mind that these fields need to also be None, in order to cover the absence of requirements and attributes.

PR #320, opened 2021-12-28
Avoid KeyError when dealing with d3a_cost in savings KPI

fievelk:
We currently have two possible names for costs in the savings KPI: `d3a_cost` and `gsy_e_cost`. 
The `d3a_cost` key was replaced by `gsy_e_cost` and can therefore be missing. However, this check currently fails with a `KeyError` exception when the old `d3a_cost` key is not found.

This PR fixes the issue.


faizan2590 (approved):

Hassan754 (approved):
LGTM

PR #321, opened 2021-12-28
 Upgrade flake8 from version 3.7.9 to 3.9.2

Hassan754:


requirements/tests.ini
```diff
@@ -1,7 +1,7 @@
 -r base.txt
 
 coverage
-pycodestyle==2.5.0
+pycodestyle==2.8.0
```
fievelk, line 4: I have the feeling that this package is also not used. flake8 should already install it as its sub-dependency automatically if I'm not mistaken.

Hassan754: I've checked and this is true, I removed the explicit installation of pycodestyle

fievelk (approved):

faizan2590 (approved):

PR #322, opened 2021-12-28
Avoid None values when retrieving savings costs

fievelk:
The fix implemented in #320 brought to light another issue with the computation of the `saving_absolute` value. This PR is an attempt to fix it.


faizan2590 (approved):
LGTM

Hassan754 (approved):

PR #323, opened 2022-01-04
Refactor energy_trade_profile

Hassan754:


gsy_framework/sim_results/energy_trade_profile.py
```diff
@@ -78,16 +115,16 @@ def update_sold_bought_energy(self, area_result_dict, core_stats, current_market
...
-        if core_stats[area_result_dict['uuid']] == {}:
+        if core_stats[area_result_dict["uuid"]] == {}:
             return
-        area_core_trades = core_stats[area_result_dict['uuid']].get('trades', [])
+        area_core_trades = core_stats[area_result_dict["uuid"]].get("trades", [])
         for trade in area_core_trades:
-            trade_seller = area_name_from_area_or_ma_name(trade["seller"])
-            trade_buyer = area_name_from_area_or_ma_name(trade["buyer"])
+            trade_seller = trade["seller"]
+            trade_buyer = trade["buyer"]
```
Hassan754, line 127: This is the actual behavior intended in the ticket

gsy_framework/sim_results/energy_trade_profile.py
```diff
@@ -123,8 +123,8 @@ def _calculate_devices_sold_bought_energy(res_dict, area_result_dict, core_stats
             return
         area_core_trades = core_stats[area_result_dict["uuid"]].get("trades", [])
         for trade in area_core_trades:
-            trade_seller = area_name_from_area_or_ma_name(trade["seller"])
-            trade_buyer = area_name_from_area_or_ma_name(trade["buyer"])
```
hannesdiedrich, line 90: Hmm, but from the implementation of `area_name_from_area_or_ma_name` i see that it does exactly what is asked in the requirements: Remove the `MA` from the area name. Can you please help me understand why this is remove here? Would be good to put this into the description of the PR. Thanks

Hassan754: @hannesdiedrich Ah I see, but if this is already implemented, I suppose the ticket is outdated? In this case I think that I misunderstood the requirement, so  I should be looking for the place where the MA prefix is saved alongside the area name in the database, can you help me find it (I suppose we have to avoid saving the prefix in the database, right)?

Hassan754: pinging @faizan2590 because I suppose he was the one working on this AFAIS

hannesdiedrich: @Hassan754 AFAIU, in order to achieve what is written in the requirements, the easiest way would be to remove the `MA` from the `make_ma_name` method (and all other `make_*_name` methods here: https://github.com/gridsingularity/gsy-e/blob/9dca2dbc18a1c19649de66456fcdae9b86234380/src/gsy_e/gsy_e_core/util.py#L171
You can get 100 points if you remove all occurrences of these methods from the code, please check gsy-web, there is also a separate function that deals with that AFAIR.

I hope this does not conflict with old simulation or running CNs. AFAIS from here it does not.

Hassan754: Thanks a lot, I can now see foresee the DoD, I will work on it in separate PRs to unblock the refactoring changes here

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #326, opened 2022-01-12
GSYE-108: Remove Storage loss per hour mechanism

fievelk:
The "loss per hour" mechanism is being deprecated, so this PR removes all validations involving it.

## Reference

- gridsingularity/gsy-e#1357


hannesdiedrich (approved):
LGTM

PR #327, opened 2022-01-24
GSYE-101: Add start_date to class constructor of ResultsHandler

faizan2590:


gsy_framework/sim_results/all_results.py
```diff
@@ -33,6 +36,9 @@ def __init__(self, should_export_plots: bool = False):
             'assets_info': SimulationAssetsInfo()
         }
         self._total_memory_utilization_kb = 0.0
+        self.start_date_time = (
```
Hassan754, line 39: Nit
```suggestion
        self.start_datetime = (
```

PR #334, opened 2022-03-04
Bug/gsye 214 Fix bug in CumulativeGridTrades

hannesdiedrich:
Actual bug fix is in 300e34f162d2ed487c5ccea462975e5ccc5d303a
Also adapted to pylint warnings and had to disable two of them, which would have lead to global refactoring of the results accumulation which should be targeted in a separate story. Shall I create a ticket for this?


spyrostz (approved):
LGTM

fievelk (approved):

PR #335, opened 2022-03-10
Added ability to change the default Kafka URL in order to treat the d…

spyrostz:
…ocker compose cluster the same way as localhost and use the same settings.


spyrostz:
Closing the PR to deal with the issue in a more straightforward way

PR #336, opened 2022-03-11
Removed Kafka default url variable.

spyrostz:
Until now the decision whether to use extra Kafka parameters (e.g. authentication) depended on the URL of the Kafka server, if it was equal to "localhost:9092" then extra Kafka parameters were not set. This did not work well with docker-compose, because we want to circumvent authentication, but at the same time Kafka is not running on localhost but on a separate container, thus having a different hostname. 
This change decouples this choice from the Kafka URL, and adds a new variable called IS_KAFKA_RUNNING_LOCALLY that should be set to false in deployed environments, and the decision on whether to use authentication will depend only on this.

### DIFF
--- a/gsy_framework/kafka_communication/__init__.py
+++ b/gsy_framework/kafka_communication/__init__.py
@@ -1,21 +1,22 @@
 import ssl
 from os import environ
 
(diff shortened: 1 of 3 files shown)


cruky (approved):
LGTM

hannesdiedrich (approved):
LGTM

PR #343, opened 2022-04-28
GSYE-269: Set an empty list as a default children in accumulated_trades

cruky:
This is a precautionary measure in a corner case of missing children key of area in accumulated_trades but the same time there are existed children in area configuration tree.

### DIFF
--- a/gsy_framework/sim_results/cumulative_grid_trades.py
+++ b/gsy_framework/sim_results/cumulative_grid_trades.py
@@ -192,11 +192,13 @@ class CumulativeGridTrades(ResultsBaseClass):
         Returns: None
         """
         current_child_uuids = [
-            c["uuid"] for c in accumulated_trades[area["uuid"]].get("children", [])
+            child["uuid"] for child in accumulated_trades[area["uuid"]].get("children", [])  or []
         ]
         for child in area.get("children", []):
             if child["uuid"] not in current_child_uuids:
-                accumulated_trades[area["uuid"]].setdefault("children", []).append(
+                if accumulated_trades[area["uuid"]].get("children") is None:
+                    accumulated_trades[area["uuid"]]["children"] = []
+                accumulated_trades[area["uuid"]]["children"].append(
                     self._generate_accumulated_trades_child_dict(
                         accumulated_trades, child
                     )
--- a/gsy_framework/sim_results/cumulative_grid_trades.py
+++ b/gsy_framework/sim_results/cumulative_grid_trades.py
@@ -192,7 +192,7 @@ class CumulativeGridTrades(ResultsBaseClass):
         Returns: None
         """
         current_child_uuids = [
-            child["uuid"] for child in accumulated_trades[area["uuid"]].get("children", [])  or []
+            child["uuid"] for child in accumulated_trades[area["uuid"]].get("children", []) or []
         ]
         for child in area.get("children", []):
             if child["uuid"] not in current_child_uuids:
--- a/gsy_framework/sim_results/cumulative_grid_trades.py
+++ b/gsy_framework/sim_results/cumulative_grid_trades.py
@@ -192,7 +192,7 @@ class CumulativeGridTrades(ResultsBaseClass):
         Returns: None
         """
         current_child_uuids = [
-            child["uuid"] for child in accumulated_trades[area["uuid"]].get("children", []) or []
+            child["uuid"] for child in accumulated_trades[area["uuid"]].get("children") or []
         ]
         for child in area.get("children", []):
             if child["uuid"] not in current_child_uuids:
--- a/gsy_framework/sim_results/cumulative_grid_trades.py
+++ b/gsy_framework/sim_results/cumulative_grid_trades.py
@@ -192,11 +192,11 @@ class CumulativeGridTrades(ResultsBaseClass):
         Returns: None
         """
         current_child_uuids = [
-            c["uuid"] for c in accumulated_trades[area["uuid"]]["children"]
+            c["uuid"] for c in accumulated_trades[area["uuid"]].get("children", [])
         ]
         for child in area.get("children", []):
             if child["uuid"] not in current_child_uuids:
-                accumulated_trades[area["uuid"]]["children"].append(
+                accumulated_trades[area["uuid"]].setdefault("children", []).append(
                     self._generate_accumulated_trades_child_dict(
                         accumulated_trades, child
                     )


fievelk (dismissed):

gsy_framework/sim_results/cumulative_grid_trades.py
```diff
@@ -192,11 +192,11 @@ def _update_area_children_in_accumulated_trades_dict(self, accumulated_trades, a
         Returns: None
         """
         current_child_uuids = [
-            c["uuid"] for c in accumulated_trades[area["uuid"]]["children"]
+            c["uuid"] for c in accumulated_trades[area["uuid"]].get("children", [])
```
fievelk, line 195: Nit:
```suggestion
            child["uuid"] for child in accumulated_trades[area["uuid"]].get("children", [])
```
I know this wasn't part of the PR, but it might be good to fix these situations while we refactor existing code. Anyway, it's just minor.

hannesdiedrich, line 195: Can you please explain, what you are trying to achieve here? 
If `children=None`, then the for loop will still fail.

fievelk: I think that if `children = None`, there would be an exception. I suppose that's why Kamil is trying to replace the `None` with a `[]`.

gsy_framework/sim_results/cumulative_grid_trades.py
```diff
@@ -192,11 +192,11 @@ def _update_area_children_in_accumulated_trades_dict(self, accumulated_trades, a
...
         ]
         for child in area.get("children", []):
             if child["uuid"] not in current_child_uuids:
-                accumulated_trades[area["uuid"]]["children"].append(
+                accumulated_trades[area["uuid"]].setdefault("children", []).append(
```
hannesdiedrich, line 199: Sorry, but I also need some info here. Why is this needed? For the case when `children` is not part of the `accumulated_trades[area["uuid"]]` ?

gsy_framework/sim_results/cumulative_grid_trades.py
hannesdiedrich: But this is not the way to replace `None` with `[]`:
```python
di = {"name": "test", "children": None}
print(di.get("children", []))
# None
```

fievelk: Sorry, my answer was incomplete/wrong: `.get()` will help in case `children` is not present at all. If `children` is `None`, then you need something like `.get("children") or []`.

cruky: I expected that children will be missing or it will be a list.  e.g. this [line](https://github.com/gridsingularity/gsy-framework/blob/8abdd3fc44527cd5aebccb621feeeba6b38ffe6a/gsy_framework/sim_results/cumulative_grid_trades.py#L229) is creating the entries in the accumulated_trades so there will be no None value for children. This [line](https://github.com/gridsingularity/gsy-framework/blob/8abdd3fc44527cd5aebccb621feeeba6b38ffe6a/gsy_framework/sim_results/cumulative_grid_trades.py#L180) is creating a child dict and there is an option for missing children key. Although, you are right, if somehow there will be children=None, then this is not a fix. I can add it.

cruky: The main error was:
```
File "/usr/local/lib/python3.8/site-packages/gsy_framework/sim_results/cumulative_grid_trades.py", line 195, in _update_area_children_in_accumulated_trades_dict
c["uuid"] for c in accumulated_trades[area["uuid"]]["children"]
KeyError: 'children'
```
which probably points to the missing `children` key, and not children: None

gsy_framework/sim_results/cumulative_grid_trades.py
cruky: We had a `KeyError: 'children'` for `accumulated_trades[area["uuid"]]["children"]`.  Which means that the `accumulated_trades` has an area which is missing `children` key. So the append method will raise an error in such case. That's why I create the `children` list and append to it, and if it's already created then it will use existed one.

gsy_framework/sim_results/cumulative_grid_trades.py
hannesdiedrich: But is this area supposed to have children? I mean assets do not have children per defenition.

gsy_framework/sim_results/cumulative_grid_trades.py
hannesdiedrich: Here, https://github.com/gridsingularity/gsy-framework/pull/343#discussion_r861098534 also applies.

gsy_framework/sim_results/cumulative_grid_trades.py
cruky: I didn't get to the root cause so I don't know for sure. What I suspect is somehow the area in the accumulated_trades was overwritten and lost their children. Because the other way, the leaf node (area without children) will never reach the point of `_accumulate_area_trades`.

cruky: Is this area supposed to have children? [_accumulate_area_trades](https://github.com/gridsingularity/gsy-framework/blob/8abdd3fc44527cd5aebccb621feeeba6b38ffe6a/gsy_framework/sim_results/cumulative_grid_trades.py#L87) is reached only by Area type, so I think that it is usually should have children.

gsy_framework/sim_results/cumulative_grid_trades.py
```diff
@@ -192,10 +192,12 @@ def _update_area_children_in_accumulated_trades_dict(self, accumulated_trades, a
         Returns: None
         """
         current_child_uuids = [
-            c["uuid"] for c in accumulated_trades[area["uuid"]]["children"]
+            child["uuid"] for child in accumulated_trades[area["uuid"]].get("children", []) or []
```
Hassan754, line 195: nit, this should be enough to cover the non existing key and the case when it exists and equals to None
```suggestion
            child["uuid"] for child in accumulated_trades[area["uuid"]].get("children") or []
```

Hassan754 (approved):

gsy_framework/sim_results/cumulative_grid_trades.py
hannesdiedrich: I see. Indeed, then this is only a quick fix and probably not the fix of the root cause. I can also give it a try when you can not find it. Please let me know.

hannesdiedrich (dismissed):
LGTM

Hassan754 (approved):

spyrostz (approved):

PR #348, opened 2022-05-30
GSYE-269: Revert unneeded children key assignment in accumulated_trades

cruky:
The previous fix was only making the corrupted CN work, but is not needed for non-corrupted simulations. The code from this fix was also populating incorrect value for `cumulative_grid_trades` and caused problems later with `readSimulationResultsQuery`.


hannesdiedrich (approved):
LGTM

PR #352, opened 2022-06-03
GSYE-159: Replace `energy_rate` with a property in BaseBidOffer

mrsaemir:

According to [GSYE-159](https://gridsingularity.atlassian.net/browse/GSYE-159), in the past it was revealed that dynamically calculating `energy_rate` consumes a lot of CPU cycles, thus the decision was to pre-calculate and cache it. We guess the situation is different now and the frequency of using the `energy_rate` member has decreased, so we decided to replace it with a property to see if it has a significant increase in the simulation runtime. In case if an increase, it will be reverted.

## Proposed changes
- Replace `enerty_rate` member by a property
- Refactor `TestBaseBidOffer.test_to_json_string`
- Make Pylint happier!

INTEGRATION_TESTS_BRANCH=master
GSY_FRAMEWORK_BRANCH=master

### DIFF
--- a/gsy_framework/data_classes.py
+++ b/gsy_framework/data_classes.py
@@ -53,19 +53,21 @@ class BaseBidOffer:
         self.original_price = original_price or price
         self.price = price
         self.energy = energy
-        self.energy_rate = limit_float_precision(self.price / self.energy)
         self.attributes = attributes
         self.requirements = requirements
+
+    @property
+    def energy_rate(self) -> float:
+        """Dynamically calculate rate of energy."""
+        return limit_float_precision(self.price / self.energy)
 
     def update_price(self, price: float) -> None:
         """Update price member."""
         self.price = price
-        self.energy_rate = limit_float_precision(self.price / self.energy)
 
     def update_energy(self, energy: float) -> None:
         """Update energy member."""
         self.energy = energy
-        self.energy_rate = limit_float_precision(self.price / self.energy)
 
     def to_json_string(self, **kwargs) -> str:
         """Convert the Offer or Bid object into its JSON representation.
@@ -544,8 +546,8 @@ class BidOfferMatch:
         """
         if "bid_requirement" in (self.matching_requirements or {}):
             return (
-                    self.matching_requirements["bid_requirement"].get("energy")
-                    or self.bid["energy"])
+                self.matching_requirements["bid_requirement"].get("energy")
+                or self.bid["energy"])
         return self.bid["energy"]
 
     @property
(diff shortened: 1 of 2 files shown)


hannesdiedrich (dismissed):
LGTM

spyrostz (approved):
LGTM

Hassan754 (commented):
LGTM once tests are fixed

gsy_framework/data_classes.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, id: str, creation_time: DateTime, price: float, energy: float
         self.energy = energy
         self.attributes = attributes
         self.requirements = requirements
+        self.type = self.__class__.__name__
```
hannesdiedrich, line 58: I remember a discussion about not using `type` as member of objects in order to not get confused with the build-in type. Can somebody else remember?

mrsaemir: I guess it must be OK to use the name `type` as an attribute since we are not overriding the `type` function in any scope of the program. Reading [this](https://stackoverflow.com/questions/30575543/is-using-type-as-an-attribute-name-a-bad-practice#:~:text=General%20rule%20is%3A%20don't,not%20really%20%22reserved%22) might also be useful.

mrsaemir: Further in the comment it says:
> In your case, since you're specifying type as a class attribute, it should be considered safe since it can only be accessed through its class (self.type or SomeProfile.type).

The reason I moved type to the __init__ method is that `obj.__dict__` will automatically return a dictionary of all the object's attributes and we almost always need `type` to be part of this dictionary; If we don't keep type in the object then we have to manually put the `type` key in the dictionary later in every place - which leads to code duplication:

- [Duplication1](https://github.com/gridsingularity/gsy-framework/pull/352/files#diff-3e3db6df6bdaeed8dd98e5be4a800b1e02fcbf5701f8486d1c577d72612a43a8L199)
- [Duplication2](https://github.com/gridsingularity/gsy-framework/pull/352/files#diff-f4f1059a8d7434f1f13437b37402c3fcfeec75b549221046672443b2ad219cf5L87)
- [Duplication3](https://github.com/gridsingularity/gsy-framework/pull/352/files#diff-f4f1059a8d7434f1f13437b37402c3fcfeec75b549221046672443b2ad219cf5L80)

I tried to avoid duplication of `self.__class__.__name__` by moving type into the `__init__` method.

hannesdiedrich: I see. Convinced.

PR #354, opened 2022-06-14
GSYE-295: Add settings for day-ahead market

hannesdiedrich:


gsy_framework/constants_limits.py
```diff
@@ -247,6 +248,7 @@ class GlobalConfig:
     DURATION_D = 1
     MARKET_COUNT = 1
     FUTURE_MARKET_DURATION_HOURS = 0
+    DAY_AHEAD_DURATION_DAYS = 1
```
fievelk, line 251: Here I'm always not sure of the difference between `GlobalSettings` and `ConstSettings.FutureMarketSettings`. Could you remind me why we don't put both in the same class?

fievelk (approved):
LGTM with one question

gsy_framework/constants_limits.py
hannesdiedrich: Good question! I always forget. Having a look at the docstring does not really help either. 
I think I just copied the behaviour from the futures markets' FUTURE_MARKET_DURATION_HOURS. Which does not conform to the above comment `Default simulation settings d3a-web side` ( I will exchange d3a with gsy, too) either.
 
I propose to move both FUTURE_MARKET_DURATION_HOURS and FUTURE_MARKET_DURATION_HOURS into ConstSettings. FUTURE_MARKET_DURATION_HOURS. What do you think?

fievelk: That works for me 👍 Thanks!

hannesdiedrich:
I am closing this PR because the design of the day-ahead market was changed. As soon as the design is finialised, we will update the ticket and I will create a new PR.

PR #355, opened 2022-06-16
GSYE-256: Added assertion on the Trade object constructor, in order t…

spyrostz:
…o avoid cases where the traded energy is negative or zero. Handled cornercase where the offer_bid member of the Trade class is None, which would lead to an AttributeError while logging.


mrsaemir (dismissed):
LGTM

hannesdiedrich (approved):
LGTM

PR #357, opened 2022-06-20
GSYE-261: Community Datasheet parser

fievelk:
Parse an excel (xlsx) file and return a `Datasheet` object (or a JSON/dict representation of it).
This is a draft to make sure we're on the same page about the output. I recommend checking out the branch locally, because it will be much easier to navigate through the code.

## Try it out

I've added a few lines to allow trying out the tool from CLI. A Mapbox API key is needed to convert addresses into geo coordinates. You can omit it but the `geo_tag_coordinates` field won't be populated.

```python
MAPBOX_API_KEY="<your-key>" python gsy_framework/community_datasheet/community_datasheet_parser.py
```

## Required file(s)

You can download a sample Community Datasheet file in the Jira ticket (`Community_DataSheet_v3`). Please note that the `v2` file is not valid and should raise validation errors.

## Details

- Location (address) details are transformed into coordinates (longitude/latitude) to be added to the grid representation.
- Loads declared in the datasheet must define their own profile.
- If a PV doesn’t define a profile it must define the following: `azimuth`, `tilt`, `capacity` and member location.

## References

- https://github.com/gridsingularity/gsy-web/pull/1739
- https://github.com/gridsingularity/gsy-backend-integration-tests/pull/59


gsy_framework/community_datasheet/community_datasheet_parser.py
```diff
@@ -0,0 +1,182 @@
...
+
+        return {
+            "name": "Grid Market",
+            "tags": None,
+            "type": "Area",
+            "uuid": str(uuid.uuid4()),
+            "children": members_grid
+        }
+
+    def _create_member_representation(self, member_name: str) -> Dict:
```
fievelk, line 161: This creates the grid representation of the "member" of the community datasheet. A member is basically a home and can have children (only assets, without further nesting).

hannesdiedrich (commented):
Thanks a lot for this draft (IMO it is quite ready)! I have some questions and comments. Happy discuss further.
As a general comment (having in mind past features that somehow relate):
This feature might not be used a lot and even deprecated at some point (not so far in the future). So IMO we can skip validating each step and care about each corner case that might happen. With this I am not trying to demotivate you to write robust code, but to support you to skip some detours if you agree that a readable error is also OK instead of thinking ahead for the user. I am curious about your opinion about that. Thanks!

gsy_framework/community_datasheet/community_datasheet_parser.py
fievelk: I have the feeling that we are supposed to use the name that was specified by the user in the datasheet (on the "Community Members" sheet). I'm saying this because the name on the grid will be shown in the UI, and if the names don't match this might be confusing. Do you have some specific reason to suggest using "Home"?

gsy_framework/community_datasheet/community_datasheet_parser.py
```diff
@@ -0,0 +1,182 @@
...
+        if any((missing := member) not in self._datasheet.members for member in assets_by_member):
+            raise CommunityDatasheetException(
+                f'Member "{missing}" was defined in one of the asset sheets'  # noqa: F821
+                'but not in the "Community Members" sheet.')
+
+        return assets_by_member
+
+    @staticmethod
+    def _merge_profiles_into_assets(profiles: Dict, assets_by_member: Dict) -> None:
+        """Merge (in-place) each energy profile into the representation of its own asset."""
```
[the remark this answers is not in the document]
fievelk: On one hand, AFAIR dictionary profiles inside the assets are supported by gsy-e. Gsy-e support was initially in the requirements and it's been later removed.

But also: having the profile inside the grid representation makes it easier to create the profiles on gsy-web. There, I would do something like this:
```py
for area, _ in Scenario.traverse_representation(representation):
    ProfileGenerator.generate_profile(configuration, area)
```
So the area will already contain all information required by the profile generator class to decide how to create the profiles (if directly using the profile in the area, or if using rebase to populate it).

gsy_framework/community_datasheet/community_datasheet_parser.py
hannesdiedrich: Ok, I see, so the nomenclature changed yet again :-( 
It was decided to call the members of the community "Homes" at some point (also reflected in the new UI). 
But apparently, now it is "Member". Could you please double check though whether this change in wording is wanted?
Not really important though, just had to remind myself every time that "member" is the house and not an asset when reviewing.

gsy_framework/community_datasheet/community_datasheet_parser.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def _parse_pvs(self, pvs_by_member: Dict, members_information: Dict):
+        self._add_pv_coordinates(pvs_by_member, members_information)
+
+        pv_assets = (
+            pv_asset for member_assets in pvs_by_member.values() for pv_asset in member_assets)
+        for pv_asset in pv_assets:
+            pv_asset["cloud_coverage"] = self._infer_pv_cloud_coverage(pv_asset)
+
+    def _infer_pv_cloud_coverage(self, pv_asset: Dict) -> int:
```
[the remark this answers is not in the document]
hannesdiedrich: Yes, this is what I meant. OK, fine by me, if only user uploaded profiles and rebase PVs are allowed (makes it easier for the first implementation).

gsy_framework/community_datasheet/community_datasheet_parser.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def _create_member_representation(self, member_name: str) -> Dict:
+        return {
+            **self._datasheet.members[member_name],
+            **{
+                "name": member_name,
+                "tags": ["Home"],
+                "type": "Area",
+                "uuid": str(uuid.uuid4()),
+                "children": []
```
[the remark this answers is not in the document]
fievelk: The order of the execution is reversed. First `_create_member_representation()` will create the empty `"children": []`, then `_create_grid()` will overwrite this with the actual `assets`. If the member representation has not been created yet, its children can't be overwritten.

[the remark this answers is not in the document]
hannesdiedrich: Ah yes, you are right. Thanks!

gsy_framework/community_datasheet/community_datasheet_parser.py
```diff
@@ -0,0 +1,182 @@
...
+            }}
+
+
+def parse_community_datasheet(filename):
+    """Parse the content of the community datasheet and return a JSON representation."""
+    parser = CommunityDatasheetParser(filename)
+    print(json.dumps(parser.parse(), indent=4))
+    # NOTE: validate datasheet before or after parsing?
```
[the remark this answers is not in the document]
fievelk: In general, when accepting user input, I'd always do some validation. Not only because users might add malicious input (not sure it's our case here) but mostly because - if something fails - the user needs to know what exactly was the issue so that they can fix it. And not all errors can be understood by the users (e.g. type errors or key errors wouldn't make much sense for them if we don't state what caused them).

If the validation only happens on gsy-web, the user won't get any meaningful information (because the errors they will see won't mention the datasheet fields, but only the grid fields (and the user doesn't know anything about them).

Also, the parser won't fail in some cases. The datasheet might be "formally" valid (e.g. it won't fail while we read the values) but it might contain reciprocally contradictory values. Or it might contain sets of values that are required together (like the Rebase ones). Or it wouldn't fail if we don't fill some mandatory field or if we fill them with inconsistent values.

This is also related to what we discussed in the last retrospective: we'd like to create hackathon tickets to implement (at least) 2 schema validations. One for grids received as input by gsy-web, the other for grids received as input by gsy-e. This might make the validation of the datasheet a bit less necessary (but still useful for the user IMO).

gsy_framework/community_datasheet/community_datasheet_parser.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def _create_grid(self, assets_by_member: Dict) -> Dict:
+        members_grid = []
+        for member_name, assets in assets_by_member.items():
+            member_representation = self._create_member_representation(member_name)
+            member_representation["children"] = assets
+            members_grid.append(member_representation)
+
+        return {
+            "name": "Grid Market",
```
hannesdiedrich, line 154: AFAIS from the implementation of a SCM [setup in Spyros' PR](https://github.com/gridsingularity/gsy-e/pull/1454/files#diff-2fbe96ece8e24e666bfea3890b8aa5ddcf9e16f66a91ba121c8de6384ed6af84R33), I think there will be no grid market, only the community. 
```suggestion
            "name": "Community",
```
Consequently, it also makes sense to rename the method.
Or did you want to be future prove when we also wan to enable adding communities (like in the new UI sense) with the datasheet? If the latter is true, we need to add another hierarchy layer including a market maker.

gsy_framework/community_datasheet/community_datasheet_parser.py
```diff
@@ -0,0 +1,182 @@
...
+
+    @classmethod
+    def _add_pv_coordinates(cls, pvs_by_member: Dict, members_information: Dict):
+        try:
+            location_converter = LocationConverter()
+        except LocationConverterException as ex:
+            logger.warning(ex)
+            return
+
+        with requests.Session() as session:
```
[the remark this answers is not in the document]
fievelk: In a way, yes. Using sessions should add a `keep-alive` header to each request. If the server supports it (and the geocoder API does), it will open only one TCP connection at the beginning, and then use that for all following requests (rather than open a new TCP connection for each request).

For `requests` docs:
> The Session object allows you to persist certain parameters across requests. It also persists cookies across all requests made from the Session instance, and will use urllib3’s connection pooling. So if you’re making several requests to the same host, the underlying TCP connection will be reused, which can result in a significant performance increase (see HTTP persistent connection).

You can check these docs:
- [Geocoder - Using a Session](https://geocoder.readthedocs.io/api.html#using-a-session)
- [Requests - Session Objects](https://requests.readthedocs.io/en/latest/user/advanced/#session-objects)

gsy_framework/community_datasheet/community_datasheet_parser.py
fievelk: Thanks, that's exactly what I was looking for. I wasn't sure about how the grid should have been structured, so I just looked into our current grid representations and copied them.

That's also why in the retro I was asking to create schemas for validation: I think we need to document how grids should be built and what's valid/invalid.

gsy_framework/community_datasheet/community_datasheet_parser.py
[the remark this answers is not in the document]
fievelk: That's good for me. The `ProfileGenerator` can `pop()` the profiles out of the representation when it saves them on the database, so the representation shouldn't contain them anyways.

gsy_framework/community_datasheet/row_converters.py
```diff
@@ -0,0 +1,88 @@
...
+
+    @classmethod
+    def convert(cls, row: Dict) -> Dict:
+        """
+        Convert the row parsed from the Excel file to the asset representation used in scenarios.
+        """
+        # One of the following:
+        # ["LoadHoursStrategy", "DefinedLoadStrategy",
+        # "LoadHoursExternalStrategy", "LoadProfileExternalStrategy",
+        # "LoadForecastExternalStrategy", "Load"]
```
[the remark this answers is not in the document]
hannesdiedrich: > Gsy-web already does this mapping (in DBToSimulationHelpers). To avoid duplicating code, I'd like to refactor that part and re-use it while importing the datasheet grid into gsy-web. This is not so straightforward though, because there are many operations performed by that class. That's what I'd like to work on today.

Now that I know that only profiles are allowed for the energy setup of Loads and PVs, I do not think your need to do that any more. Also because it was decided to not support direct use of the datasheet on gsy-e. The gsy-webs 'DBToSimulationHelpers' will take care of it right ? (I am aware of the fact that this class might need some refactorisation, but nothing to do now IMO).

gsy_framework/community_datasheet/community_datasheet_parser.py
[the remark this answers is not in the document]
hannesdiedrich: Thanks a lot for the summary!

gsy_framework/community_datasheet/community_datasheet_parser.py
spyrostz: I think that specifically for SCM it is up to us and not up to the frontend. Since the only way to create an SCM community is via the community datasheet, we have the liberty to use any internal representation we want. And, in order to hopefully facilitate things hopefully for everyone, I took the easiest way out: remove the grid market, throw away the market maker area/strategy, and keep only the community as a root-level area. 

WRT validation of the grid structure, I was under the impression on our discussion that it was mostly about validation of the parameters and not of the grid structure. If it was about the GSy maps, I would not agree with you because they can in theory not be used only with the 4-level hierarchy that we currently have. However for the SCM you are probably right that a proper validation would save us from some bugs along the way. However I do not think it is very urgent, and can even be offloaded to a 'hackathon' ticket, since it is bite-size and can probably be completed in a couple of hours.

gsy_framework/community_datasheet/community_datasheet_parser.py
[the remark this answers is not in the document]
hannesdiedrich: Ah, ok. I forgot that this here is a step before actually creating the scenario on the DB. Looking forward to the gsy-web PR. Thanks!

gsy_framework/community_datasheet/row_converters.py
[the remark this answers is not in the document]
fievelk: I'll double-check soon. I had this plan before encountering the hiccups with `CloudCoverage`, but you might be right saying that this might not be needed anymore!

gsy_framework/community_datasheet/community_datasheet_parser.py
fievelk: > WRT validation of the grid structure, I was under the impression on our discussion that it was mostly about validation of the parameters and not of the grid structure

I think the two can come together: if we use the JSON schema validator, we can check for parameters types but also structure. Schemas can also be inherited/composed, so some parts of them can be shared (and we can have one schema for the SCM structure and another one for the older hierarchy that reuse the same sub-schemas).

In general, I think a possible DoD might be something like:
- always validate the data we pass from service X to service Y (e.g. gsy-web <-> gsy-e, or datasheet -> gsy-web);
- execute validation in just a few key locations (without having too many validation methods scattered around the codebase);
- have a well-defined, exhaustive schema that can be used as documentation (this is what I missed while working on this feature).

gsy_framework/community_datasheet/community_datasheet_parser.py
hannesdiedrich: Ah, I do not mean the name of the members that the user inputs. I mean the name that we use internally (and also in code).

fievelk: Ah sorry, I had misunderstood. To be sure: do you mean that I should rename some `member` occurrences to `home` (when they concern the grid representation)? For example, renaming this method to `_create_home_representation()`?

hannesdiedrich: Yes, because of the way it is implemented, these members will always be market areas containing assets: homes.
Of course, if you like to leave it as generic as it is, leave it (calling a very late "NIT" :-))

spyrostz (approved):
LGTM

gsy_framework/community_datasheet/community_datasheet_validator.py
```diff
@@ -0,0 +1,94 @@
...
+        self._validate_pvs(datasheet)
+        self._validate_unique_names(datasheet)
+        self._validate_grid(datasheet.grid)
+
+    @staticmethod
+    def _is_timedelta(_checker, instance):
+        return isinstance(instance, timedelta)
+
+    @staticmethod
+    def _validate_unique_names(datasheet):
```
hannesdiedrich, line 45: This is only for mapping the profiles to the correct assets, right? From a simulations' view it only makes sense to omit duplicated names inside one market area.

gsy_framework/community_datasheet/row_converters.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Example:
+            >>> StringToTimedeltaParser.parse("24:01")
+            datetime.timedelta(days=1, seconds=60)
+        """
+        parts = cls.REGEX.fullmatch(time_string)
+        if not parts:
+            raise StringToTimedeltaConversionException(
+                f'Value "{time_string}" does not use the supported format. '
+                'Please use [h]:mm (e.g. "1h30m").')
```
hannesdiedrich, line 32: Does this mean that the hour part is optional?

hannesdiedrich (dismissed):
LGTM

gsy_framework/community_datasheet/row_converters.py
fievelk: Yes, if you only specify two digits they're interpreted as minutes. E.g.:
```python
from gsy_framework.community_datasheet.row_converters import StringToTimedeltaParser

StringToTimedeltaParser.parse("02")
datetime.timedelta(seconds=120)
```

gsy_framework/community_datasheet/row_converters.py
```diff
@@ -0,0 +1,139 @@
...
+    CommunityDatasheetException, StringToTimedeltaConversionException)
+from gsy_framework.community_datasheet.sheet_headers import (
+    LoadSheetHeader, PVSheetHeader, StorageSheetHeader)
+from gsy_framework.constants_limits import DATE_TIME_FORMAT, ConstSettings
+
+
+class StringToTimedeltaConverter:
+    """Class to convert strings to timedelta objects."""
+
+    REGEX = re.compile(r"((?P<hours>\d+?):)?((?P<minutes>\d+?))?")
```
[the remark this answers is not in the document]
fievelk: Thanks! I fixed the regex in 60cfd185125d477efa11a8313de936af06de69dc 👍

gsy_framework/community_datasheet/community_datasheet_validator.py
```diff
@@ -0,0 +1,81 @@
+from jsonschema.validators import validate
+
+from gsy_framework.community_datasheet.community_datasheet_reader import CommunityDatasheet
+from gsy_framework.community_datasheet.exceptions import CommunityDatasheetException
+from gsy_framework.constants_limits import FIELDS_REQUIRED_FOR_REBASE
+
+COMMUNITY_DATASHEET_SCHEMA = {
```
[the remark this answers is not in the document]
hannesdiedrich: How about moving this into `schemas.py`?

tests/test_community_datasheet/test_community_datasheet_parser.py
```diff
@@ -0,0 +1,500 @@
...
+        returned_pvs_by_member["Member 2"][0]["geo_tag_location"] = (-3.1518571, 52.7885236)
+
+        asset_coordinates_builder_mock = MagicMock()
+        asset_coordinates_builder_mock.add_coordinates_to_assets.return_value = (
+            returned_pvs_by_member)
+        asset_coordinates_builder_cls_mock.return_value = asset_coordinates_builder_mock
+
+        datasheet = CommunityDatasheetParser(filename=filename).parse()
+
+        # import ipdb; ipdb.set_trace()
```
hannesdiedrich, line 78: Leftover I guess

hannesdiedrich (dismissed):
Two minor comments. LGTM already

gsy_framework/schemas.py
```diff
@@ -32,7 +32,7 @@ class ScenarioSchemas:
                     "coefficient_percentage": {"type": ["number", "null"]},
                     "fixed_monthly_fee": {"type": ["number", "null"]},
                     "marketplace_monthly_fee": {"type": ["number", "null"]},
-                    "market_maker_rate": {"type": "number"},
+                    "market_maker_rate": {"type": ["number", "null"]},
```
hannesdiedrich, line 35: So I guess the market_maker_rate can be None, too right? Do you think it would be worth it to reflect this also in the typing [here](https://github.com/gridsingularity/gsy-e/blob/af681360fdb93b1793f9b8f799e636dc43b486f6/src/gsy_e/models/area/__init__.py#L264) ?

gsy_framework/community_datasheet/community_datasheet_validator.py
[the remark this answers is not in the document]
fievelk: Done in 822c77d

tests/test_community_datasheet/test_community_datasheet_parser.py
```diff
@@ -0,0 +1,500 @@
...
+                        "2021-08-02T00:00": 11.5,
+                        "2021-08-02T00:15": 11.5,
+                        "2021-08-02T00:30": 11.5,
+                        "2021-08-02T00:44": 11.5,
+                        "2021-08-02T00:59": 11.5,
+                        "2021-08-02T01:14": 11.5,
+                        "2021-08-02T01:29": 11.5,
+                        "2021-08-02T01:44": 11.5,
+                        "2021-08-02T01:59": 11.5,
+                        "2021-08-02T02:14": 11.5,
```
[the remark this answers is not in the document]
fievelk: Done! 11c9a58

gsy_framework/schemas.py
fievelk: I actually think that the way I changed the schema is not correct here. AFAIS, the "Utility Price" value (mapped to `market_maker_rate`) should always be provided in the community datasheet, so all datasheet items with `"type": "Area"` should have it. Not sure why I allowed `null` values at some point; probably it was because the sample datasheet that was provided in the ticket wasn't correct and had many missing values that should be required instead. I'll revert this change, thanks!

hannesdiedrich (dismissed):
LGTM

spyrostz (dismissed):
LGTM

hannesdiedrich (approved):
LGTM

PR #359, opened 2022-06-24
GSYE-327: Added unit tests for the attributed matching algorithm.

spyrostz:
Created a unit test for preferred trading partners that tests the full flow of the algorithm, and another one for the green energy.

### DIFF
--- a/tests/test_matching_algorithms/test_attributed_matching_algorithm.py
+++ b/tests/test_matching_algorithms/test_attributed_matching_algorithm.py
@@ -15,66 +15,33 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-# pylint: disable=missing-function-docstring, protected-access
-import uuid
-from typing import Optional
+# pylint: disable=protected-access
 
-from pendulum import DateTime
-
-from gsy_framework.data_classes import Offer, BidOfferMatch, Bid
+from tests.test_matching_algorithms import offer_factory, bid_factory
+from gsy_framework.data_classes import BidOfferMatch
 from gsy_framework.matching_algorithms.attributed_matching_algorithm import (
     AttributedMatchingAlgorithm)
 
 
 class TestAttributedMatchingAlgorithm:
     """Tester class for the PreferredPartnersMatchingAlgorithm."""
-    @staticmethod
-    def offer_factory(additional_data: Optional[dict] = None):
-        """Create and return an offer object from default and input values."""
-        additional_data = additional_data or {}
-        return Offer(
-            **{"id": str(uuid.uuid4()),
-               "creation_time": DateTime.now(),
-               "time_slot": DateTime.now(),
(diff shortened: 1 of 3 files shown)


tests/test_matching_algorithms/test_attributed_matching_algorithm.py
```diff
@@ -0,0 +1,143 @@
...
+            **{"id": str(uuid.uuid4()),
+               "creation_time": DateTime.now(),
+               "time_slot": DateTime.now(),
+               "price": 10,
+               "energy": 30,
+               "seller": "seller",
+               "seller_id": "seller_id",
+               "seller_origin": "seller",
+               "seller_origin_id": "seller_id",
+               **additional_data})
```
Hassan754, line 45: it seems that we're using these factories in another test module, can we reuse them somehow?

Hassan754 (approved):

fievelk (dismissed):

tests/test_matching_algorithms/test_attributed_matching_algorithm.py
fievelk: We could even use pytest factories and move them into `conftest.py` (as we already do in gsy-web).
- [conftest.py: sharing fixtures across multiple files](https://docs.pytest.org/en/6.2.x/fixture.html#conftest-py-sharing-fixtures-across-multiple-files)

spyrostz: We probably could, however this would introduce dependencies across test classes, that's why I think it is better to have these as static methods outside of all test classes. I will leave it as a post-hackathon TODO and refactor it.

tests/test_matching_algorithms/test_attributed_matching_algorithm.py
```diff
@@ -0,0 +1,143 @@
...
+        offer = self.offer_factory().serializable_dict()
+        bid = self.bid_factory(
+            {"requirements": [{"trading_partners": [offer["seller_id"]]}]}
+        ).serializable_dict()
+
+        offer2 = self.offer_factory()
+        # Both the seller and seller_origin ids are used for preferred trading partner matching.
+        offer2.seller_id = "second seller"
+        offer2.seller_origin_id = "second seller"
+        offer2 = offer2.serializable_dict()
```
hannesdiedrich, line 76: NIT: Wouldn't this also work?
```suggestion
        offer2 = self.offer_factory({"seller_id": "second seller", "seller_origin_id": "second seller"}).serializable_dict()
```

hannesdiedrich (dismissed):
One NIT comment. LGTM

Hassan754 (dismissed):

tests/test_matching_algorithms/test_attributed_matching_algorithm.py
spyrostz: Indeed! I fixed the comment and pushed again, thanks!

hannesdiedrich (approved):
LGTM

PR #362, opened 2022-06-30
GSYE-338: Added SCM parameters to the scenario representation JSON sc…

spyrostz:
…hema.


gsy_framework/schemas.py
```diff
@@ -27,7 +27,11 @@ class ScenarioSchemas:
                     "name": {"type": "string"},
                     "number_of_clones": {"type": "number"},
                     "const_fee_rate": {"type": "number"},
-                    "feed_in_tariff": {"type": "number"},
+                    "feed_in_tariff": {"anyOf": [{"type": "number"}, {"type": "null"}]},
+                    "taxes_surcharges": {"anyOf": [{"type": "number"}, {"type": "null"}]},
+                    "coefficient_percentage": {"anyOf": [{"type": "number"}, {"type": "null"}]},
+                    "fixed_monthly_fee": {"anyOf": [{"type": "number"}, {"type": "null"}]},
+                    "marketplace_monthly_fee": {"anyOf": [{"type": "number"}, {"type": "null"}]},
```
fievelk, line 34: I think this can be simplified:
```suggestion
                    "feed_in_tariff": {"type": ["number", "null"]},
                    "taxes_surcharges": {"type": ["number", "null"]},
                    "coefficient_percentage": {"type": ["number", "null"]},
                    "fixed_monthly_fee": {"type": ["number", "null"]},
                    "marketplace_monthly_fee": {"type": ["number", "null"]},
```

> [...] If the value of "type" is an array, then an instance validates successfully if its type matches any of the types indicated by the strings in the array.

- [Json Schema docs (type)](https://json-schema.org/draft/2020-12/json-schema-validation.html#section-6.1.1)

fievelk (dismissed):
LGTM with one suggestion.

fievelk (approved):

gsy_framework/schemas.py
spyrostz: Thanks, fixed it and the integration tests pass so this indeed works as expected!

PR #368, opened 2022-07-19
Add Grid Market and InfiniteBus to the Datasheet grid

fievelk:
The UI doesn't correctly render the current version of the scenario grid provided by the `CommunityDatasheetParser`.
@arslansajid recommended adding "Grid Market", "Community" and "InfiniteBus" to the output.

The original structure of the grid returned by the community datasheet was supposed to be independent of the frontend implementation (see @spyrostz's [comment here](https://github.com/gridsingularity/gsy-framework/pull/357#discussion_r903503447)), but this doesn't seem to be the case, unfortunately.


fievelk:
@spyrostz I'm not sure this is consistent with the SCM requirements (we had originally decided to avoid the Grid Market, and I'm not sure about the role of the InfiniteBus here). This was requested by the FE to fix map visualization issues. If this grid representation is not correct, the FE team would need to change their logic.

mrsaemir (approved):
Thanks

spyrostz:
> @spyrostz I'm not sure this is consistent with the SCM requirements (we had originally decided to avoid the Grid Market, and I'm not sure about the role of the InfiniteBus here). This was requested by the FE to fix map visualization issues. If this grid representation is not correct, the FE team would need to change their logic.

I think that this is only in order for the FE to maintain the same grid structure that they have in GSy Maps. However, from the standpoint of SCM requirements, Infinite Bus and Grid market are both useless and there is really no point to include them. That's why I am really hesitant to support this change.
@rafaeldalpra @arslansajid can you point if this is included somewhere in the UI designs? because I do not recall the "Grid" market nor the "Grid" asset (Infinite Bus) being included in the SCM design, just the community.

arslansajid:
@spyrostz I am not sure if the "Grid" market or the "Grid" asset (Infinite Bus) being included in the SCM design but it was requested to maintain the same structure of scenarioData in the `uploadCommunityDatasheet ` as we have in the `readConfiguration`. Currently we are facing a problem of plotting assets on the map due to this inconsistent response in scenarioData. I hope this explains the need of making this change.

spyrostz:
> @spyrostz I am not sure if the "Grid" market or the "Grid" asset (Infinite Bus) being included in the SCM design but it was requested to maintain the same structure of scenarioData in the `uploadCommunityDatasheet ` as we have in the `readConfiguration`. Currently we are facing a problem of plotting assets on the map due to this inconsistent response in scenarioData. I hope this explains the need of making this change.

@arslansajid , in order to better understand the situation here, you mean that after you upload a datasheet, you receive a scenarioData that does not include "Grid" market / asset, and if you perform the readConfiguration query right after you receive a scenarioData that includes the "Grid" market / asset ? If yes this is a bug in the backend indeed, and IMO the readConfiguration should return a scenarioData that does not include "Grid" market / asset. If you can validate this it would be great. Thanks in advance! 
Meanwhile I will check the SCM designs and see if the Grid market / asset are indeed included in the designs, in order to adhere to it.

arslansajid:
@spyrostz I confirm that after uploading a datasheet, we receive a scenarioData that does not include "Grid" market / asset, and if we perform the readConfiguration query right after we receive a scenarioData that does not include the "Grid" market / asset.

spyrostz:
> @spyrostz I confirm that after uploading a datasheet, we receive a scenarioData that does not include "Grid" market / asset, and if we perform the readConfiguration query right after we receive a scenarioData that does not include the "Grid" market / asset.

@arslansajid this is probably a typo, but you have written here that in both cases you receive a scenarioData that does NOT include "Grid" market / asset. If this is really the case, then the scenarioData is not inconsistent, since in both cases it fetches the same results. Again probably this is a typo, just mentioning it in order to clarify this.

fievelk:
@spyrostz @arslansajid as far as I know, the scenario data is consistent. Running `readConfiguration` and `uploadCommunityDatasheet` on the same configuration will return the same scenario. Please let me know if this not correct (it would be a bug) but I believe it works properly.

arslansajid:
@spyrostz This is indeed not a typo, what I have written is indeed correct. What I meant is that the scenario data is inconsistent to other simulations which are not created by `uploadCommunityDatasheet`. If you need more clarification I am available to setup a call anytime.

spyrostz:
> @spyrostz This is indeed not a typo, what I have written is indeed correct. What I meant is that the scenario data is inconsistent to other simulations which are not created by `uploadCommunityDatasheet`. If you need more clarification I am available to setup a call anytime.

@arslansajid this is not an inconsistency though, but normal behavior. The scenarioData is not limited to having a 4-level hierarchy (Grid-Community-House-Assets) neither the consumer of the API should expect that this will indeed be the case, and this is evident from the GraphQL output as well, which is modular and can return grids with different number of levels. The case you describe is only for GSy Maps, where the 4-layer hierarchy I described is expected. For the case of SCM though this is not the case, and a 3-level hierarchy is expected instead as far as I recall (Community-House-Assets), therefore it is important to represent it correctly in the scenarioData. 

I am stressing this that much in order to also note that this will also be the case for Electric Blue exchanges, where we will have a 2-level hierarchy (Grid and Assets only), where the same GraphQL output (scenarioData) will be used to represent the grid assets / market. Therefore it is really important to keep scenarioData modular and support different number of levels. Also a trip down the memory lane, for the case of d3a.io, the number of levels was user-defined, and the user could add an arbitrary number of levels.

spyrostz (approved):
LGTM

PR #372, opened 2022-08-02
GSYE-370: Added extra constructor argument to the ResultsHandler clas…

spyrostz:
…s in order to denote that the results are for an SCM simulation. Added current energy bills. Handled empty self sufficiency and self consumption values for some areas. Omitted the calculation of the cumulative market fees for the SCM.


fabriciocarboni (approved):
LGTM

PR #373, opened 2022-08-03
Catch BadZipFile exception for datasheet unsupported xls files

fievelk:
openpyxl doesn't support old xls files:

> openpyxl.utils.exceptions.InvalidFileException: openpyxl does not support the old .xls file format, please use xlrd to read this file, or convert it to the more recent .xlsx file format.

Passing a "plain" xls file to openpyxl correctly raises the expected `InvalidFileException`. However, passing an instance of django's `InMemoryUploadedFile` that refers to an xls raises a `BadZipFile` exception instead.

This PR refactors the code so that this `BadZipFile` exception is caught, to show the proper error (not the “Server Error” we currently have).

### DIFF
--- a/gsy_framework/community_datasheet/community_datasheet_reader.py
+++ b/gsy_framework/community_datasheet/community_datasheet_reader.py
@@ -6,6 +6,7 @@ import pathlib
 from dataclasses import asdict, dataclass, field
 from itertools import chain
 from typing import IO, Dict, List, Union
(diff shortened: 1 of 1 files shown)


mrsaemir (approved):
Thanks

cruky (approved):

spyrostz (approved):
LGTM

PR #375, opened 2022-08-04
Removed the 'accumulated' and 'current' parts from the energy bills i…

spyrostz:
…n order to be in the same format as the GSy Maps bills. Added energy_benchmark to the bills.


mrsaemir (approved):
Thanks for handling it.

fievelk (approved):

PR #382, opened 2022-08-15
GSYE-375: Modify gsy-framework to handle forward market results.

mrsaemir:
As requested by [GSY-373](https://gridsingularity.atlassian.net/browse/GSYE-375), gsy-framework should be modified to be able to handle forward market results and provide appropriate aggregated data for `device_statistics` and the `trade_profile` of each asset.

## Proposed changes:
- [X] Add framework for calculating time window and global stats
- [X] Move AvailableMarketTypes from gsy-e to gsy-framework
- [X] Add related enums and constants
- [x] Add unit tests

### MarketResultsAggregator
* Use the provided results file until strategy classes for forward markets are ready.
* Decompress the provided results.zip file.

```python
if __name__ == '__main__':

    import json
    import pprint
    from pendulum import duration, DateTime
    from gsy_framework.sim_results.aggregate_results import MarketResultsAggregator

    with open("results", "r") as file:
        slot_data = file.read().split("<--->")
        data = [json.loads(d) for d in slot_data if d]

    accumulators = {
        "total_traded_energy": lambda x, y: x + 1 if x else 1
    }
    aggregators = {
        "total_traded_energy": lambda x: 1
    }

    mrh = MarketResultsAggregator(
        resolution=duration(hours=1),
        simulation_slot_length=duration(minutes=15),
        aggregators=aggregators,
        accumulators=accumulators
    )

    for slot_data in data:
        if not slot_data["current_market"]:
            continue
        c_timeslot = DateTime.fromisoformat(slot_data["current_market"])
        for uuid, market in slot_data["simulation_raw_data"].items():
            day_forward_forward_market_stats = market.get("forward_market_stats",
                                                          {}).get("4", {})
            mrh.update(c_timeslot, day_forward_forward_market_stats)

            for i in mrh.generate():
                pprint.pprint(i)
                
>>>
{'accumulated_results': {'total_traded_energy': 1},
 'aggregated_results': {'total_traded_energy': 1},
 'start_time': DateTime(2022, 8, 22, 0, 0, 0)}
>>>
{'accumulated_results': {'total_traded_energy': 2},
 'aggregated_results': {'total_traded_energy': 1},
 'start_time': DateTime(2022, 8, 22, 1, 0, 0)}
>>>
{'accumulated_results': {'total_traded_energy': 3},
 'aggregated_results': {'total_traded_energy': 1},
 'start_time': DateTime(2022, 8, 22, 2, 0, 0)}
>>>
...
```
[results.zip](https://github.com/gridsingularity/gsy-framework/files/9430549/results.zip)


gsy_framework/sim_results/all_results.py
```diff
@@ -28,7 +34,9 @@ def __init__(self, should_export_plots: bool = False, is_scm: bool = False):
             "cumulative_bills": CumulativeBills(),
             "cumulative_grid_trades": CumulativeGridTrades(),
             "device_statistics": DeviceStatistics(should_export_plots),
+            "forward_device_statistics": ForwardMarketDeviceStatistics(should_export_plots),
```
cruky, line 37: Is it possible to make these new keys optional, to not populate these statistics for simulations without forward markets? So for example we could add these keys later, only when there is `ConstSettings.ForwardMarketSettings.ENABLE_FORWARD_MARKETS`.

gsy_framework/sim_results/all_results.py
```diff
@@ -28,7 +34,9 @@ def __init__(self, should_export_plots: bool = False, is_scm: bool = False):
...
             "trade_profile": EnergyTradeProfile(should_export_plots),
+            "forward_trade_profile": ForwardMarketEnergyTradeProfile(should_export_plots),
```
cruky, line 39: If we would like to keep these new keys, then probably we should also include this change in [results_schema](https://github.com/gridsingularity/gsy-framework/blob/072e780a8ac46dc8c9ac516179a337136d577099/gsy_framework/schemas.py#L202) used for validation.

gsy_framework/sim_results/all_results.py
mrsaemir: Well, yes. And to be honest, based on our last discussion with Spyros, I'm about to do so.

gsy_framework/sim_results/aggregate_results.py
```diff
@@ -144,16 +151,15 @@ def generate(self):
 
         sorted_market_timeslots = sorted(self.bids_offers_trades)
         start_time = sorted_market_timeslots[0]
-        next_time = start_time + self.resolution
+        next_time = start_time + self.resolution - self.simulation_slot_length
```
spyrostz, line 156: I am actually having a hard time to find a name instead of simulation_slot_length in order to not correlate the term slot length (which is market-related) to these calculations (which are timeseries - related). Maybe datapoint_interval or datapoint_frequency could work but I also do not like them since they are generic.

gsy_framework/sim_results/aggregate_results.py
```diff
@@ -154,7 +154,13 @@ def generate(self):
         next_time = start_time + self.resolution - self.simulation_slot_length
         timeslots_to_aggregate = []
         for timeslot in sorted_market_timeslots:
-            if timeslot >= next_time:
+            if timeslot > next_time:
```
spyrostz, line 159: In order to avoid code duplication, I would maintain the >= and add an extra clause inside the block:
```
if timeslot > next_time:
    timeslots_to_aggregate = [timeslot]
else:
    timeslots_to_aggregate = []
```
since as far as I see this is the only difference in the 2 blocks.

mrsaemir: The order of things is also different. I preferred the code to be easier to read rather than perfectly written here. But still, I don't like how this is written either.

cruky (approved):
LGTM, Thanks for the docstrings and explanatory tests!

gsy_framework/sim_results/aggregate_results.py
spyrostz: I see, then it is fine by me. I do not see an immediate or simple way to fix this unfortunately

PR #383, opened 2022-08-16
PH-1246: Set CDS initial_soc to be equal to min_allowed_soc

fievelk:
With the Community Datasheet there is no way to specify the `initial_soc` of the battery. The `initial_soc` of battery devices is therefore by default set to `10` ([see code](https://github.com/gridsingularity/gsy-web/blob/fc442f43ae7b7d7b781e1e7972404126f54b8537/simulation/graphql/configuration_objects.py#L140)). 

However, if the `minimum_soc` defined by the user in the CDS excel file is higher than the `initial_soc`, an exception is raised.
To avoid this, this PR sets the `initial_soc` to be equal to the `min_allowed_soc` specified by the user.


cruky (approved):
LGTM

spyrostz (approved):
LGTM

PR #386, opened 2022-08-24
GSYE-337: Refactor: Remove encode_and_compress_dict from gsy-web and decompress…

fabriciocarboni:
…_and_decode_queued_strings from gsy-e and create it in gsy-framework

As described in [GSYE-337](https://gridsingularity.atlassian.net/browse/GSYE-337) both `encode_and_compress_dict` and `decode_queued_strings` should be removed from gsy-web and gsy-e and placed in gsy-framework.

### DIFF
--- a/gsy_framework/serialization.py
+++ b/gsy_framework/serialization.py
@@ -1,6 +1,6 @@
 import pickle
 from typing import Dict
-from zlib import decompress, compress
+from zlib import compress, decompress
 
 
 class CompressSerialization:
--- a/gsy_framework/serialization.py
+++ b/gsy_framework/serialization.py
@@ -1,0 +1,17 @@
+import pickle
+from typing import Dict
+from zlib import decompress, compress
+
+
+class CompressSerialization:
+    """Custom Serializer / Deserializer."""
+    @staticmethod
+    def encode_and_compress_dict(uncompressed_dict):
+        """Convert a dict into str by pickling and compressing."""
+        pickled_str = pickle.dumps(uncompressed_dict)
+        return compress(pickled_str)
+
+    @staticmethod
+    def decompress_and_decode_queued_strings(compressed_dict: bytes) -> Dict:
+        """Decompress and decode data sent via redis queue."""
+        return pickle.loads(decompress(compressed_dict))
--- a/tests/test_dataserializer.py
+++ b/tests/test_dataserializer.py
@@ -36,4 +36,3 @@ class TestCompressDecompressDict:
         data = DataSerializer.encode_and_compress_dict({'key1': 'value1', 'key2': 'value2'})
         uncompressed_data = DataSerializer.decompress_and_decode(data)
         assert type(uncompressed_data) == dict
-
--- a/tests/test_dataserializer.py
+++ b/tests/test_dataserializer.py
@@ -1,0 +1,39 @@
+"""
+Copyright 2018 Grid Singularity
+This file is part of Grid Singularity Exchange.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+from typing import Dict
+from gsy_framework.data_serializer import DataSerializer
+
+
+class TestCompressDecompressDict:
+    """Test compress and decompress dictionaries"""
+
+    @staticmethod
+    def test_compress() -> bytes:
+        """Assert if data is bytes type"""
+        data = {'key1': 'value1', 'key2': 'value2'}
+        compressed_data = DataSerializer.encode_and_compress_dict(data)
+        assert type(compressed_data) == bytes
+
+    @staticmethod
+    def test_decompress() -> Dict:
+        """Assert if data is dict type"""
+        data = DataSerializer.encode_and_compress_dict({'key1': 'value1', 'key2': 'value2'})
+        uncompressed_data = DataSerializer.decompress_and_decode(data)
+        assert type(uncompressed_data) == dict
+
--- a/gsy_framework/data_serializer.py
+++ b/gsy_framework/data_serializer.py
@@ -1,17 +1,17 @@
 import pickle
-from typing import Dict
 from zlib import compress, decompress
+from typing import Dict
 
 
-class CompressSerialization:
+class DataSerializer:
     """Custom Serializer / Deserializer."""
     @staticmethod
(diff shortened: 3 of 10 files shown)


gsy_framework/serialization.py
```diff
@@ -0,0 +1,17 @@
+import pickle
+from typing import Dict
+from zlib import decompress, compress
```
mrsaemir, line 3: Just a tiny isort issue here:
`from zlib import compress, decompress`

mrsaemir (commented):
Just one minor issue.
LGTM when resolved.

gsy_framework/serialization.py
fabriciocarboni: Thanks!

gsy_framework/serialization.py
```diff
@@ -0,0 +1,17 @@
+import pickle
+from typing import Dict
+from zlib import compress, decompress
+
+
+class CompressSerialization:
+    """Custom Serializer / Deserializer."""
+    @staticmethod
+    def encode_and_compress_dict(uncompressed_dict):
```
fievelk, line 9: Nit: we try to use type hints to document what kind of input we have. For example, in this case it would be:
```suggestion
from typing import Dict
# ...
    def encode_and_compress_dict(uncompressed_dict: Dict) -> bytes:
```
The same would apply to other methods below

gsy_framework/serialization.py
```diff
@@ -0,0 +1,17 @@
+import pickle
+from typing import Dict
+from zlib import compress, decompress
+
+
+class CompressSerialization:
```
fievelk, line 6: I think we can find a way to further improve this name. On one side, we try to use nouns as class names and not verbs, so in this case it could be something like `SerializationCompressor`. But I would still not be completely convinced.
This class does: 
- compression/de-compression
- encoding/decoding
- Also, in the docstring you mention that this is a serializer/deserializer. 

What about just calling it something like `DataSerializer`? There are probably better alternatives, so feel free to suggest any other option.

fievelk (commented):
I've left some comments. I'd also like to have unit-tests for this class. I know it's mostly using external libraries (that are already tested) but I think it would be a good opportunity to practice unit-testing and our guidelines about them.

gsy_framework/serialization.py
```diff
@@ -0,0 +1,17 @@
...
+    """Custom Serializer / Deserializer."""
+    @staticmethod
+    def encode_and_compress_dict(uncompressed_dict):
+        """Convert a dict into str by pickling and compressing."""
+        pickled_str = pickle.dumps(uncompressed_dict)
+        return compress(pickled_str)
+
+    @staticmethod
+    def decompress_and_decode_queued_strings(compressed_dict: bytes) -> Dict:
```
fievelk, line 15: Here we don't need to know that we'll only decode "queued strings": the class is quite flexible, so it can be re-used for any kind of compressed byte data. Also, we're decompressing/decoding bytes, not strings (which would actually be dicts). Maybe something simpler would still work? For example
```suggestion
    def decompress_and_decode(compressed_dict: bytes) -> Dict:
```
Not sure if this would be optimal. What do you think?

tests/test_dataserializer.py
```diff
@@ -0,0 +1,38 @@
...
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+from typing import Dict
+from gsy_framework.data_serializer import DataSerializer
+
+
+class TestCompressDecompressDict:
```
fievelk, line 23: Following our testing guidelines, I'd name this following the name of the class we're testing:
```suggestion
class TestDataSerializer:
```
This helps mapping tests to their class and also helps when using the search feature of IDEs.

tests/test_dataserializer.py
```diff
@@ -0,0 +1,38 @@
...
+    """Test compress and decompress dictionaries"""
+
+    @staticmethod
+    def test_compress() -> bytes:
```
fievelk, line 27: Similar thing here: since we're testing the `encode_and_compress_dict`, I'd rename this as
```suggestion
    def test_encode_and_compress_dict():
```
I've also removed the `-> bytes` part. The `->` is used to document what type of output is returned by the method. In this case, the test method isn't returning bytes, but rather `None`.

tests/test_dataserializer.py
```diff
@@ -0,0 +1,38 @@
...
+        """Assert if data is bytes type"""
+        data = {'key1': 'value1', 'key2': 'value2'}
+        compressed_data = DataSerializer.encode_and_compress_dict(data)
+        assert type(compressed_data) == bytes
```
fievelk, line 31: Nit: I think this test is enough for the moment, because I don't expect this class to change much in the future. But I'd like to show you one shortcoming of this.
Here, we're testing that the `encode_and_compress_dict()` method will return an object of type `bytes`. However, this doesn't guarantee that the bytes object will actually be the compressed version of the input dictionary. Suppose I modify the `encode_and_compress_dict()` method and just return a random byte object:
```py
    @staticmethod
    def encode_and_compress_dict(uncompressed_dict: Dict) -> bytes:
        return bytes("some string", "utf-8")
```
This test will still succeed, even though the method is now clearly not doing what we expect.

I think it could be tricky to fix this because we're dealing with bytes, but it's still worth knowing.

tests/test_dataserializer.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @staticmethod
+    def test_decompress() -> Dict:
```
fievelk, line 34: Same as above:
```suggestion
    def test_decompress_and_decode():
```

tests/test_dataserializer.py
```diff
@@ -0,0 +1,38 @@
...
+        """Assert if data is dict type"""
+        data = DataSerializer.encode_and_compress_dict({'key1': 'value1', 'key2': 'value2'})
+        uncompressed_data = DataSerializer.decompress_and_decode(data)
+        assert type(uncompressed_data) == dict
```
fievelk, line 38: Nit: Here we're trusting that `DataSerializer.encode_and_compress_dict()` will actually do what we expect. In theory, we have no assurance about this. So we would be feeding a `data` object that we don't know much about to `DataSerializer.decompress_and_decode()`.
What we would usually try to avoid this is to pass hardcoded data. Here, the thing becomes more complicated because we're dealing with bytes objects. But in theory this could be rewritten as:
```suggestion
        data = b'x\x9ck`\x99*\xcc\x00\x01\xb5Sz\xb8\x8a\xf3sSu\xcb\x12sJS\xa7xW\x17\xeb\x01\x00p\x08\x08\xcd'
        uncompressed_data = DataSerializer.decompress_and_decode(data)
        assert uncompressed_data == {'some-value': 123}
```
To be honest, dealing with `bytes` like this is an unexplored territory for me, that's why I consider this to be a nit.

tests/test_dataserializer.py
```diff
@@ -0,0 +1,38 @@
+"""
```
fievelk, line 1: I'd recommend renaming this module to `test_data_serializer.py` (with an underscore between `data` and `serializer`) because that's the name of the tested module.

fievelk (commented):
Left some comments. They're not vital, but hopefully you'll find them useful :D

mrsaemir (approved):

PR #388, opened 2022-09-06
feature/GSYE-386-2: Add EBEnergyProfileAggregator

mrsaemir:
### This PR requires https://github.com/gridsingularity/gsy-framework/pull/387 to be merged first. So, please review that before checking this one.

As requested by [GSYE-386](https://gridsingularity.atlassian.net/browse/GSYE-386), we need aggregator functions to support the graphs in EB exchanges:
Screenshot 2022-09-05 at 10 39 47

## Proposed changes:
- [X] Implement `EBEnergyProfileAggregator` class ([here](https://github.com/gridsingularity/gsy-framework/blob/d624c5896239dbec4e1c5c1d6932aa65c2f585ca/gsy_framework/sim_results/electric_blue/aggregators.py))
- [x] Add `EBEnergyProfileAggregator` to `RESOLUTION_AGGREGATIONS` dictionary.
- [x] Add unit tests.
- [X] Add stricter validations for timeslots in MarketResultsAggregator

### DIFF
--- a/gsy_framework/sim_results/electric_blue/aggregators.py
+++ b/gsy_framework/sim_results/electric_blue/aggregators.py
@@ -1,0 +1,56 @@
+from abc import ABC, abstractmethod
+from collections import defaultdict
+from typing import Dict
+
+
+class BaseAggregator(ABC):
+    """Base accumulator class."""
+    @abstractmethod
+    def __call__(self, collected_raw_data: Dict):
+        """Calculate and return accumulated results."""
+
+
+class EBEnergyTradeProfileAggregator(BaseAggregator):
+    """Aggregator class for EB energy trade profile."""
+
+    DEFAULT_STATS = {
+        # open, still unmatched bids/offers
+        "open_buy_orders": 0.0,
+        "open_sell_orders": 0.0,
+        # matched bids/offers
+        "matched_buy_orders": 0.0,
+        "matched_sell_orders": 0.0,
+        # all
+        "all_buy_orders": 0.0,
+        "all_sell_orders": 0.0,
+    }
+
+    def __call__(self, collected_raw_data: Dict):
+        current_result = defaultdict(lambda: dict(self.DEFAULT_STATS))
+
+        for trade in collected_raw_data["trades"]:
+            seller_id = trade["seller_id"]
+            buyer_id = trade["buyer_id"]
+            energy = trade["energy"]
+            # SELLER
+            current_result[seller_id]["matched_sell_orders"] += energy
+            current_result[seller_id]["all_sell_orders"] += energy
+            # BUYER
+            current_result[buyer_id]["matched_buy_orders"] += energy
+            current_result[buyer_id]["all_buy_orders"] += energy
+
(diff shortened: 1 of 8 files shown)


gsy_framework/sim_results/electric_blue/aggregators.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def __call__(self, collected_raw_data: Dict):
+        current_result = defaultdict(lambda: dict(self.DEFAULT_STATS))
+
+        for trade in collected_raw_data["trades"]:
+            seller_id = trade["seller_id"]
+            buyer_id = trade["buyer_id"]
+            energy = trade["energy"]
+            # SELLER
+            current_result[seller_id]["matched_sell_orders"] += energy
```
spyrostz, line 36: In order to be explicit, I would append _kWh to all these dict keys to denote that this is energy:

```suggestion
            current_result[seller_id]["matched_sell_orders_kWh"] += energy
```

tests/test_sim_results/test_aggregate_results.py
```diff
@@ -68,17 +74,18 @@ def test_bids_offers_trades_buffer_gets_updated(self):
 
         # check no duplicated entries are added for each timeslot
         for order in ("bids", "offers", "trades"):
-            buffered_orders = set()
+            buff_orders = set()
             for market_stats in market_results_aggr.bids_offers_trades.values():
-                buffered_orders = buffered_orders.union(set(market_stats[order]))
-            assert len(buffered_orders) == 4 * MARKET_TIMESLOTS * ORDERS_PER_MARKET_SLOT
+                buff_orders = buff_orders.union(set(o["uuid"] for o in market_stats[order]))
```
spyrostz, line 79: Out of curiosity, are there any orders that do not include a uuid ?

spyrostz (approved):
LGTM, 2 minor comments.

gsy_framework/sim_results/electric_blue/aggregators.py
mrsaemir: Will fix that soon

tests/test_sim_results/test_aggregate_results.py
mrsaemir: I'm not sure. I did this to ensure collected orders are unique and the same order is not repeatedly appended as a fault. Why are asking this?

spyrostz: OK I see the rationale behind this now. I was just wondering because the old version of the code would end up in exactly the same outcome since the buff_orders is only used for counting the length of the order list. Since the order uuids are created as part of the test using uuid4 there is (almost) never a chance of hash collision, since it ends up to the same outcome though no problem at all keeping it like this.

mrsaemir: Good. I will add some comments to explain the rationale behind this in the code. Thanks for pointing it out.

spyrostz (approved):
LGTM

PR #389, opened 2022-09-12
refactor/GSYE-375: Replace MarketResultsAggregator with ForwardDeviceStats

mrsaemir:
## Proposed Changes:
- Completely remove `MarketResultsAggregator` class
- Completely remove `AggregationTimeManager` class
- Add `ForwardDeviceStats` data class for accumulated results
- Add `ForwardDeviceTimeSeries` class to generate time series data for each (device_uuid, timeslot, market_type)
- Add functionality to save open bids and offers

## In Progress:
- [x] Unit tests
- [x] Documents

### How to proceed?
1- Receive forward market stats from Kafka
2- For each market type, call `handle_forward_results` and call its return value `results`
3- For each timeslot in the results, iterate over devices
4- For each device, fetch previous stats of the device from the DB and create the corresponding ForwardDeviceStats object: ```previous_device_stats = ForwardDeviceStats(**fetched_from_db)```
5- Sum up the previous and the current ForwardDeviceStats ```updated_device_stats = previous_device_stats + current_device_stats```
6- Save the updated device stats to DB
7- Generate device time series using `ForwardDeviceTimeSeries`
8- Save device time series.


gsy_framework/sim_results/electric_blue/aggregate_results.py
```diff
@@ -0,0 +1,145 @@
...
+from typing import Dict, List
+
+from pendulum import DateTime
+
+from gsy_framework.enums import AggregationResolution, AvailableMarketTypes
+
+# Used by forward markets; the following dictionary defines
+# what aggregations are needed for each market type.
+MARKET_RESOLUTIONS = {
+    AvailableMarketTypes.YEAR_FORWARD: [AggregationResolution.RES_1_MONTH],
```
hannesdiedrich, line 12: I probably missed a discussion about this, but why do you use enums here and not directly assign a duration here?
```suggestion
    AvailableMarketTypes.YEAR_FORWARD: duration(months=1),
```

gsy_framework/sim_results/electric_blue/aggregate_results.py
```diff
@@ -0,0 +1,145 @@
...
+    total_energy_sold: float = 0
+    total_earned_eur: float = 0
+
+    # BUYER
+    total_energy_consumed: float = 0
+    total_buy_trade_count: int = 0
+    total_energy_bought: float = 0
+    total_spent_eur: float = 0
+
+    open_offers: List[Dict] = field(default_factory=list)
```
hannesdiedrich, line 40: AFAIU, you do not need this, as there is no default list, right?
Wouldn't the following be sufficient?
```suggestion
    open_offers: List[Dict] = []
```

gsy_framework/sim_results/electric_blue/aggregate_results.py
mrsaemir: We couldn't easily persist `Duration` objects in DB, so we created this enum.

gsy_framework/sim_results/electric_blue/aggregate_results.py
mrsaemir: Well, the @dataclss decorator doesn't allow you to use `open_offers: List[Dict] = []`. It needs a factory object.

gsy_framework/sim_results/electric_blue/time_series.py
```diff
@@ -0,0 +1,148 @@
...
+        trade_profile_generator = ForwardTradeProfileGenerator(
+            peak_kWh=all_sell_orders_kWh)
+        all_sell_profile = trade_profile_generator.generate_trade_profile(
+            energy_kWh=all_sell_orders_kWh,
+            market_slot=DateTime.fromisoformat(self.device_stats.timeslot),
+            product_type=self.product_type
+        )
+        return resample_data(all_sell_profile, resolution, aggregator_fn=sum)
+
+    def generate(self, resolution: Duration):
```
hannesdiedrich, line 139: NIT: Probably personal taste, but I remember that Pierpaolo suggested to move the public methods to the top of the class in order to see the interface right away without having to scroll down. I agree with that. Thanks for considering!

gsy_framework/sim_results/electric_blue/aggregate_results.py
hannesdiedrich: Ahh, I seen, thanks! Dataclasses are still new to me.

gsy_framework/sim_results/electric_blue/time_series.py
mrsaemir: Will fix it soon!

gsy_framework/sim_results/electric_blue/aggregate_results.py
hannesdiedrich: Ah, ok. But I am wondering: Will we save the timeseries data returned by `ForwardDeviceTimeSeries` in the DB for different resolutions? I see that you wrote this in point 8 in the PR description. But I fail to find the point where this is actually done. Can you please help me? Thanks!

mrsaemir: Yes. The `generate` method on `ForwardDeviceTimeSeries` accepts the resolution to calculate. The returned data points will then be saved to DB. However, I'm not sure whether we save each each data point as a DB row or we store one giant json containing all data point. The saving operation will be done by @cruky in the next step.

hannesdiedrich: I see, thanks!
This is a discussion about performance I think. If we want to make high resolution trade profiles available very quickly, we probably need to pre-calculate these and store them into the DB as soon as there are new markets slot started.
If the delay due to the calculation is not so large, I would vote to not clutter the DB. 
Tagging @spyrostz that has probably spent some thought on this

gsy_framework/sim_results/electric_blue/aggregate_results.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def add_offer(self, offer: Dict) -> None:
+        """Add offer information to device stats."""
+        assert offer["timeslot"] == self.timeslot
+        assert offer["seller_id"] == self.device_uuid, "Device is not seller of the offer."
+        self.open_offers.append(offer)
+
+
+def handle_forward_results(
+        current_timeslot: DateTime, market_stats: Dict[str, Dict[str, List]]) -> Dict:
```
hannesdiedrich, line 109: AFAIU:
```suggestion
        current_timeslot: DateTime, market_stats: Dict[str, Dict[str, ForwardDeviceStats]]) -> Dict:
```

hannesdiedrich, line 109: I understand that `market_stats` are coming from the gsy-e and contain the latest orders, meaning the orders that are performed in the forward markets, right?
Can you please sketch the content of the `market_stats` dict here? Is it the following?
```
market_stats = {
    "20220101T00:00": {"offers": [], "bids": [], "trades": []},
    ...
}
```
?
What about the trades that happened in the same market slots but are not part of the `market_slots` any more. 
Are their buffered somewhere? Or is there no need?
I am asking because you are performing `self.total_energy_sold += trade["energy"]` in `ForwardDeviceStats.add_trade()`. So the result of this function should also be input in the next market slot, right?
Or am I missing something here?

gsy_framework/sim_results/electric_blue/aggregate_results.py
```diff
@@ -0,0 +1,145 @@
...
+            trades=self.trades + other.trades
+        )
+
+    def to_dict(self) -> Dict:
+        """Generate a dictionary for saving the data into DB."""
+        return {f.name: getattr(self, f.name) for f in fields(self)}
+
+    def add_trade(self, trade: Dict) -> None:
+        """Add trade information to device stats."""
+        assert trade["timeslot"] == self.timeslot
```
cruky, line 80:
```suggestion
        assert trade["time_slot"] == self.timeslot
```
Here is a typo probably. There are a few below as well.

gsy_framework/sim_results/electric_blue/aggregate_results.py
mrsaemir: The market_stats is exactly like what you mentioned.
Also, trades of the same forward slot that have happened in the previous slot are buffered in `ForwardDeviceStats.trades`. That's exactly why we need to sum the previous ForwardDeviceStats with the current one. In other words, the result of the last slot should be feeded into the next one.

gsy_framework/sim_results/electric_blue/time_series.py
mrsaemir: Fixed.

gsy_framework/sim_results/electric_blue/aggregate_results.py
mrsaemir: Thanks for mentioning. I fixed them all.

gsy_framework/sim_results/electric_blue/aggregate_results.py
spyrostz: I think that it is better to save them in the database because as Hannes mentioned, we need to make high resolution profiles available very quickly.  Have a look at this selector on the energy bought / sold graph: 

Screenshot 2022-09-14 at 10 49 39

In order to do this reasonably fast, as our experience with d3a.io showed, it needs to be precalculated.

gsy_framework/sim_results/electric_blue/aggregate_results.py
hannesdiedrich: But I do not see, where the `why we need to sum the previous ForwardDeviceStats with the current one` is done. In `handle_forward_results` the `results` is an empty dict at the beginning of the method. Can you please help me understand where the previous results are buffered? Thanks!

gsy_framework/sim_results/electric_blue/aggregate_results.py
```diff
@@ -0,0 +1,147 @@
...
+# what aggregations are needed for each market type.
+MARKET_RESOLUTIONS = {
+    AvailableMarketTypes.YEAR_FORWARD: [AggregationResolution.RES_1_MONTH],
+    AvailableMarketTypes.MONTH_FORWARD: [AggregationResolution.RES_1_DAY],
+    AvailableMarketTypes.WEEK_FORWARD: [AggregationResolution.RES_1_DAY]
+}
+
+
+@dataclass
+class ForwardDeviceStats:  # pylint: disable=too-many-instance-attributes
```
spyrostz, line 19: Nothing to do here, but I have an almost-duplicate dataclass in gsy-web. I will import it in gsy-web from here if possible, once this PR is merged in order to not have duplicated code.

gsy_framework/sim_results/electric_blue/aggregate_results.py
mrsaemir: Adding the previous results to the current results is not done here. It will be done by @cruky. In order to do that:
1- Fetch the last results from DB
2- Create the ForwardDeviceStats object from last results
3- Using the __add__ magic method, he will be able to do: last_results + current_results
4- Save the updated results into the DB to be used for next market slot.

I only provide the __add__ method so that Kamil can easily accumulate the results.

hannesdiedrich: Thanks a lot for explaining!

mrsaemir: The `handle_forward_results` only creates the current results. Handling the last results would be another ticket in the future.

gsy_framework/sim_results/electric_blue/aggregate_results.py
mrsaemir: You may also want to add some more properties to this class such as: `executed_trades`.

gsy_framework/sim_results/electric_blue/time_series.py
```diff
@@ -0,0 +1,148 @@
...
+    def _generate_open_sell_orders_kWh(self, resolution: Duration) -> Dict:
+        """Generate time series for all offers."""
+        if not self.device_stats.open_offers:
+            return {}
+        total_offered_energy = sum([offer["energy"] for offer in self.device_stats.open_offers])
+        trade_profile_generator = ForwardTradeProfileGenerator(
+            peak_kWh=total_offered_energy)
+        open_sell_profile = trade_profile_generator.generate_trade_profile(
+            energy_kWh=total_offered_energy,
+            market_slot=DateTime.fromisoformat(self.device_stats.time_slot),
```
spyrostz, line 97: I think that it is better to unify our datetime conversions, in order to use the same methods throughout our codebase. Therefore how about replacing this with `str_to_pendulum_datetime` ?

mrsaemir: Regarding the other discussion in [this](https://github.com/gridsingularity/gsy-e/pull/1530#discussion_r970419992) PR, I'm about to convert them all to pendulum DateTime objects.

gsy_framework/sim_results/electric_blue/time_series.py
```diff
@@ -0,0 +1,148 @@
...
+        return {
+            "matched_buy_orders_kWh": self._generate_matched_buy_orders_kWh(resolution),
+            "matched_sell_orders_kWh": self._generate_matched_sell_orders_kWh(resolution),
+            "open_sell_orders_kWh": self._generate_open_sell_orders_kWh(resolution),
+            "open_buy_orders_kWh": self._generate_open_buy_orders_kWh(resolution),
+            "all_buy_orders_KWh": self._generate_all_buy_orders_kWh(resolution),
+            "all_sell_orders_kWh": self._generate_all_sell_orders(resolution)
+        }
+
+    def _generate_matched_buy_orders_kWh(self, resolution: Duration) -> Dict:
```
spyrostz, line 60: I think that the implementation of all these methods could be refactored so that the code can be a bit more DRY. If I see correctly, the construction of the ForwardTradeProfileGenerator, the call to the generate_trade_profile method and the resampling can be abstracted into a single method and used by all 6 respective methods.

gsy_framework/sim_results/electric_blue/time_series.py
```diff
@@ -0,0 +1,148 @@
...
+    result = {}
+
+    sorted_time_slots = sorted(timeseries_data.keys())
+    start_time = sorted_time_slots[0]
+    next_time = start_time + resolution
+
+    time_slots_to_aggregate = []
+    for time_slot in sorted_time_slots:
+        if time_slot >= next_time:
+            result[time_slots_to_aggregate[0]] = aggregator_fn(
```
spyrostz, line 26: I keep on wondering about the possibility to need a different aggregation function, however does not hurt to leave it that way in order to have that possibility in the future.

spyrostz (approved):
One comment, LGTM in general.

gsy_framework/sim_results/electric_blue/time_series.py
mrsaemir: Fixed.

gsy_framework/sim_results/electric_blue/time_series.py
mrsaemir: Fixed.

spyrostz (dismissed):
LGTM, thanks!

cruky (approved):
LGTM! Thank you!

spyrostz (approved):
LGTM

PR #393, opened 2022-09-20
GSYE-392: Aggregated SSP in different resolutions

mrsaemir:
As requested by [GSYE-392](https://gridsingularity.atlassian.net/browse/GSYE-392), we need aggregated SSP in different resolutions to support to below time series graph:

Screenshot from 2022-09-20 13-07-13

## Proposed changes:
- Pre-calculating SSP in different resolutions
- Adding classes to adap start_time, end_time of the graph to the pre-calculated profiles
- Adding `DeviceVolumeTimeSeries` class to calculate device volume time series
- Convert `resample_data` to a generator (will break a small part of Kamil's implementation)

## Remaining changes:
- [x] Adding unit tests


hannesdiedrich (commented):
I have left several question for understanding. I would like to give you the opportunity to answer them before I continuing the review. 
Thanks a lot in advance!

gsy_framework/forward_markets/aggregated_profile.py
```diff
@@ -0,0 +1,170 @@
...
+class MonthlyAggregatedSSPProfile(AggregatedSSPProfileBase):
+    """Return monthly-aggregated profile of the SSP."""
+    SSP_AGGREGATED_AGGREGATED_PROFILE_PATH = RESOURCES_PATH / "aggregated_ssp/non_leap_monthly.csv"
+    LEAP_YEAR_SSP_AGGREGATED_PROFILE_PATH = RESOURCES_PATH / "aggregated_ssp/leap_monthly.csv"
+
+    def _get_timeslots(self, start_time: DateTime, end_time: DateTime) -> Iterable:
+        assert end_time - start_time >= duration(months=1), "Time period should be >= 1 month."
+        return period(
+            start=start_time.start_of("month"),
+            end=end_time.start_of("month") - duration(days=1)
```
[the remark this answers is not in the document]
mrsaemir: As a general rule in this design, the `_get_timeslots` will always contain the start_date and exclude the end_date:
```
range = [start_date, end_date)
```
I couldn't find such a behavior in the range function, so I manually subtracted 1 day out of a month to exclude the end_date.
Thus, the result of calling 
```
  aggregated_profile = MonthlyAggregatedSSPProfile(capacity_kWh=1)
  start_time = DateTime(2020, 2, 1, 0, 0)
  end_time = DateTime(2020, 4, 1, 0, 0)

  profile = list(aggregated_profile.generate(start_time, end_time))
  print(profile)
 ```
 would be this:
 ```
[
  (DateTime(2020, 2, 1, 0, 0, 0), 373.2430992258762),
  (DateTime(2020, 3, 1, 0, 0, 0), 598.5182483021474)
]
 ```

File: gsy_framework/forward_markets/aggregated_profile.py [hunk 2]
```diff
@@ -0,0 +1,170 @@
...
+        ).range("months", 1)
```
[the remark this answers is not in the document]
mrsaemir: The number of returned `time_slot`s will depend on `start_date` and `end_date` e.g. there could be a year of duration between start and end times.

gsy_framework/forward_markets/aggregated_profile.py
```diff
@@ -0,0 +1,170 @@
...
+    def _get_timeslot_energy_kWh(self, timeslot: DateTime) -> float:
+        return sum([
+            float(self._SSP_AGGREGATED_PROFILE[str(t.month)])
+            for t in period(timeslot, timeslot.add(days=6)).range("days", 1)
+        ])
+
+    def generate(self, start_time: DateTime, end_time: DateTime) -> Iterable:
+        """Generate SSP profile with respect to start and end times in the correct resolution."""
+        for timeslot in self._get_timeslots(start_time, end_time):
+            yield timeslot, self._get_timeslot_energy_kWh(timeslot) * self.capacity_kWh
```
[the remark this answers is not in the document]
mrsaemir: You're absolutely right!

gsy_framework/forward_markets/aggregated_profile.py
```diff
@@ -0,0 +1,170 @@
+import csv
```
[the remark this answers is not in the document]
mrsaemir: It can be renamed to aggregated_ssp_profile. Thanks

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,158 @@
...
+    AggregationResolution.RES_1_MONTH: "month",
+    AggregationResolution.RES_1_YEAR: "year",
+}
+
+
+class AssetVolumeTimeSeries:
+    """This class generates combined volume time series for the whole year for each asset.
+    The result would be like this for a monthly resolution for the year 2020:
+    {
+        DateTime(2020, 1, 1, 0, 0, 0): {
```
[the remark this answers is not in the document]
mrsaemir: It's intended. The first level is always the year.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def add(self, asset_stats: ForwardDeviceStats, market_type: AvailableMarketTypes):
+        """Add asset time series to asset volume time series."""
+
+        self._add_total_energy_bought(asset_stats, market_type)
+        self._add_total_energy_sold(asset_stats, market_type)
+
+    def save_time_series(self):
+        """Save asset volume time series in the DB."""
+        # TODO: should be implemented by Kamil.
```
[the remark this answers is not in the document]
mrsaemir: Done.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,158 @@
...
+            time_series = self._fetch_asset_volume_time_series_from_db(time_slot)
+            if time_series is None:
+                time_series = {
+                    ts: self._get_time_series_template(value)
+                    for ts, value in self._generate_asset_volume_time_series(time_slot)}
+            self._asset_volume_time_series_buffer[time_slot] = time_series
+        return time_series
+
+    def _fetch_asset_volume_time_series_from_db(self, time_slot: DateTime):
+        # TODO: should be implemented by Kamil.
```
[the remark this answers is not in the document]
mrsaemir: I added some probable type hints, although it's not decided yet. Thanks

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,158 @@
...
+        if self.resolution == AggregationResolution.RES_15_MINUTES:
+            return time_slot.set(minute=(time_slot.minute // 15) * 15)
+        return time_slot.start_of(START_OF[self.resolution])
+
+    def _get_asset_volume_time_series(self, time_slot):
+        """Return asset volume time series for the required time slot.
+        If not found in the buffer, it tries fetching it from DB.
+        If not found in the DB, it will generate a new one for the whole year."""
+        if time_slot in self._asset_volume_time_series_buffer:
+            time_series = self._asset_volume_time_series_buffer[time_slot]
```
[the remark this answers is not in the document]
mrsaemir: That's true.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,158 @@
...
+        # TODO: should be implemented by Kamil.
+        return None
+
+    def _generate_asset_volume_time_series(self, time_slot: DateTime):
+        """Generate asset volume time series for the whole year."""
+        start_time = self._adapt_time_slot(time_slot)
+        if start_time < time_slot:
+            start_time += self.resolution.duration()
+
+        end_time = self._adapt_time_slot(time_slot.add(years=1))
```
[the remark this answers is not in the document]
mrsaemir: The time series will always contain a complete year (from it's beginning to it's end). So, if you start from 1st of March this year unitl 1st of March next year, you will get time series for 2 complete years!

Now, if it's the first time, the whole time series will be generated. Otherwise it will be fetched from DB.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,158 @@
...
+# a dictionary showing the start of time slot for each resolution.
+START_OF = {
+    AggregationResolution.RES_1_HOUR: "hour",
+    AggregationResolution.RES_1_WEEK: "week",
+    AggregationResolution.RES_1_MONTH: "month",
+    AggregationResolution.RES_1_YEAR: "year",
+}
+
+
+class AssetVolumeTimeSeries:
```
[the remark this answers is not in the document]
mrsaemir: That's true. The GraphQL API should be able to easily fetch the data without any further processing.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,158 @@
...
+                    "bought_kWh": 201.08320595131318,
+                    "sold_kWh": 201.08320595131318,
+                },
+                "WEEK_FORWARD": {
+                    "bought_kWh": 32.432775153437625,
+                    "sold_kWh": 32.432775153437625,
+                },
+                "DAY_FORWARD": {
+                    "bought_kWh": 6.486555030687522,
+                    "sold_kWh": 6.486555030687522,
```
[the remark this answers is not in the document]
mrsaemir: each *_FORWARD represents the produced/consumed energy of that specific product in the context of the resolution.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -127,8 +127,10 @@ def _get_asset_volume_time_series(self, time_slot):
             self._asset_volume_time_series_buffer[time_slot] = time_series
         return time_series
 
-    def _fetch_asset_volume_time_series_from_db(self, time_slot: DateTime):
-        # TODO: should be implemented by Kamil.
+    def _fetch_asset_volume_time_series_from_db(
+            self, time_slot: DateTime) -> Union[Dict[DateTime, Dict], None]:
```
hannesdiedrich, line 131: Nit: I think the following is correct:
```python
    -> Optional[Dict[DateTime, Dict]]:
```

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
hannesdiedrich: I do not understand that. Why is the 
MONTH_FORWARD/bought_kWh = 201 and WEEK_FORWARD/bought_kWh= 32 ?
A consumed energy must always connected to a timeframe when it was consumed. 

Taking your answer literally, the dictionary should contain timestamps for each resolution. But I only see a daily resolution. Can you please explain a bit more verbose? Thanks!

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
hannesdiedrich: Is this really needed? This nesting will not be represented in the DB anyway right?

gsy_framework/forward_markets/aggregated_profile.py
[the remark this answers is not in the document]
hannesdiedrich: Sorry to be even more nitpicky, but the `p` in `ssp` already stands for profile.

gsy_framework/enums.py
```diff
@@ -50,19 +50,34 @@ class AvailableMarketTypes(Enum):
...
+            self.SPOT: "SPOT",
+            self.BALANCING: "BALANCING",
+            self.SETTLEMENT: "SETTLEMENT",
+            self.FUTURE: "FUTURE",
+            self.DAY_FORWARD: "DAY_FORWARD",
+            self.WEEK_FORWARD: "WEEK_FORWARD",
+            self.MONTH_FORWARD: "MONTH_FORWARD",
+            self.YEAR_FORWARD: "YEAR_FORWARD",
+            self.INTRADAY: "INTRADAY"
+        }[self]
```
fievelk, line 64: Is there a reason why you don't want to use `.name`? For example:
```py
>>> market_type = AvailableMarketTypes(3)
>>> market_type.name
# FUTURE
```

fievelk: Another note: I think that we have some naming issue here: we're mixing market types and product types. It would be nice to pick just one of the two terms (or decide what's the difference between one and the other). @spyrostz FYI

gsy_framework/forward_markets/aggregated_profile.py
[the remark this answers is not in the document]
hannesdiedrich: I see now, thanks!

gsy_framework/forward_markets/aggregated_ssp_profile.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class AggregatedSSPProfileBase(ABC):
+    """Base class representing the Standard Solar Profile in different resolutions."""
+
+    SSP_AGGREGATED_AGGREGATED_PROFILE_PATH: Path = None
+    LEAP_YEAR_SSP_AGGREGATED_PROFILE_PATH: Path = None
+
+    def __init__(self, capacity_kWh: float):
+        self.capacity_kWh = capacity_kWh
```
fievelk, line 22: There seem to be a few pylint warnings in this file. Could you check and maybe disable the unnecessary ones?

gsy_framework/forward_markets/aggregated_ssp_profile.py
```diff
@@ -0,0 +1,165 @@
...
+    def _get_timeslot_energy_kWh(self, timeslot: DateTime) -> float:
+        if timeslot.is_leap_year():
+            return float(self._LEAP_YEAR_SSP_AGGREGATED_PROFILE[""])
+        return float(self._SSP_AGGREGATED_PROFILE[""])
+
+
+def get_aggregated_SSP_profile(
+        capacity_kWh: float, start_time: DateTime,
+        end_time: DateTime, resolution: AggregationResolution
+):
```
fievelk, line 149: Nit: I'd suggest staying consistent with the rest of the codebase and put the `):` at the end of the previous line:
```suggestion
def get_aggregated_SSP_profile(
        capacity_kWh: float, start_time: DateTime,
        end_time: DateTime, resolution: AggregationResolution):
```

gsy_framework/forward_markets/aggregated_ssp_profile.py
```diff
@@ -0,0 +1,165 @@
...
+    """Return aggregated SSP profile with the specified resolution and with respect to
+    start_time, end_time and asset capacity."""
+    resolution_mapping = {
+        AggregationResolution.RES_15_MINUTES: QuarterHourAggregatedSSPProfile,
+        AggregationResolution.RES_1_HOUR: HourlyAggregatedSSPProfile,
+        AggregationResolution.RES_1_WEEK: WeeklyAggregatedSSPProfile,
+        AggregationResolution.RES_1_MONTH: MonthlyAggregatedSSPProfile,
+        AggregationResolution.RES_1_YEAR: YearlyAggregatedSSPProfile
+    }
```
fievelk, line 158: Nit: I don't know how often this function will be called, but if you want we could move the instantiation of the dictionary outside the function definition (at module level). But it's just a very small optimization/nit.

tests/test_sim_results/electric_blue/test_volume_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+                            DateTime(2020, 1, 1, 0, 0, 0),
+                            DateTime(2020, 1, 1, 0, 0, 0).end_of("day"),
+                            1
+                        ), "sold_kWh": sum_of_energy(
+                            DateTime(2020, 1, 1, 0, 0, 0),
+                            DateTime(2020, 1, 1, 0, 0, 0).end_of("day"),
+                            1
+                        )}}},
+            DateTime(2019, 1, 1, 0, 0, 0): {
+                DateTime(2019, 1, 1, 0, 0, 0): {
```
hannesdiedrich, line 135: I am really confused now. Why is there an entry for 2019 here and only for the WEEK_FORWARD?

tests/test_sim_results/electric_blue/test_volume_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+                            1
+                        ),
+                        "sold_kWh": sum_of_energy(
+                            DateTime(2020, 1, 1, 0, 0, 0),
+                            DateTime(2020, 1, 1, 0, 0, 0).end_of("month"),
+                            1
+                        )
+                    },
+                    "WEEK_FORWARD": {
+                        "bought_kWh": sum_of_energy(
```
hannesdiedrich, line 115: So, do I get it right: This is the sum of energy that has been bought in the WEEK_FORWARD markets for the whole year?

hannesdiedrich: Isn't the user interested in a profile of the weekly markets?

gsy_framework/sim_results/electric_blue/aggregate_results.py
```diff
@@ -10,9 +10,31 @@
 # Used by forward markets; the following dictionary defines
 # what aggregations are needed for each market type.
 MARKET_RESOLUTIONS = {
```
fievelk, line 12: Does this need to be in gsy-framework, or can we move it to gsy-e?

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,160 @@
...
+        # TODO: should be implemented.
+        raise NotImplementedError()
+
+    def _add_total_energy_bought(
+            self, asset_stats: ForwardDeviceStats, market_type: AvailableMarketTypes):
+        """Add statistics for the total amount of bought energy."""
+        energy_kWh = asset_stats.total_energy_bought
+        if energy_kWh <= 0:
+            return
+        profile = ForwardTradeProfileGenerator(self.asset_capacity).generate_trade_profile(
```
fievelk, line 84: Since `self.asset_capacity` is always the same, it looks like we can instantiate the `ForwardTradeProfileGenerator` just once as an instance variable (without repeating it everytime `_add_total_energy_bought()` and `_add_total_energy_sold()` are called).

tests/test_sim_results/electric_blue/test_volume_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+            )
+    return func()
+
+
+class TestAssetVolumeTimeSeries:
+    @staticmethod
+    def check_time_slot_pairs(time_slots: List, resolution: AggregationResolution):
+        """Check the returned value of _adapt_time_slot is equal to the expected value."""
+        volume_time_series = AssetVolumeTimeSeries(
+            "UUID", 1, resolution)
```
fievelk, line 34: Here and in the rest of the module, I'd recommend using named parameters to instantiate this class:
```suggestion
        volume_time_series = AssetVolumeTimeSeries(
            asset_uuid="UUID", 
            asset_capacity=1, 
            resolution=resolution)
```
This makes it easier to refactor in the future. 
Using positional parameters makes sense in some cases (e.g. when the name of the function makes it obvious what they represent, like `multiply(2, 4)`) but in this case we have no idea what `1` is, for example.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,160 @@
...
+            DateTime(2020, 12, 1, 0, 0, 0): {...},
+        },
+    }
+    """
+    def __init__(
+            self, asset_uuid: str, asset_capacity: float,
+            resolution: AggregationResolution):
+
+        self.asset_uuid = asset_uuid
+        self.asset_capacity = asset_capacity
```
fievelk, line 62: What's the unit of the capacity here? I'd recommend specifying it, e.g.:
```suggestion
        self.asset_capacity_kWh = asset_capacity
```
That's especially useful since the `ForwardProfileGeneration` works with kWh.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
mrsaemir: Each timeslot shows the amount of each product sold/bought only in that slot. In this example only 201 kWh of energy was sold for the monthly product in 2020 which can be different from the total sold/bought weekly product.

gsy_framework/forward_markets/aggregated_profile.py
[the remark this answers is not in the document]
mrsaemir: No worries. I fixed it.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
mrsaemir: It will not be represented in the DB like this, but it's needed here. Let me explain more ...

Consider you want to update the graphs with a new trade just received from the weekly product.
The timeslot for the weekly product is: `2019-12-30T00:00`. Part of the sold energy for the weekly product is in 2019 and the rest of it is in 2020.

Now consider we want the monthly resolution of the device volume time series which starts from 1s of Jan; should we add the energy of the weekly product to Jan 2020? Should we add it to December 2019?

IMO, we should split the weekly product into smaller time quantoms and then distributed each quantom regarding to its year.

And for this purpose, we need volume time series of both 2019 and 2020. So we fetch them both (on demand) and save them after updating them.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
mrsaemir: They are really the same but sure I'll do it in order to have the same convention in the whole repo. Thanks

gsy_framework/forward_markets/aggregated_ssp_profile.py
mrsaemir: Actually I've thought of it before, but the performance boost will be negligible as this function is only called once for each graph (while generating it for the first time).

tests/test_sim_results/electric_blue/test_volume_timeseries.py
mrsaemir: I've explained it here: https://github.com/gridsingularity/gsy-framework/pull/393#discussion_r983067160

tests/test_sim_results/electric_blue/test_volume_timeseries.py
mrsaemir: This is the yearly resolution of the profile. If one is interested in a weekly resolution, they shall check another graph!

gsy_framework/sim_results/electric_blue/aggregate_results.py
mrsaemir: gsy-results will make use of it for its implementation. If we move it to gsy-e, then gsy-results can't access it!

gsy_framework/sim_results/electric_blue/volume_timeseries.py
mrsaemir: Makes sense. Thanks for pointing it out.

tests/test_sim_results/electric_blue/test_volume_timeseries.py
mrsaemir: Fixed.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
hannesdiedrich: Ok, thanks. But then I do not see the connection to the UI picture provided in the description.
I see a yearly profile there, but really, only 4x2 numbers are calculated here. How will the profile be created out of these?

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
hannesdiedrich: IMO adding the week that starts at `2019-12-30T00:00` to 2019 is sufficient. No need to complicate the code only for one corner case. But we need to discuss this in the team.

tests/test_sim_results/electric_blue/test_volume_timeseries.py
hannesdiedrich: Ahhhh, I think this answer made it all clear to me. let me re-review.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
hannesdiedrich: Thanks to your answer [here](https://github.com/gridsingularity/gsy-framework/pull/393#discussion_r984363388) it is clear to me now. Thanks!

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
mrsaemir: Calculating for the whole year actually makes the code look simpler.
If we don't do that, then when other 2019 weeks come to get added, we have to recalculate time series for 2019 to compensate for the missing weeks (and this can make code even more complex).

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,160 @@
...
+        },
+    }
+    """
+    def __init__(
+            self, asset_uuid: str, asset_peak_kWh: float,
+            resolution: AggregationResolution):
+
+        self.asset_uuid = asset_uuid
+        self.asset_peak_kWh = asset_peak_kWh
+        self.resolution = resolution
```
hannesdiedrich, line 62: So there will be one `AssetVolumeTimeSeries` per resolution? So I understand that not all resolutions have to be updated every time a new trade has happened, right? So we could update the weekly resolution only if a new week has started, right?
Or are you planning to run `add()` after every trade?

mrsaemir: Yes. There will be one AssetVolumeTimeSeries per resolution and per asset.
No. All resolutions (of the specified device) will get updated if you receive a trade.
Also, not all years get updated when you receive a trade.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,160 @@
...
+        for time_slot, value in profile.items():
+            self._add_to_volume_time_series(time_slot, value, market_type, "sold_kWh")
+
+    def _add_to_volume_time_series(
+            self, time_slot: DateTime, value: float,
+            market_type: AvailableMarketTypes, attribute_name: str):
+        """Add time slot value to the correct time slot to asset volume time series."""
+        time_slot = self._adapt_time_slot(time_slot)
+        volume_time_series = self._get_asset_volume_time_series(time_slot.start_of("year"))
+        volume_time_series[time_slot][market_type.name][attribute_name] += value
```
hannesdiedrich, line 107: As discussed offline, the time_slot here has a different meaning. It is the key that divides the data into yearly chunks. So I suggest to rename it here into e.g. `year_selector` or`year_key` or similar.
Also, i think there is a bug here, because also in Line 107 you need to select the start of the year, right?
```suggestion
        time_slot = self._adapt_time_slot(time_slot)
        year_key = time_slot.start_of("year")
        volume_time_series = self._get_asset_volume_time_series(year_key)
        volume_time_series[year_key][market_type.name][attribute_name] += value
```

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def _adapt_time_slot(self, time_slot: DateTime) -> DateTime:
+        """Find the containing time slot of the smaller time slot."""
+        if self.resolution == AggregationResolution.RES_15_MINUTES:
+            return time_slot.set(minute=(time_slot.minute // 15) * 15)
+        return time_slot.start_of(START_OF[self.resolution])
+
+    def _get_asset_volume_time_series(self, time_slot):
```
hannesdiedrich, line 115: Here also, please rename `time_slot` in the whole method. Thanks!

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,160 @@
...
+            self._asset_volume_time_series_buffer[time_slot] = time_series
+        return time_series
+
+    def _fetch_asset_volume_time_series_from_db(
+            self, time_slot: DateTime) -> Optional[Dict[DateTime, Dict]]:
+        """Fetch already saved asset volume time series."""
+        # TODO: should be implemented.
+        return None
+
+    def _generate_asset_volume_time_series(self, time_slot: DateTime):
```
hannesdiedrich, line 136: AFAIU, this method "only" creates a time series in the self.resolution that contains the SSP values. It does not return the volume time_series, right? So I suggest to rename the method (to e.g. `generate_SSP_time_series`) and adapt the docstring.

tests/test_sim_results/electric_blue/test_volume_timeseries.py
```diff
@@ -0,0 +1,203 @@
...
+        volume_time_series = AssetVolumeTimeSeries(
+            asset_uuid="UUID", asset_peak_kWh=5, resolution=AggregationResolution.RES_1_MONTH)
+
+        try:
+            for market_type, asset_stats in forward_time_series:
+                volume_time_series.add(asset_stats, market_type)
+        except KeyError as exc:
+            pytest.fail(str(exc))
+
+        self.check_time_slots_year(volume_time_series)
```
hannesdiedrich, line 164: Minor, but maybe valuable would be to add a more detailed check here. E.g. an easy test could be to check the number of time steps present in the nested dict. For example the number of time stamps for the monthly resolution should be 12 (and equivalent in the other resolutions too. What do you think?

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,160 @@
...
+            resolution: AggregationResolution):
+
+        self.asset_uuid = asset_uuid
+        self.asset_peak_kWh = asset_peak_kWh
+        self.resolution = resolution
+
+        self._asset_volume_time_series_buffer: Dict[DateTime, Dict] = {}
+        self._trade_profile_generator = ForwardTradeProfileGenerator(self.asset_peak_kWh)
+
+    def add(self, asset_stats: ForwardDeviceStats, market_type: AvailableMarketTypes):
```
hannesdiedrich, line 67: The there is an instance of `ForwardDeviceStats` for each 15 min time slot, right?

hannesdiedrich: Also: Here you use `market_type` (which I prefer) but it was decided to use `product_type` right?

hannesdiedrich: Last but not least: How about rename this method to be more descriptive: e.g. `add_results_to_time_series` or `add_trade_to_time_series`?

hannesdiedrich (commented):
Thanks a lot for the offline explanation! It helped a lot. 
I left some more comments for improving the readability (from my point of view). 
In general: LGTM

gsy_framework/sim_results/electric_blue/volume_timeseries.py
mrsaemir: Thanks. I renamed it to `year`:
```python
    def _get_asset_volume_time_series(self, year: DateTime):
```
Also, it's not a bug as the function returns the requested volume time series (as the name `_get_...` suggests).

gsy_framework/sim_results/electric_blue/volume_timeseries.py
mrsaemir: Fixed here: https://github.com/gridsingularity/gsy-framework/pull/393#discussion_r986527958

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,160 @@
...
+# a dictionary showing the start of time slot for each resolution.
+START_OF = {
+    AggregationResolution.RES_1_HOUR: "hour",
+    AggregationResolution.RES_1_WEEK: "week",
+    AggregationResolution.RES_1_MONTH: "month",
+    AggregationResolution.RES_1_YEAR: "year",
+}
+
+
+class AssetVolumeTimeSeries:
```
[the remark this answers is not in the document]
mrsaemir: I can't be sure about the scheme of the DB now but I'll try to add something abstract to demonstrate it. Thanks

gsy_framework/sim_results/electric_blue/volume_timeseries.py
mrsaemir: Fixed.

tests/test_sim_results/electric_blue/test_volume_timeseries.py
mrsaemir: Actually this is done [here](https://github.com/gridsingularity/gsy-framework/blob/c79701368431749b81f8b90d342da71a3ec3a083/tests/forward_markets/test_aggregated_profile.py). But as you just mentioned, more assertions could be added to make it safer. Let me know if you had any other ideas on this. Thanks

gsy_framework/sim_results/electric_blue/volume_timeseries.py
mrsaemir: - There's an instance of the `ForwardDeviceStats` every 15 minutes but only ones that have new trades are taken into account.
- I think `market_type` is as valid as `product_type` as the enum itself is called: `AvailableMarketTypes`. What do you think? Should we change it to product_type anyway? 
- I'll rename it to `update_volume_time_series`.

hannesdiedrich: > I think market_type is as valid as product_type as the enum itself is called: AvailableMarketTypes. What do you think? Should we change it to product_type anyway

IMO `product_type` is not a good naming. But now as it was decided to used this name, lets be consistent throughout the code. Indeed, this means als to rename `AvailableMarketTypes`. This all can be done in a refactoring PR IMO.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -0,0 +1,167 @@
...
+        else:
+            time_series = self._fetch_asset_volume_time_series_from_db(year)
+            if time_series is None:
+                time_series = {
+                    ts: self._get_time_series_template(value)
+                    for ts, value in self._generate_SSP_time_series(year)}
+            self._asset_volume_time_series_buffer[year] = time_series
+        return time_series
+
+    def _fetch_asset_volume_time_series_from_db(
```
cruky, line 136: Could you elaborate on this method, how do you see the implementation for reading and writing to db? Since gsy-web is responsible for interacting with db, we were using `ResultsHandler` and it's [update](https://github.com/gridsingularity/gsy-framework/blob/e10b82e797ab44bfa612d4a81b26bc43dd315dfa/gsy_framework/sim_results/all_results.py#L71) method to transfer data. We provide it with the core stats, and we could provide it with the pre-fetched data, but do you expect to have more interaction with db, like query it from this method?

mrsaemir: This method needs to fetch data from DB. So, yes! You need to run queries from this class. The class also needs to save the updated results by another query. Pre-fetching can't be a solution because we don't know what to fetch at the beginning.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
mrsaemir: Fixed.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
mrsaemir: Umm, it's returning the time series actually: At the end of the function ...

hannesdiedrich: Is github failing me again?
I do not see a return statement. 
Screenshot 2022-10-04 at 10 58 10

mrsaemir: The return statement is inside `_get_asset_volume_time_series`. After time series is returned, we assign it to `volume_time_series` and then modify it.

gsy_framework/forward_markets/aggregated_ssp.py
```diff
@@ -0,0 +1,164 @@
...
+        ).range("hours", 1)
+
+    def _get_timeslot_energy_kWh(self, timeslot: DateTime) -> float:
+        return float(self._SSP_AGGREGATED_PROFILE[timeslot.format("M-H")])
+
+
+class WeeklyAggregatedSSPProfile(AggregatedSSPProfileBase):
+    """Return weekly-aggregated profile of the SSP."""
+
+    SSP_AGGREGATED_AGGREGATED_PROFILE_PATH = RESOURCES_PATH / "aggregated_ssp/daily.csv"
```
cruky, line 94: Sorry, could you help me here, why we have daily ssp for weekly aggregation?

gsy_framework/sim_results/electric_blue/volume_timeseries.py
hannesdiedrich: Yeah, but the `self._asset_volume_time_series_buffer` is not adapted here AFAIU.
IMO, the 
``` python
self._asset_volume_time_series_buffer[year] = time_series
```
should be performed after
```python
volume_time_series[str(time_slot)][product_type.name][attribute_name] += value
```
Even if this works because there is a reference, to the `_asset_volume_time_series_buffer` I suggest to work with return values. At some point we decided to not use references if possible but to return and assign. 
I hope this makes sense

gsy_framework/forward_markets/aggregated_ssp.py
mrsaemir: This class uses the daily aggregation of the ssp to generate the weekly aggreation of it!
It wasn't possible to pre-calculate weekly aggregation of the ssp.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
hannesdiedrich: So, what I mean is:
``` python
def _add_to_volume_time_series(
        self, time_slot: DateTime, value: float,
        product_type: AvailableMarketTypes, attribute_name: str):
    """Add time slot value to the correct time slot to asset volume time series."""
    time_slot = self._adapt_time_slot(time_slot)
    year = time_slot.start_of("year")
    volume_time_series = self._get_asset_volume_time_series(year=year)
    volume_time_series[str(time_slot)][product_type.name][attribute_name] += value
    self._asset_volume_time_series_buffer[year] = volume_time_series
```

mrsaemir: Fixed.

tests/test_sim_results/electric_blue/test_volume_timeseries.py
hannesdiedrich: I see, you mean if the `*AggregatedSSPProfile.generate()` is broken, it will be cought by the tests you linked and does not have to be tested here again, because `AssetVolumeTimeSeries` is calling it. 
LGTM

hannesdiedrich (approved):
LGTM

cruky (approved):
LGTM

PR #394, opened 2022-09-27
GSYE-333: Do not terminate the two sided market clearing calculation …

spyrostz:
…of all time slots if one time slot does not have a clearing.


mrsaemir (approved):
LGTM

cruky (approved):

PR #395, opened 2022-10-06
EB-32: Added a constant for retrieving all available forward market t…

spyrostz:
…ypes.

In order to quickly iterate over all forward markets, it was proven handy to have a common list that would define the market types which can be used as forward markets.


hannesdiedrich (approved):
LGTM

PR #397, opened 2022-10-11
GSYE-374 Persist the EB simulation orders, device statistics and time series to the database 

cruky:
## Proposed Changes:

- Added `all_results` module with `ForwardResultsHandler` to electric blue simulation results
- `ForwardResultsHandler` behaves like a `ResultsHandler`, with `update` and `all_db_results`
- For each `forward_market_stats` in core stats which goes from the simulation, we: 
- - retrive the `bids, offers, trades` for each `area_uuid`, `market_type` and `time_slot` and name it "**orders**"
- - accumulate orders for each device with `handle_forward_results`
- - using previous device stats from db we calculate new  `ForwardDeviceStats` depends on the market type
- - depends on resolution for market types, we generate all `ForwardDeviceTimeSeries`
- The above results are written to attributes: `orders`, `current_device_stats`, `device_time_series` as nested dictionaries
- Unit tests added

## Future to do:
- When `ForwardTradeProfileGenerator` profile generation functionality will be extend with all forward markets types, we can delete check if market_type id in ALLOWED_MARKET_TYPES
- Improve `total_memory_utilization_kb`


gsy_framework/sim_results/electric_blue/aggregate_results.py
```diff
@@ -103,6 +104,17 @@ def to_dict(self) -> Dict:
         """Generate a dictionary for saving the data into DB."""
         return {f.name: getattr(self, f.name) for f in fields(self)}
 
+    @classmethod
+    def from_dict(cls, device_stats_dict: Dict) -> "ForwardDeviceStats":
+        """Return ForwardDeviceStats from dict, with deserialized DateTimes
+        from YYYY-MM-DDTHH:mm format."""
+        if time_slot := device_stats_dict.get("time_slot"):
```
mrsaemir, line 111: Since gsy-framework is also used in gsy-e. Using of the Walrus operator will cause PyPy to panic. So, I guess we should remove it from here.

gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -0,0 +1,103 @@
...
+    def total_memory_utilization_kb(self):
+        """Get the total memory allocated by the results."""
+        return self._total_memory_utilization_kb
+
+    def _clear_device_statistics(self) -> None:
+        self.orders.clear()
+        self.current_device_stats.clear()
+        self.device_time_series.clear()
+
+    def _get_bids_offers_trades(self, area_uuid: str, market_type: int,
```
mrsaemir, line 73: You may want to rename `_get_bids_offers_trades` to `_buffer_bids_offers_trades` as you're not returning anything.

mrsaemir (commented):
LGTM. However, I need to read it once again. Thanks a lot

gsy_framework/sim_results/electric_blue/aggregate_results.py
cruky: I thought that if we use this module only in gsy-web we are fine. But I can change it if it's otherwise.

gsy_framework/sim_results/all_results.py
```diff
@@ -21,7 +21,9 @@
 
 class ResultsHandler:
     """Calculate all results for each market slot."""
+
     def __init__(self, should_export_plots: bool = False, is_scm: bool = False):
+        self.forward_market_enabled = False
```
hannesdiedrich, line 26: I do not see this member being used. Can you please help me understand why it is needed?

gsy_framework/utils.py
```diff
@@ -238,6 +238,11 @@ def datetime_to_string_incl_seconds(date_time: DateTime) -> str:
     return date_time.format(DATE_TIME_FORMAT_SECONDS)
 
 
+def datetime_to_string(datetime_object: DateTime) -> str:
+    """Convert a datetime object into a string with YYYY-MM-DDTHH:mm format."""
```
hannesdiedrich, line 242: The type hint suggests, that the input type is a pendulum DateTime object. If this is true, lets rename the method to `pendulum_to_string` in order not to get confused by the "normal" DateTime object that has different and less properties.

gsy_framework/sim_results/electric_blue/time_series.py
```diff
@@ -21,7 +22,8 @@ def resample_data(
     time_slots_to_aggregate = []
     for time_slot in sorted_time_slots:
         if time_slot >= next_time:
-            yield time_slots_to_aggregate[0], aggregator_fn(
+            time_slot_str = time_slots_to_aggregate[0].format(DATE_TIME_FORMAT)
```
hannesdiedrich, line 25: Don't you want to re-use `datetime_to_string` here?

gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -0,0 +1,103 @@
...
+            str, Dict[int, Dict[DateTime, Dict]]] = defaultdict(lambda: defaultdict(dict))
+        self.previous_device_stats = {}
+        self.current_device_stats: Dict[
+            int, Dict[DateTime, Dict[str, Dict]]] = defaultdict(lambda: defaultdict(dict))
+        self.device_time_series: Dict[
+            int, Dict[int, Dict[DateTime, Dict[str, Dict]]]] = defaultdict(
+            lambda: defaultdict(lambda: defaultdict(dict)))
+        self._total_memory_utilization_kb = 0.0
+
+    # pylint: disable=unused-argument
```
hannesdiedrich, line 31: You can also prepend a `_` to `area_dict` to solve this warning

gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        Update the forward market aggregated results with bids_offers_trades of current slot.
+        """
+        self._clear_device_statistics()
+        for area_uuid, area_result in core_stats.items():
+            if "forward_market_stats" not in area_result:
+                return
+
+            current_market_dt = from_format(current_market, DATE_TIME_FORMAT) if (
+                current_market := current_market_slot) else ""
```
hannesdiedrich, line 42: I am not very familiar with the `:=` operator and do not really get what is done in this line here. Can you please help me understand? Thanks!

cruky: Wow, I think that it can be the remnant of previous implementations, anyway it does not make sense to use walrus operator here. Thanks for catching this!

gsy_framework/sim_results/all_results.py
cruky: This is something which probably could be improve. I added this attribute to differentiate between `ForwardResultsHandler` and `ResultsHandler`, to know if we are using forward market or not. It helps to tell in gsy-web which ResultsHandler should be used like [here](https://github.com/gridsingularity/gsy-web/blob/41fcacfdd8a8cffff2d3bc3fb071b8ba5d7ff206/simulation_results/results_calculation_helper.py#L94) or [here](https://github.com/gridsingularity/gsy-web/blob/41fcacfdd8a8cffff2d3bc3fb071b8ba5d7ff206/simulation_results/results_calculation_helper.py#L52).

gsy_framework/sim_results/electric_blue/aggregate_results.py
hannesdiedrich: I personally do not like this operator. Doing two things in on go is somehow I con not follow easily and I do not see the need here. What is the harm in doing the following? 
```suggestion
        time_slot = device_stats_dict.get("time_slot")
        if time_slot:
```

gsy_framework/sim_results/all_results.py
hannesdiedrich: you could use the following on the gsy-web side.
```python
isinsttance( self.simulation_results[data["job_id"]], ForwardResultsHandler)
```
But the fact that you have to add an if-block there shows to me, that the interface of the `ResultsHandler` and `ForwardResultsHandler` are not the same. This can probably be improved. Let me further review to check this out.

gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -0,0 +1,103 @@
...
+                                                                       MARKET_RESOLUTIONS)
+from gsy_framework.sim_results.electric_blue.time_series import ForwardDeviceTimeSeries
+
+
+class ForwardResultsHandler:
+    """Calculate all results for each market slot for forward markets."""
+
+    def __init__(self, slot_length: Duration):
+        self.forward_market_enabled = True
+        self.slot_length = slot_length
```
hannesdiedrich, line 20: This does not seem to be used in this class. Is it a placeholder for later?
If this is removed from here, it will also make the calling code on gsy-web side a little cleaner IMO.

gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def _clear_device_statistics(self) -> None:
+        self.orders.clear()
+        self.current_device_stats.clear()
+        self.device_time_series.clear()
+
+    def _get_bids_offers_trades(self, area_uuid: str, market_type: int,
+                                market_stats: Dict[str, Dict]):
+        for time_slot, orders in market_stats.items():
+            time_slot_dt = from_format(time_slot, DATE_TIME_FORMAT)
```
hannesdiedrich, line 76: There are a lot of forth and backs in the code between pendulum and string. I wonder if you can all remove them because you are not using the pendulums functionalities AFAIS. How about just using the already provided time_slot strings from the DB or from the results. Please correct me, if I overlooked something. Thanks!

gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -0,0 +1,103 @@
...
+                order: orders.get(order, []) for order in ("offers", "bids", "trades")}
+
+    def _update_stats_and_time_series(self, forward_results: Dict, market_type_value: int):
+        market_type = AvailableMarketTypes(market_type_value)
+        for time_slot, device_results in forward_results.items():
+            for device_uuid, current_device_stats in device_results.items():
+                if previous_forward_stats := self.previous_device_stats.get(
+                        market_type_value, {}).get(time_slot, {}).get(device_uuid, {}):
+                    previous_device_stats = ForwardDeviceStats.from_dict(previous_forward_stats)
+                    current_device_stats += previous_device_stats
```
hannesdiedrich, line 87: This works , because the `ForwardDeviceStats` has an `__add__` method implemented, right? 
This is great (only realising this now and wanted to share my joy :-)) !

gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -0,0 +1,103 @@
...
+                    # TODO: delete this check, when ForwardTradeProfileGenerator profile generation
+                    #  functionality will be extend with all forward markets types
+                    continue
+                for resolution in MARKET_RESOLUTIONS.get(market_type, []):
+                    device_time_series = ForwardDeviceTimeSeries(current_device_stats, market_type)
+                    all_time_series_generators = device_time_series.generate(
+                        resolution=resolution.duration())
+                    all_time_series = {k: dict(v) for k, v in all_time_series_generators.items()}
+                    self.device_time_series[market_type_value][resolution.value][time_slot][
+                        device_uuid] = all_time_series
```
hannesdiedrich, line 100: IMO, for better readability, this part could be refactored into a separate function. What do you hink?

gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -0,0 +1,103 @@
...
+            for market_type_value, market_stats in area_result["forward_market_stats"].items():
+                market_type = int(market_type_value)
+                self._get_bids_offers_trades(area_uuid, market_type, market_stats)
+                current_results = handle_forward_results(current_market_dt, market_stats)
+                self._update_stats_and_time_series(current_results, market_type)
+        self._update_memory_utilization()
+
+    def update_from_repr(self, area_representation: Dict):
+        """
+        Updates the simulation results using area_representation data that arrive from the d3a-web.
```
hannesdiedrich, line 52: NIT:
```suggestion
        Updates the simulation results using area_representation data that arrive from the gsy-web.
```
Can you please also replace this in the other occurrences? Thanks a lot

gsy_framework/sim_results/electric_blue/all_results.py
hannesdiedrich: Let's see how [this](https://github.com/gridsingularity/gsy-web/pull/1804#discussion_r995541361) discussion evolves. We may not need it.

gsy_framework/sim_results/electric_blue/aggregate_results.py
hannesdiedrich: Sorry, I did not want to dominate the discussion here. I see that you have already changed the code accordingly. So i guess that you are convinced? If you have a pro-argument of using the walrus operator here, please explain. Thanks a lot!

gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        Update the forward market aggregated results with bids_offers_trades of current slot.
+        """
+        self._clear_device_statistics()
+        for area_uuid, area_result in core_stats.items():
+            if "forward_market_stats" not in area_result:
+                return
+
+            current_market_dt = str_to_pendulum_datetime(current_market_slot) if (
+                current_market_slot) else ""
```
hannesdiedrich, line 41: the name already suggests, that it should be a DateTime object. So, writing an empty string to it in the else case, will not work. So, `current_market_slot` can be None, right? Can you please give a hint in the interface of the method: `Optional[str]`?  When is the current_market_slot empty? In the very first send of the results?
How should `handle_forward_results` behave in this case? It expects a DateTime object, right?

gsy_framework/sim_results/electric_blue/all_results.py
hannesdiedrich: I started a new discussion about these lines https://github.com/gridsingularity/gsy-framework/pull/397#discussion_r995747813
marking this as resolved though.

gsy_framework/sim_results/electric_blue/aggregate_results.py
spyrostz: Did not respond till now because it is a highly subjective matter (there are numerous flame wars online about the matter), however I am also with Hannes in this (and more specifically, I find that the walrus operator increases readability only in while loops). I might be overly biased by C++ where the assignment operator could be used in the same way as the Python walrus operator, but it was the source of a lot of hard-to-find-bugs, mainly because the control flow of the code was not evident to the developer on first glance.

cruky: Well, I've just used that to make the code shorter, but in fact it made the code harder to read in such simple case. Agree, this operator is indeed useful in special cases, and probably should be used with caution. Thanks for the comments!

gsy_framework/sim_results/electric_blue/all_results.py
cruky: Two first messages always have `current_market_slot` as an empty string and slot number 0 and usually no `simulation_raw_data` (That's why `handle_forward_results` had no issue with this, since there was nothing to iterate through). Then the third message is 'T00:00' and slot number is 1 and starts to have data in `simulation_raw_data`. I think that we should filter out these entries with empty string as `current_market_slot` at the very beginning of processing. I will fix it, thanks!

gsy_framework/sim_results/electric_blue/all_results.py
cruky: 65137e486638e2d819a7a2a7cc2d214d89c0ed91

gsy_framework/sim_results/electric_blue/all_results.py
cruky: 98d31ba846e6a411599492c96f834a50cb51bbd2

gsy_framework/sim_results/all_results.py
cruky: So should I change the approach with the additional attribute `self.forward_market_enabled`  for both classes and check `isinsttance` whenever needed?

gsy_framework/sim_results/electric_blue/all_results.py
hannesdiedrich: https://github.com/gridsingularity/gsy-web/pull/1804#discussion_r1000723049

gsy_framework/utils.py
```diff
@@ -238,6 +238,11 @@ def datetime_to_string_incl_seconds(date_time: DateTime) -> str:
     return date_time.format(DATE_TIME_FORMAT_SECONDS)
 
 
+def pendulum_datetime_to_string(datetime_object: DateTime) -> str:
+    """Convert a datetime object into a string with YYYY-MM-DDTHH:mm format."""
+    return datetime_object.format(DATE_TIME_FORMAT)
```
spyrostz, line 243: Minor, however I think that the `format_datetime` function could be used instead of creating a separate function here.

cruky: [68f70c5](https://github.com/gridsingularity/gsy-framework/pull/397/commits/68f70c5b60c290d45c1332c8350e7316b1da0c5e)

hannesdiedrich (approved):
LGTM

hannesdiedrich (approved):
LGTM

PR #398, opened 2022-10-20
feature/GSYE-414: Add calculatation of price profile graph

mrsaemir:
### Please review commit by commit. Thanks

As requested by [GSYE-414](https://gridsingularity.atlassian.net/browse/GSYE-414), we need to pre-calculate price profile in different resolutions for each asset.

Proposed changes:
- Refactor AssetVolumeTimeSeries into two classes (AssetVolumeTimeSeries & AssetTimeSeriesBase)
- Add price and energy_rate calculation to AssetVolumeTimeSeries
- Major refactor for unit tests

Screenshot from 2022-10-18 13-04-07

### DIFF
--- a/gsy_framework/sim_results/electric_blue/aggregate_results.py
+++ b/gsy_framework/sim_results/electric_blue/aggregate_results.py
@@ -130,6 +130,20 @@ class ForwardDeviceStats:  # pylint: disable=too-many-instance-attributes
         assert offer["seller_id"] == self.device_uuid, "Device is not seller of the offer."
         self.open_offers.append(offer)
 
+    @property
+    def average_sell_price(self) -> float:
+        """Return average sell price for the timeslot."""
+        if self.total_energy_sold == 0:
+            return 0
+        return self.total_earned_eur / self.total_energy_sold
+
+    @property
+    def average_buy_price(self) -> float:
+        """Return average buy price for the timeslot."""
+        if self.total_energy_bought == 0:
+            return 0
+        return self.total_spent_eur / self.total_energy_bought
+
 
 def handle_forward_results(
         current_time_slot: DateTime, market_stats: Dict[str, Dict[str, List]]) -> Dict:
--- a/gsy_framework/sim_results/electric_blue/aggregate_results.py
+++ b/gsy_framework/sim_results/electric_blue/aggregate_results.py
@@ -131,14 +131,14 @@ class ForwardDeviceStats:  # pylint: disable=too-many-instance-attributes
         self.open_offers.append(offer)
 
     @property
-    def average_sell_price(self) -> float:
+    def average_sell_rate(self) -> float:
         """Return average sell price for the timeslot."""
         if self.total_energy_sold == 0:
             return 0
         return self.total_earned_eur / self.total_energy_sold
 
     @property
-    def average_buy_price(self) -> float:
+    def average_buy_rate(self) -> float:
         """Return average buy price for the timeslot."""
         if self.total_energy_bought == 0:
             return 0
--- a/gsy_framework/sim_results/electric_blue/aggregate_results.py
+++ b/gsy_framework/sim_results/electric_blue/aggregate_results.py
@@ -53,12 +53,14 @@ class ForwardDeviceStats:  # pylint: disable=too-many-instance-attributes
     total_sell_trade_count: int = 0
     total_energy_sold: float = 0
     total_earned_eur: float = 0
+    accumulated_sell_trade_rates: float = 0
 
     # BUYER
     total_energy_consumed: float = 0
     total_buy_trade_count: int = 0
     total_energy_bought: float = 0
     total_spent_eur: float = 0
+    accumulated_buy_trade_rates: float = 0
 
     def __post_init__(self):
         """Creates non-field attributes for the objects which are needed for further
@@ -94,6 +96,10 @@ class ForwardDeviceStats:  # pylint: disable=too-many-instance-attributes
             total_buy_trade_count=self.total_buy_trade_count + other.total_buy_trade_count,
             total_energy_bought=self.total_energy_bought + other.total_energy_bought,
             total_spent_eur=self.total_spent_eur + other.total_spent_eur,
+            accumulated_buy_trade_rates=(self.accumulated_buy_trade_rates +
+                                         other.accumulated_buy_trade_rates),
+            accumulated_sell_trade_rates=(self.accumulated_sell_trade_rates +
+                                          other.accumulated_sell_trade_rates)
         )
         forward_device_stats.open_bids = open_bids
         forward_device_stats.open_offers = open_offers
@@ -111,10 +117,12 @@ class ForwardDeviceStats:  # pylint: disable=too-many-instance-attributes
             self.total_sell_trade_count += 1
             self.total_energy_sold += trade["energy"]
             self.total_earned_eur += trade["price"]
+            self.accumulated_sell_trade_rates += trade["energy_rate"]
         elif trade["buyer_id"] == self.device_uuid:
             self.total_buy_trade_count += 1
             self.total_energy_bought += trade["energy"]
             self.total_spent_eur += trade["price"]
+            self.accumulated_buy_trade_rates += trade["energy_rate"]
         else:
             raise AssertionError("Device is not seller/buyer of the trade.")
 
@@ -133,16 +141,16 @@ class ForwardDeviceStats:  # pylint: disable=too-many-instance-attributes
     @property
     def average_sell_rate(self) -> float:
         """Return average sell price for the timeslot."""
-        if self.total_energy_sold == 0:
+        if self.total_sell_trade_count == 0:
             return 0
-        return self.total_earned_eur / self.total_energy_sold
+        return self.accumulated_sell_trade_rates / self.total_sell_trade_count
 
     @property
     def average_buy_rate(self) -> float:
         """Return average buy price for the timeslot."""
-        if self.total_energy_bought == 0:
+        if self.total_buy_trade_count == 0:
             return 0
-        return self.total_spent_eur / self.total_energy_bought
+        return self.accumulated_buy_trade_rates / self.total_buy_trade_count
 
 
 def handle_forward_results(
--- a/gsy_framework/sim_results/electric_blue/volume_timeseries.py
+++ b/gsy_framework/sim_results/electric_blue/volume_timeseries.py
@@ -24,24 +24,45 @@ class AssetVolumeTimeSeries(AssetTimeSeriesBase):
     {
         DateTime(2020, 1, 1, 0, 0, 0): {
             DateTime(2020, 1, 1, 0, 0, 0): {
-                "SSP": 1005.4160297565659,
-                "YEAR_FORWARD": {
-                    "bought_kWh": 201.08320595131318,
-                    "sold_kWh": 201.08320595131318,
+                'DAY_FORWARD': {
+                    'bought': {
+                        'energy_kWh': 0.0,
+                        'energy_rate': 0.0,
+                        'price': 0.0},
+                    'sold': {
+                        'energy_kWh': 0.0,
+                        'energy_rate': 0.0,
+                        'price': 0.0}
                 },
-                "MONTH_FORWARD": {
-                    "bought_kWh": 201.08320595131318,
-                    "sold_kWh": 201.08320595131318,
+                'MONTH_FORWARD': {
+                    'bought': {
+                        'energy_kWh': 402.132,
+                        'energy_rate': 0.3,
+                        'price': 120.28},
+                    'sold': {
+                        'energy_kWh': 402.132,
+                        'energy_rate': 0.45,
+                        'price': 181.97}
                 },
-                "WEEK_FORWARD": {
-                    "bought_kWh": 32.432775153437625,
-                    "sold_kWh": 32.432775153437625,
-                },
-                "DAY_FORWARD": {
-                    "bought_kWh": 6.486555030687522,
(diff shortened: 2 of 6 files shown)


gsy_framework/sim_results/electric_blue/aggregate_results.py
```diff
@@ -130,6 +130,20 @@ def add_offer(self, offer: Dict) -> None:
         assert offer["seller_id"] == self.device_uuid, "Device is not seller of the offer."
         self.open_offers.append(offer)
 
+    @property
+    def average_sell_price(self) -> float:
+        """Return average sell price for the timeslot."""
+        if self.total_energy_sold == 0:
+            return 0
+        return self.total_earned_eur / self.total_energy_sold
```
hannesdiedrich, line 138: I have two comments here:
1. this method returns an "energy rate" in EUR/kWh. We decided in the past to at least stick to this nomenclature in the backend
2. I am not sure if this really is the average sell rate here. I would expect the average rate to be the arithmetic mean of the rates at each time slot (calculation 1) not the quotient of summed earned money and summed sold energy (calculation 2)
 ```python
# calculation 1
print((1/2 + 1/3 + 1/4)/3)
#0.3611111111111111

calculation 2
print((1+1+1)/(2+3+4))
# 0.333333333333333
 ```
If we can live with this and only need the average as an approximation, then leave it as is, but if the user needs exact data, we would need to change this. But then the calculation is not that easy implemented I guess.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -100,44 +120,41 @@ def _add_total_energy_sold(
...
+
+        # UPDATING energy_kWh, price and energy_rate
+        attribute_data = volume_time_series[str(time_slot)][product_type.name][attribute_name]
+        attribute_data["energy_kWh"] = round_floats_for_ui(
+            attribute_data["energy_kWh"] + time_slot_info["energy_kWh"])
+        attribute_data["price"] = round_prices_to_cents(
+            attribute_data["price"] + time_slot_info["price"])
+        try:
+            attribute_data["energy_rate"] = round_prices_to_cents(
+                attribute_data["price"] / attribute_data["energy_kWh"])
```
hannesdiedrich, line 148: I guess it is a requirement to return both "price" and "energy_rate", right? AFAIU, either of them can easily be calculated by the FE and we can save the DB space.

tests/test_sim_results/electric_blue/test_volume_timeseries.py
```diff
@@ -76,79 +78,128 @@ def test_adapt_time_slot_for_1_year_resolution(self):
...
+                            AvailableMarketTypes.YEAR_FORWARD, "sold"
                         )
-                    },
-                    "DAY_FORWARD": {"bought_kWh": 0, "sold_kWh": 0}}}}
+                    )
+                assert add_to_volume_time_series_mock.call_count == 3
+                add_to_volume_time_series_mock.assert_has_calls(expected_calls)
+
+    @staticmethod
+    def test__add_to_volume_time_series_works_correctly():
```
hannesdiedrich, line 180: NIT
```suggestion
    def test_add_to_volume_time_series_works_correctly():
```

tests/test_sim_results/electric_blue/test_volume_timeseries.py
```diff
@@ -76,79 +78,128 @@ def test_adapt_time_slot_for_1_year_resolution(self):
...
+                )
+                expected_calls = []
+                for time_slot, energy in profile.items():
+                    expected_calls.append(
+                        call.do_work(
+                            time_slot,
+                            {
+                                "energy_kWh": energy,
+                                "price": energy * asset_stats.average_sell_price},
+                            AvailableMarketTypes.YEAR_FORWARD, "sold"
```
hannesdiedrich, line 173: 👍

hannesdiedrich (commented):
Thanks! Left some comments. Will review again once discussed.

gsy_framework/sim_results/electric_blue/aggregate_results.py
mrsaemir: Thank you for your comments.

2- Actually, I was uncertain which calculation is correct, so I just picked the one that seemed most appropriate to me. I also had a very small chat about this with @spyrostz. Now, I'm ready to switch to the other calculation if you think this one is incorrect. Please let me know.

1- Can you explain more? Should I change any variable name? Do you have any suggestions for me?

Thanks

gsy_framework/sim_results/electric_blue/volume_timeseries.py
mrsaemir: In order to calculate the `energy_rate`, I needed it's compnents (`energy_kWh` and `price`) to be present, otherwise we can't update the `energy_rate` every 15 minutes. And you're correct about saving space, FE can easilly calculate `energy_rate` using price and energy_kWh. The reason I did it in backend was to not letting FE to enter our business logic. What do you think?

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -60,25 +76,23 @@ class AssetVolumeTimeSeries:
...
 
     def save_time_series(self):
         """Save asset volume time series in the DB."""
         # TODO: should be implemented.
-        raise NotImplementedError()
+
+    def _fetch_asset_time_series_from_db(
+            self, year: DateTime) -> Optional[Dict[str, Dict]]:
+        """Fetch already saved asset volume time series."""
+        # TODO: should be implemented.
```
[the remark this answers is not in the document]
mrsaemir: Yes. Actually there should be a seprate PR for this in the future. It should be done after corresponding tables are implemented.

tests/test_sim_results/electric_blue/test_volume_timeseries.py
mrsaemir: I missed that. Thanks a lot.

gsy_framework/sim_results/electric_blue/aggregate_results.py
hannesdiedrich: > - Can you explain more? Should I change any variable name? Do you have any suggestions for me?

What I mean is here to rename the method to: `average_sell_energy_rate` or `average_sell_rate` and also in the variable in the calling code. Thanks

mrsaemir: Oh, sure! Thanks

gsy_framework/sim_results/electric_blue/volume_timeseries.py
hannesdiedrich: > The reason I did it in backend was to not letting FE to enter our business logic

IMO they can to a simple devision without putting too much effort on the FE. Maybe @spyrostz begs to differ?

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
hannesdiedrich: In GSYE-398 I used GSYE-374 as base and started already. This can be done here, too, right?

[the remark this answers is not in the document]
mrsaemir: I have 2 suggestions to tackle this problem:

1- In order for this method to be able to reach DB it needs to have access to Django model classes. As it's not possible to import model classes while gsy-framework is installed in gsy-e, I suggest you to use dependency injection for the models (This way you can avoid using pony or peewee).
Something like this:
```python
    def _fetch_asset_time_series_from_db(
            self, year: DateTime, model_class) -> Optional[Dict[str, Dict]]:
```

2- As I have done some refactoring in this PR, you may want to use this branch as your base to avoid conflicts.

Thanks a lot

gsy_framework/sim_results/electric_blue/aggregate_results.py
mrsaemir: Done.

mrsaemir: Done.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
spyrostz: IMO you should not pass the model at all, but the asset data instead as a dependency (or, even better, as a dependency of the class itself and not of the method). That way the gsy-framework will remain free from DB-specific stuff and will only contain data processing, therefore allowing it to also operate for generating CSV results.
I also agree with Hannes on branching off GSYE-374, no need to wait for it to be merged since the database tables are more or less ready in this PR.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
spyrostz: Well, I am with Amirhossein on this one. I know that it is not a big deal for the UI to do a simple division, however on timeseries data these are going to be *some* divisions. The calculation time of these will essentially be passed on to the user, and judging by the performance of the UI, I think that we should do our best to decrease as much as possible the computations that the UI has to perform.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
mrsaemir: @spyrostz ,
When designing this method, I tried to only fetch asset time series that need to be updated, and that's why you see this method that dynamically decides on what to fetch. And, of course we always have the option of buffering everything. If you think we should stick with the buffering strategy, we can even remove this method.

gsy_framework/sim_results/electric_blue/volume_timeseries.py
hannesdiedrich: OK

gsy_framework/sim_results/electric_blue/aggregate_results.py
hannesdiedrich: Thanks! But how about the calculation of the mean. @spyrostz you know the requirements better. Which "mean" is actually wanted?

spyrostz: The reason that this method is named as `average_sell_price` is in order to match the badly-named UI element (which mentions "Price" while the unit is Euro / MWh). Therefore I agree with Hannes that it should be renamed to reflect the term "rate" and not use the term "price".
For the other topic, I am also not sure about the exact requirements here since I doubt that this chart is so thought through. I tend to agree with Hannes though, that probably what is required here is a timeseries of an average energy rate. In which case we need calculation #1 here.

mrsaemir: No worries then I can swtich to #1. But in order to do that, I need to include `sum_of_energy_rate`s and `number_of_trades` in the time series (and remove `price`). Do we all agree on that?

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
spyrostz: @mrsaemir correct me if I am wrong, but as far as I see the time series data are filtered by asset_uuid and year. I assume the latter part makes things more complex, since it is deduced while updating the timeseries. Even in this case though, there is a way that will allow you to keep DB packages outside of project dependencies, and the framework code clean from any DB models: closures. You can always pass a function / method (defined in gsy-web, which has access to the DB models) with one argument (the year) as an argument in the constructor of this class, which will include the DB connection in its scope. Keep the function in a separate variable, and invoke it from `_fetch_asset_time_series_from_db` . The interface of the class will become more complex, but at least the repo will remain without DB dependencies.

gsy_framework/sim_results/electric_blue/aggregate_results.py
spyrostz: I agree that it would be good to include both sum_of_energy_trades and number_of_trades, they are required in order to calculate this especially in longer duration / resolution options.

spyrostz (approved):
LGTM, looking forward to the integration of the DB!

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
hannesdiedrich: Sorry, I do not see a PR or story for this, can you please point me to this? Thanks a lot!

hannesdiedrich (commented):
Once the DB integration is tracked in a ticket, LGTM

gsy_framework/sim_results/electric_blue/volume_timeseries.py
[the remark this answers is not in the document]
mrsaemir: Actually this will be related to the next phase of the task. After someone created gsy-web tables, he should also adapt this method with the suggestions mentioned by Spyros.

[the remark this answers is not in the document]
mrsaemir: If you think we should at least fix the signature of the function, we can do that, but I have no idea about it for now. Please let me know if you have any suggestions.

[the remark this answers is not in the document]
hannesdiedrich: > Actually this will be related to the next phase of the task.

So, this will include a PR to gsy-web (probably with target GSYE-374) where a new model is added, right? Are you taking care of it in the frame of the ticket? If so, then I will leave my LGTM here.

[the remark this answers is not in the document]
mrsaemir: Exactly. I will help to correctly integrate the new model to this method. Thanks

hannesdiedrich (approved):
LGTM

PR #400, opened 2022-11-02
Feature/gsye 333

spyrostz:


gsy_framework/live_events/b2b.py
```diff
@@ -0,0 +1,84 @@
...
+            self._logger("Cannot deserialize market type parameter (%s).", args)
+            return False
+        return True
+
+    def _validate_start_end_time(self, args: Dict) -> bool:
+        try:
+            if not isinstance(args["start_time"], datetime):
+                str_to_pendulum_datetime(args["start_time"])
+            if not isinstance(args["end_time"], datetime):
+                str_to_pendulum_datetime(args["end_time"])
```
hannesdiedrich, line 66: I think we should enforce pendulum datetime, because the strategy might depend on features of pendulum. Also, I understand that the type of the args["start_time"] will never be a datetime object (neither pendulum nor normal datetime) right? So I propose the following:
```suggestion
            assert isinstance(args["start_time"], str):
            str_to_pendulum_datetime(args["start_time"])
            str_to_pendulum_datetime(args["end_time"])
```

spyrostz: I unfortunately wanted to use this method for both gsy-e (for which, as you said, we use either strings or pendulum datetimes) but also for gsy-web mutations (for which normal datetime objects are used, and this is out of our control, since they are generated by Graphene). In order to avoid casting in the mutation method from python datetimes to pendulum datetimes, I decided to allow both in this validator.

hannesdiedrich: OK, I see, thanks!

gsy_framework/live_events/b2b.py
```diff
@@ -0,0 +1,84 @@
...
+            cls.POST_ORDER_EVENT_NAME, cls.REMOVE_ORDER_EVENT_NAME
+        ]
+
+
+class LiveEventArgsValidator:
+    """Validator class for the forward strategy live event arguments."""
+    def __init__(self, logging_function: Callable):
+        self._logger = logging_function
+
+    def validate_start_trading_event_args(self, args: Dict) -> bool:
```
hannesdiedrich, line 29: I am not sure what our consensus was when we discussed about using `validate` in method names, but I would expect here that the method would raise an exception and not return anything. Do you agree?
If so, I suggest to rename to `are_start_trading_event_args_valid`.

spyrostz: I agree, and I will rename all the methods in this class to reflect this. Thanks!

hannesdiedrich (approved):
LGTM

PR #401, opened 2022-11-09
feature/GSYE-446: Add calculation of asset volume time series

mrsaemir:
As requested by GSYE-446, we want to add calculation of asset volume time series for forward markets.

## Proposed changes:
- Add `ForwardResultsHandler._generate_asset_volume_time_series`
- Add support for INTRADAY market to `ForwardTradeProfileGenerator`
- [x] Add unit tests


gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -25,26 +26,34 @@ def __init__(self):
...
-                market_type = int(market_type_value)
-                self._buffer_bids_offers_trades(area_uuid, market_type, market_stats)
-                current_results = handle_forward_results(current_market_dt, market_stats)
-                self._update_stats_and_time_series(current_results, market_type)
+
+        forward_results = None
+        for area_result in core_stats.values():
+            if "forward_market_stats" in area_result:
+                forward_results = area_result
+                break
```
hannesdiedrich, line 47: NIT: Could be refactored into a separate method

gsy_framework/sim_results/electric_blue/volume_timeseries.py
```diff
@@ -81,24 +82,24 @@ class AssetVolumeTimeSeries(AssetTimeSeriesBase):
...
+        self.get_asset_volume_time_series_db = get_asset_volume_time_series_db
 
     def update_time_series(
             self, asset_stats: ForwardDeviceStats, product_type: AvailableMarketTypes):
         self._add_total_energy_bought(asset_stats, product_type)
         self._add_total_energy_sold(asset_stats, product_type)
 
-    def save_time_series(self):
-        """Save asset volume time series in the DB."""
-        # TODO: should be implemented.
```
hannesdiedrich, line 96: Did you make `asset_time_series_buffer` public in order to remove this method here and implement savinf to the DB on the gsy-web side?

hannesdiedrich (commented):
Will wait for the gsy-web side to leave my approval. LGTM already though. Thanks!

gsy_framework/sim_results/electric_blue/volume_timeseries.py
mrsaemir: You're right. I made it public to access it from gsy-web.

gsy_framework/sim_results/electric_blue/all_results.py
mrsaemir: Nice suggestion. Fixed it.

hannesdiedrich (approved):
LGTM

gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -25,26 +26,29 @@ def __init__(self):
...
-                current_results = handle_forward_results(current_market_dt, market_stats)
-                self._update_stats_and_time_series(current_results, market_type)
+
+        forward_results = self._get_forward_results(core_stats)
+        if not forward_results:
+            return
+
+        current_market_dt = str_to_pendulum_datetime(current_market_slot)
+        for market_type_value, market_stats in forward_results["forward_market_stats"].items():
+            market_type = int(market_type_value)
```
spyrostz, line 49: Minor question, how come this is not an int ? IIRC this method is called in the simulation_results service, therefore it receives serialized dicts and not objects, therefore it will probably not contain the enum but the enum value (therefore an int)

mrsaemir: You're correct, however the problem is that integers can not be used as keys in JSON, so the serializer automatically casts it to a string.

```
In [1]: my_dict = {3: 'GSY'}                                                    

In [2]: import json                                                             

In [3]: json.dumps(my_dict)                                                     
Out[3]: '{"3": "GSY"}'
```

spyrostz: Got it. Thanks for the clarification!

spyrostz (approved):
LGTM

PR #402, opened 2022-11-17
GSYE-445: Replace MarketEnergyBills with SCMBills for SCM simulations

mrsaemir:
As requested by [GSYE-445](https://gridsingularity.atlassian.net/browse/GSYE-445), we want to replace MarketEnergyBills with SCMBills for SCM simulations.

Old SCM bills format (for each area):
old_format

New SCM bills format (for each area):
new_format

### DIFF
--- a/gsy_framework/sim_results/all_results.py
+++ b/gsy_framework/sim_results/all_results.py
@@ -4,19 +4,20 @@ from typing import Dict
 
 from gsy_framework.sim_results.area_throughput_stats import AreaThroughputStats
 from gsy_framework.sim_results.bills import CumulativeBills, MarketEnergyBills
-from gsy_framework.sim_results.cumulative_grid_trades import \
-    CumulativeGridTrades
-from gsy_framework.sim_results.cumulative_net_energy_flow import \
-    CumulativeNetEnergyFlow
+from gsy_framework.sim_results.cumulative_grid_trades import (
+    CumulativeGridTrades)
+from gsy_framework.sim_results.cumulative_net_energy_flow import (
+    CumulativeNetEnergyFlow)
 from gsy_framework.sim_results.device_statistics import DeviceStatistics
 from gsy_framework.sim_results.energy_trade_profile import EnergyTradeProfile
 from gsy_framework.sim_results.kpi import KPI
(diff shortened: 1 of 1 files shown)


mrsaemir:
> Hey, can you please share your findings of the main part of the task:
> 
> > The reason why I have not applied the changes already, is that the DoD of this ticket includes some rudimentary checks in order to validate that the SCM results endpoints report the energy bills in the SCM format (of course when running SCM simulations).
> 
> Thanks a lot!

Very good question! Honestly, I was hoping to get more information on that from your feedbacks. I replaced the class and nothing bad happened after I ran a scm simulation. Also, all the unit tests are passing. So, I have 2 theories:

1- The new class is identical to the old one so everything remained unchanged (which makes no sense to me)
2- We have no unit tests for scm bills to fail (which is highly probable)

I would be extremely happy to know your opinion about it. Thanks

spyrostz:
> > Hey, can you please share your findings of the main part of the task:
> > > The reason why I have not applied the changes already, is that the DoD of this ticket includes some rudimentary checks in order to validate that the SCM results endpoints report the energy bills in the SCM format (of course when running SCM simulations).
> > 
> > 
> > Thanks a lot!
> 
> Very good question! Honestly, I was hoping to get more information on that from your feedbacks. I replaced the class and nothing bad happened after I ran a scm simulation. Also, all the unit tests are passing. So, I have 2 theories:
> 
> 1- The new class is identical to the old one so everything remained unchanged (which makes no sense to me) 2- We have no unit tests for scm bills to fail (which is highly probable)
> 
> I would be extremely happy to know your opinion about it. Thanks

It was my bad that I did not define that this includes testing the UI. Meaning that, you need to check which parts of the SCM UI fail, and to raise the matter to either the frontend channel or in a frontend ticket (let me know and I will create a ticket for you) in order for them to be aware that the structure of the SCM bills has changed (and also to show them the new format of the bills). Unit tests and integration tests should also not fail but you tested all these already so all good on that front!

spyrostz (approved):
LGTM, as discussed in the comment also the FE needs to be informed of any potential breaking changes here.

PR #413, opened 2022-12-20
GSYE-477: Make use of KPICalculationHelper

mrsaemir:


gsy_framework/sim_results/kpi_calculation_helper.py
```diff
@@ -59,4 +59,4 @@ def saving_percentage(
             cls, saving_absolute: float, total_base_energy_cost_excl_revenue: float) -> float:
         """Calculate the percentage of saving."""
         return (abs((saving_absolute / total_base_energy_cost_excl_revenue) * 100)
-                if total_base_energy_cost_excl_revenue else None)
+                if total_base_energy_cost_excl_revenue > FLOATING_POINT_TOLERANCE else None)
```
hannesdiedrich, line 62: Here I don not agree. The quotient `x / FLOATING_POINT_TOLERANCE` is defined. Only `x / 0` is not.

mrsaemir: This is mainly to avoid comparing the number to 0. I expect the FLOATING_POINT_TELERANCE to be so small that makes no difference to zero.

hannesdiedrich: I got that even at first the fist time. With this we are limiting the results to the values > FLOATING_POINT_TOLERANCE. Currently, this values is not THAT small. But ok, leave it as is for now. This is really a nitpicky comment. Leaving my LGTM

hannesdiedrich (approved):
LGTM

gsy_framework/sim_results/kpi_calculation_helper.py
mrsaemir: Thank you. I'd suggest to discuss about this some day in our meetings. I'd like to know more ...

PR #417, opened 2023-01-10
GSYE-488: Included Rebase power profiles to the conversion from power…

spyrostz:
… to energy values.


hannesdiedrich (approved):
Wow, did not remember that we have this profile type (although I implemented it IIRC :-) )
LGTM

PR #418, opened 2023-01-11
GSYE-496: Savings should be calculated based on the energy bill that …

spyrostz:
…does not include the revenue on both the base and the gsy energy bill.


hannesdiedrich (approved):
LGTM

PR #421, opened 2023-01-19
GSYE-452: Add AVRO schema validation

mrsaemir:
As requested by GSYE-452, we want to add validatiion for the simulation results service.

# TODO:
  - [x] Add unittests to test successful compilation of each schema
  - [ ] Add integration tests to check compatibility of the schema with simulation results
  - [x] Add AVRO for `raw_simulation_results` of SCM simulations
  - [x] Add AVRO for `raw_simulations_results` of normal simulations
    - [x] Area
    - [x] PV
    - [x] Load
    - [x] Storage
    - [x] Heatpump
    - [x] InfiniteBus
    - [x] Smart meter
    - [x] Electrolyzer -> Same as storage
    - [x] Wind -> Same as PV
    - [x] Forward markets
    - [x] Settlement markets
    - [x] Future markets
  - [ ] Add AVRO schema for `configuration tree`
  - [ ] Add AVRO schema for `simulation_state`
  - [ ] Add AVRO schema for the `bids_offers_trades`


gsy_framework/schema/avro_schemas/simulation_raw_data_v1.json
```diff
@@ -0,0 +1,562 @@
...
+          }
+        },
+        {
+          "name": "trades",
+          "type": {
+            "name": "Trades",
+            "type": "array",
+            "items": ["Trade"]
+          }
+        }
```
hannesdiedrich, line 396: This is repeated a lot of times. Can't we do some definition and re-use it, like in JSONschema? An example is here:
https://github.com/gridsingularity/gsy-framework/blob/cd361fc1d9d0c920722930e7f117b2b82d020341/gsy_framework/schemas.py#L47

gsy_framework/schema/validators.py
```diff
@@ -0,0 +1,47 @@
...
+        try:
+            validate_avro_schema(self.schema, data, raise_on_error=True)
+            return True, ""
+        except AvroTypeException as exc:
+            if raise_exception:
+                raise SchemaError(message="The provided data is invalid for the schema.") from exc
+            return False, str(exc)
+
+
+def get_schema_validator(schema_name: str) -> BaseSchemaValidator:
```
hannesdiedrich, line 45: I am not sure what do think about this interface WRT of usability. As a user of this function, I need to know which schema file I have to use.
IMO,[ this initiation](https://github.com/gridsingularity/gsy-e/pull/1598/files#diff-334c6d546ca90b8e453ec70f72cff8ead015cd681337d29179dab26338858fcbR48-R49) should be done here in the gsy_framework. This can be re-used on the gsy-web side, too.

gsy_framework/schema/avro_schemas/simulation_raw_data_v1.json
```diff
@@ -0,0 +1,562 @@
+{
```
hannesdiedrich, line 1: I did not understand for which data this schema is for (from the file name). I think that "raw_data" is a too broad expression, but I can also not think of a better name. Nothing to do here, only if you can think of another word for it. Thanks!

gsy_framework/schema/avro_schemas/scm_simulation_raw_data_v1.json
```diff
@@ -0,0 +1,199 @@
+{
```
hannesdiedrich, line 1: Can you please explain why you use the `_v1` suffix? Are we planning to have multiple versions?
IMO, if the data structure changes, we need to change the schema, too.

hannesdiedrich (commented):
Left some questions. Looks good in general. Thanks!

gsy_framework/schema/avro_schemas/simulation_raw_data_v1.json
mrsaemir: This is almost doing the same thing (referencing to a previously defined schema), but I'll try to make it shorter in size (if possible).

gsy_framework/schema/validators.py
mrsaemir: I admit that the initiation can be done in gsy-framework and be used in other services, but then we have to initialize every schema (with all their different versions) here. And this is something I'm trying to avoid as there may be different versions of each schema in the future and obviously initiating all of them in all of the services will use extra amounts of memory. It's important to point out that each service may be tied to an specific version of the schema in the future independently - and this is something to be decided by the service and not gsy-framework.

gsy_framework/schema/avro_schemas/simulation_raw_data_v1.json
mrsaemir: I borrowed this name from [here](https://github.com/gridsingularity/gsy-e/blob/3c5303ace1d0b46b3a7b2aadec4d8b84d8890cf6/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py#L158). And I'm open to any good name you suggest.

gsy_framework/schema/avro_schemas/scm_simulation_raw_data_v1.json
mrsaemir: According to my research about schema management, different services of a system may be using different versions of the same schema. I just added the _v1 suffix to conform to the standards.

gsy_framework/schema/avro_schemas/simulation_raw_data_v1.json
hannesdiedrich: > This is almost doing the same thing (referencing to a previously defined schema) 

I see now. OK, I guess there is no possibility of improving this then. Thanks!

gsy_framework/schema/validators.py
hannesdiedrich: OK, you convinced me here. Leave it as is. Thanks!

gsy_framework/schema/avro_schemas/scm_simulation_raw_data_v1.json
hannesdiedrich: Hmm, I do not foresee us using different schema versions. And IMO it is confusing for the implementor of the validator to have to decide between different versions. There should only be one schema for the raw_data IMO. 
Do you agree?

mrsaemir: OK. I can remove the _v1 suffix for now. We are still able to use them in the future if required.

mrsaemir: Done.

hannesdiedrich (approved):
LGTM

PR #424, opened 2023-01-27
GSYE-492: Added geo_tag_location, asset_count and member_name as addi…

spyrostz:
…tional parameters in the create_member_dict method that creates a new SCM member dict.


gsy_framework/community_datasheet/row_converters.py
```diff
@@ -101,10 +108,8 @@ def create_member_dict(
             "fixed_monthly_fee": fixed_monthly_fee,
             "marketplace_monthly_fee": marketplace_monthly_fee,
             "coefficient_percentage": coefficient_percentage,
-            "geo_tag_location": AssetCoordinatesBuilder().get_member_coordinates({
-                "address": address, "zip_code": zip_code
-            }),
-            "asset_count": 0
```
mrsaemir, line 107: I used to think we are the one that populates the asset_count. or am I wrong?

mrsaemir (approved):
LGTM

gsy_framework/community_datasheet/row_converters.py
spyrostz: Exactly, we are the ones that calculate this, and this was desynced from time to time (essentially when assets were added / removed). That's why with my recent gsy-web changes, we recalculate the asset count on every configuration update and removal.

hannesdiedrich (approved):
LGTM

PR #427, opened 2023-02-16
GSYE-428: Add HeatPump to ScenarioSchemas

mrsaemir:
- [X] Add HeatPump to the scneario schema
- [X] Fix pylint errors
- [X] Add HeatPumpValidator class
- [x] Implement HeatPumpValidator class
- [ ] Add unittests (if needed)

### DIFF
--- a/gsy_framework/constants_limits.py
+++ b/gsy_framework/constants_limits.py
@@ -186,7 +186,7 @@ class ConstSettings:
         """Default values for the heat pump."""
         # range limits
         MAX_POWER_RATING_KW_LIMIT = RangeLimit(0, sys.maxsize)
-       TEMP_C_LIMIT = RangeLimit(-273.15, 6000)
+        TEMP_C_LIMIT = RangeLimit(-273.15, 6000)
         TANK_VOLUME_L_LIMIT = RangeLimit(0, sys.maxsize)
         BUYING_RATE_LIMIT = RateRange(0, 10000)
 
--- a/gsy_framework/constants_limits.py
+++ b/gsy_framework/constants_limits.py
@@ -22,10 +22,9 @@ from datetime import date, datetime
 
 from pendulum import duration, instance
 
-from gsy_framework.enums import (
-    BidOfferMatchAlgoEnum, SpotMarketTypeEnum, CoefficientAlgorithm, AvailableMarketTypes,
-    HeatPumpSourceType
-)
+from gsy_framework.enums import (AvailableMarketTypes, BidOfferMatchAlgoEnum,
+                                 CoefficientAlgorithm, HeatPumpSourceType,
+                                 SpotMarketTypeEnum)
 
 RangeLimit = namedtuple("RangeLimit", ("min", "max"))
 RateRange = namedtuple("RateRange", ("initial", "final"))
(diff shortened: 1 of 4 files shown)


gsy_framework/constants_limits.py
```diff
@@ -185,7 +184,13 @@ class WindSettings:
 
     class HeatPumpSettings:
         """Default values for the heat pump."""
+        # range limits
+        MAX_POWER_RATING_KW_LIMIT = RangeLimit(0, sys.maxsize)
+        TEMP_C_LIMIT = RangeLimit(-sys.maxsize, sys.maxsize)
```
hannesdiedrich, line 189: Let's stay in the boundaries of physics and use the absolute zero and the temperature of the sun as boundaries ;-) 
 ```suggestion
        TEMP_C_LIMIT = RangeLimit(-273.15, 6000)
```

File: gsy_framework/schemas.py [hunk 2]
```diff
@@ -202,8 +208,32 @@ class ScenarioSchemas:
...
+                    "max_temp_C": {"type": "number"},
+                    "initial_temp_C": {"type": "number"},
+                    "external_temp_C": {"anyOf": [{"type": "number"}, {"type": "null"}]},
+                    "external_temp_profile_uuid": {
+                        "anyOf": [{"type": "string"}, {"type": "null"}]},
+                    "tank_volume_l": {"type": "number"},
+                    "consumption_kW": {"anyOf": [{"type": "number"}, {"type": "null"}]},
+                    "consumption_profile_uuid": {"anyOf": [{"type": "string"}, {"type": "null"}]},
+                    "source_type": {
+                        "enum": [HeatPumpSourceType.AIR.value, HeatPumpSourceType.GROUND.value]},
```
hannesdiedrich, line 229: interesting. Did not know that there is an enum field in JSON schema

gsy_framework/validators/heat_pump_validator.py
```diff
@@ -0,0 +1,117 @@
...
+
+        validate_range_limit(
+            initial_buying_rate, preferred_buying_rate, final_buying_rate,
+            {"misconfiguration": [
+                "Requirement "
+                "'initial_buying_rate <= preferred_buying_rate <= final_buying_rate' is not met."
+            ]})
+
+    @classmethod
+    def _check_range(cls, name, value, _min, _max):
```
hannesdiedrich, line 109: Why did you create this wrapper around `validate_range_limit` and why are the input arguments prepended with `_`?

gsy_framework/schemas.py
mrsaemir: Admitedly, JsonSchema is much more powerful than AVRO. It has some mind-blowing features.

hannesdiedrich: Why did we decided for AVRO then? ;-)

gsy_framework/validators/heat_pump_validator.py
mrsaemir: I felt bad about copy-pasting the same code for populating the `error_message` everytime I wanted to validate something. At first, I tried to refactor validate_range_limit but then I saw it's heavily used in many places. So, I created my little wrapper here. Also, I couldn't use `min` and `max` as my arguments as they will shadow the built-in min and max functions of Python. I can rename it to `min_value` and `max_value` if you feel uncomforable.

gsy_framework/schemas.py
mrsaemir: Mostly because of it's binary encoding and better compatibility with Kafka. However we can still debate using of AVRO over JsonSchema.

gsy_framework/validators/heat_pump_validator.py
spyrostz: I would actually like the latter, because all arguments that begin with _ look like unused arguments to my eyes, and seeing them being used below confuses me a bit.

mrsaemir: Done.

spyrostz (approved):
LGTM

PR #428, opened 2023-02-21
GSYE-502: Added new schema for the configuration tree. Added unit tests.

spyrostz:


mrsaemir (commented):
It's surprisingly simple! I used to think that the configuration tree is much more complex. BTW, where are able to run a full simulation without any validation errors?

spyrostz:
> It's surprisingly simple! I used to think that the configuration tree is much more complex. BTW, where are able to run a full simulation without any validation errors?

Yes we were! I tried both SCM and normal simulations locally and they both did not report errors. The only one I did not try was the B2B simulation, however since this is a WIP I decided to leave it for now, despite having taken this into account already into the schema (capacity_kW parameter)

mrsaemir (approved):
LGTM

PR #429, opened 2023-02-23
Bug/gsye 539 Handle case when user inputs 0 values for asset settings

hannesdiedrich:


gsy_framework/community_datasheet/row_converters.py
```diff
@@ -217,9 +220,18 @@ def convert(cls, row: Dict) -> Dict:
         """
         cls._validate_row(row)
 
+        user_min_allowed_soc = row[StorageSheetHeader.MINIMUM_ALLOWED_SOC]
         min_allowed_soc = (
-            row[StorageSheetHeader.MINIMUM_ALLOWED_SOC]
-            or ConstSettings.StorageSettings.MIN_ALLOWED_SOC)
+            user_min_allowed_soc if user_min_allowed_soc not in NULL_VALUES
```
mrsaemir, line 225: I'm tempted to suggest to convert these two lines to a separate method and use it everywhere else. What do you think?

mrsaemir (commented):
LGTM

gsy_framework/community_datasheet/row_converters.py
hannesdiedrich: Done [9adcde3](https://github.com/gridsingularity/gsy-framework/pull/429/commits/9adcde361121cbefe62f11796199558dac3edb85)

mrsaemir (approved):
LGTM

PR #430, opened 2023-02-28
Feature/gsye 501

spyrostz:


gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
```diff
@@ -0,0 +1,341 @@
+{
+  "name": "ScenarioArea",
+  "type": "record",
+  "namespace": "my.types",
```
hannesdiedrich, line 4: Can you please explain this?
From the documentation, I only read `namespace, a JSON string that qualifies the name;` Which I do not understand.

gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
```diff
@@ -0,0 +1,341 @@
...
+  "fields": [
+    {"name": "name", "type": "string"},
+    {"name": "type", "type": ["null", "string"], "default":  "undefined"},
```
hannesdiedrich, line 7: Here the documentatation is also not that clear. Does `"default":  "undefined"` mean that the default value is `None`?

gsy_framework/schema/validators.py
```diff
@@ -49,6 +56,128 @@ def validate(self, data, raise_exception: bool = False) -> (bool, str):
...
+        is_valid, errors = self.validate(data, raise_exception)
+        if not is_valid:
+            raise GSySerializationException(f"Failed to serialize data: {errors}")
+        writer = avro.io.DatumWriter(self.schema)
+        bytes_writer = BytesIO()
+        encoder = avro.io.BinaryEncoder(bytes_writer)
+        writer.write(data, encoder)
+        return bytes_writer.getvalue()
+
+    def deserialize(self, data: bytes) -> Dict:
```
hannesdiedrich, line 83: I do not see these serializing and deserializing methods being used on the gsy-web or gsy-e side. I guess this is for the future, when you plan to exchange pickling end unpickling the payload with these methods, right?

gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
```diff
@@ -0,0 +1,341 @@
...
+    {"name": "target_market_kpi", "type": ["null", "int"]},
+    {"name": "export_capacity_kVA", "type": ["null", "float"]},
+    {"name": "import_capacity_kVA", "type": ["null", "float"]},
+    {"name": "grid_fee_percentage", "type": ["null", "float"]},
+    {"name": "coefficient_percentage", "type": ["null", "float"]},
+    {"name": "marketplace_monthly_fee", "type": ["null", "float"]},
+    {"name": "allow_external_connection", "type": ["null", "boolean"]},
+    {"name": "baseline_peak_energy_export_kWh", "type": ["null", "float"]},
+    {"name": "baseline_peak_energy_import_kWh", "type": ["null", "float"]},
+    {"name": "children", "type": [
```
hannesdiedrich, line 40: I have to admit, that I did not review all lines in detail here (if you want me to, please let me know), but I do not see an entry for the HeatPump. Shall we add it in the frame of https://github.com/gridsingularity/gsy-framework/pull/427 ?

gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
mrsaemir: If I remember correctly, the `default` key, if specefied, populates the dictionary key if it doesn't exists in the original payload. I'm not sure if we really need it as I have always seen AVRO as a tool to only validate our data and not correct our mistakes. What do you think?

gsy_framework/schema/validators.py
```diff
@@ -49,6 +56,128 @@ def validate(self, data, raise_exception: bool = False) -> (bool, str):
             return False, str(exc)
 
 
+class AVROSchemaSerializer(AVROSchemaValidator):
+    """Serializer class that uses AVRO schemas for serializing and deserializing dicts to bytes."""
+    # pylint: disable=no-self-use
+
+    def modify_input(self, data: Dict) -> Dict:
```
mrsaemir, line 63: I find the `modify_input`, `modify_output`, `serialize` and `deserialize` methods very useful. However, I suggest to also add them to BaseSchemaValidator as abstract methods to abstract implementation details from the using code.

gsy_framework/schema/validators.py
```diff
@@ -49,6 +56,128 @@ def validate(self, data, raise_exception: bool = False) -> (bool, str):
...
+            data["slot_length_realtime"] = timedelta(seconds=data["slot_length_realtime"])
+        except (KeyError, ValueError, TypeError) as ex:
+            error_log = (
+                "Time related parameters on the settings were not set or incorrectly assigned.")
+            logger.error(error_log)
+            raise GSySerializationException(error_log) from ex
+        return data
+
+
 def get_schema_validator(schema_name: str) -> BaseSchemaValidator:
```
mrsaemir, line 129: I suggest to update and stick with this function in order to abstract AVRO from the using code.

gsy_framework/schema/validators.py
```diff
@@ -49,6 +56,128 @@ def validate(self, data, raise_exception: bool = False) -> (bool, str):
...
     """Return the appropriate schema validator class based on the schema name."""
     return AVROSchemaValidator(schema_name=schema_name)
+
+
+class SimulationLaunchSerializer:
```
mrsaemir, line 134: Is there any reason you're not inheriting from `BaseSchemaValidator`? Sorry but I'm a big fan of the `Ports and Adapters` architecture :)

tests/schema/test_settings_validator.py
```diff
@@ -0,0 +1,103 @@
...
+        assert not settings["external_connection_enabled"]
+        assert settings["currency"] == 0
+        assert not settings["settlement_market_enabled"]
+        assert settings["relative_std_from_forecast_percent"] == 10.0
+        assert settings["bid_offer_match_algo"] == 1
+        assert settings["scm_coefficient_algorithm"] == 1
+        assert settings["type"] == 0
+
+    def test_simulation_settings_validator_works(self):
+        serializer = AVROSimulationSettingsSerializer()
```
mrsaemir, line 72: I suggest to do

```python
serializer = get_schema_validator(SOME_NAME)
```
here and in the following test methods.

mrsaemir (commented):
LGTM in general, only some comments. Thanks for your PR.

gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
spyrostz: This is a remnant from an older implementation and I will remove it. In general, namespaces are used in order to reference types from different schemas. That way another schema could refer to this one like this: "my.types.InfiniteBus" for instance. I will remove it though since we do not need this functionality yet, and I will add it when needed.

gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
spyrostz: It is exactly as @mrsaemir describes, it populates the default value with None. We unfortunately need this because these classes are not only validators, but serializers too (essentially they intend to also "correct our mistakes" to an extent before serializing and deserializing the dict).
The reason for its existence in the first place is that the "type" key in the area representation is optional in our model. Non-existence means that the area is of type "Area", which is equal to "type": None. I tried to use "Area" as the default value in a former attempt but was unsuccessful. I'll try it again now and maybe it will work out.

gsy_framework/schema/validators.py
spyrostz: Exactly, essentially the class `DataSerializer` needs to be removed and replaced with the serialize / deserialize methods in the long run. I thought that since we already use AVRO for validation, why not use it for serialization too and remove one extra class that we do not need meanwhile.

gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
spyrostz: Darn, I had a discussion with @mrsaemir about this but I unfortunately forgot to add it. Let me quickly do it now, thanks for noticing !

gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
mrsaemir: At the time of writing this comment, I wasn't fully convinced to act as a serializer, too. So, I'm all good with default values now. Thanks

gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
mrsaemir: We also had a discussion about modifying the log handler when running tests and report ERROR logs as test failures. I just remembered it now!

gsy_framework/schema/validators.py
spyrostz: I was also in-between this actually, and I did not do this in order to distinguish pure validators from validators + serializers. On second thought though you are right, I think that we can have only serializer classes that expose the validate method separately, that way the same class can be used to just validate without raising exceptions and also to serialize / deserialize depending on the users' desire. Fixing this

gsy_framework/schema/validators.py
spyrostz: Will do so, thanks

gsy_framework/schema/validators.py
spyrostz: Indeed, fixing

tests/schema/test_settings_validator.py
spyrostz: The main problem with this was that the Settings validator needed a different class than the normal validator. No problem at all though, I fixed it and changed the behavior of get_schema_validator to be similar to a factory method.

gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
hannesdiedrich: Sound good! Thanks!

gsy_framework/schema/avro_schemas/launch_simulation_scenario.json
```diff
@@ -1,10 +1,9 @@
 {
   "name": "ScenarioArea",
   "type": "record",
-  "namespace": "my.types",
   "fields": [
     {"name": "name", "type": "string"},
-    {"name": "type", "type": ["null", "string"], "default":  "undefined"},
+    {"name": "type", "type": ["null", "string"], "default": "Area"},
```
hannesdiedrich, line 6: Ah, it worked. Great!

gsy_framework/schema/validators.py
```diff
@@ -181,3 +197,12 @@ def deserialize(self, data: Dict) -> Dict:
...
         return data
+
+
+def get_schema_validator(schema_name: str) -> BaseSchemaValidator:
+    """Return the appropriate schema validator class based on the schema name."""
+    if schema_name == "launch_simulation_settings":
+        return AVROSimulationSettingsSerializer()
+    if schema_name == "launch_simulation":
+        return SimulationLaunchSerializer()
+    return AVROSchemaSerializer(schema_name=schema_name)
```
hannesdiedrich, line 208: I understand that this was requested in the PR comments, but I do not understand the benefits of this abstraction. This method will gain if blocks when we add serialisers. I personally preferred the version before where specific Serializers where selected in the calling class. Especially the different interfaces of the Serializers make this method not worth it to me. 
Only a minor doubt, no need to convince me, but maybe you have a good reason for this that you could share and help me understand. Thanks a lot!

hannesdiedrich (dismissed):
LGTM

gsy_framework/schema/validators.py
mrsaemir: I take the opportunity to answer this question.
If we imagine the `Ports and Adapters a.k.a. Hexagonal` architecuture as a city, the factory function would be the door and key point of this city. With the help of this function, we can change the whole serialization/validation system in a matter of minutes (or choose the correct serializer dynamically by simply updating this function). As a rule of thump, we never want to see occurances of our custom classes in our code and the code should only know about the interface. A good example of this would be the `django.contrib.auth.get_user_model` which returns the correct user model by checking our settings module.

mrsaemir (approved):
LGTM. Thank you very much.

gsy_framework/schema/validators.py
mrsaemir: I take the opportunity to answer this question.
If we imagine the Ports and Adapters a.k.a. Hexagonal architecuture as a city, the factory function would be the door and key point of this city. With the help of this function, we can change the whole serialization/validation system in a matter of minutes (or choose the correct serializer dynamically by simply updating this function). As a rule of thump, we never want to see occurances of our custom classes in our code and the code should only know about the interface. A good example of this would be the django.contrib.auth.get_user_model which returns the correct user model by checking our settings module.

mrsaemir: I take the opportunity to answer this question.
If we imagine the Ports and Adapters a.k.a. Hexagonal architecuture as a city, the factory function would be the door and key point of this city. With the help of this function, we can change the whole serialization/validation system in a matter of minutes (or choose the correct serializer dynamically by simply updating this function). As a rule of thump, we never want to see occurances of our custom classes in our code and the code should only know about the interface. A good example of this would be the django.contrib.auth.get_user_model which returns the correct user model by checking our settings module.

spyrostz: Just to put my feedback here too, factory classes have the advantage that you have one entry point for all parts of the code. This means that the caller code does not need to select the class that it needs to instantiate, but only to know how to use this method. That way it is far easier to create / use the serializer no matter the implementation or the protocol used as Amirhossein describes. Also it makes it much more easy to test, because only a single method needs to be mocked instead of many different classes. 
Personally what I do not like in this implementation is that the dispatching is done via strings and not by e.g. an enum, because this requires knowledge about the exact string that needs to be sent to the `get_schema_validator` method. Nevertheless I thought to skip it in order to not overcomplicate the PR. If you would like this approach though let me know to adapt it, thanks!

hannesdiedrich (approved):
LGTM

PR #439, opened 2023-04-21
Feature/gsye 564 Add profile validator and resampler

hannesdiedrich:


gsy_framework/validators/profile_validator.py
```diff
@@ -0,0 +1,44 @@
...
+                 slot_length: timedelta):
+        """Validate if profile corresponds to the start_time, end_time and slot_length setting."""
+        assert len(profile) > 0, "profile is empty"
+        ordered_profile = OrderedDict(convert_str_to_pendulum_in_dict(profile))
+        cls._validate_start_end_date(ordered_profile, start_time, end_time, slot_length)
+        cls._validate_slot_length(ordered_profile, slot_length)
+
+    @staticmethod
+    def _validate_start_end_date(
+            profile: Dict[DateTime, float], start_time: DateTime, end_time: DateTime,
```
spyrostz, line 24: This should be OrderedDict, trivial otherwise however it helps understanding the calculation of profile_start_time and profile_end_time

gsy_framework/validators/profile_validator.py
```diff
@@ -0,0 +1,44 @@
...
+            slot_length: timedelta):
+        profile_start_time = next(iter(profile))
+        profile_end_time = next(reversed(profile))
+        assert profile_start_time <= start_time, (
+            f"start_time is not valid {profile_start_time}, should be <= {start_time}")
+        if profile_end_time - profile_start_time == duration(days=1) - slot_length:
```
spyrostz, line 30: This is an exact comparison though, and I do not recall the exact condition when we copy to other days. Do we also check out for exactly one day in that part of the code too ?

gsy_framework/validators/profile_validator.py
```diff
@@ -0,0 +1,44 @@
...
+            # if the profile is only one day long, it will be copied to other days and
+            # consequently, the end_date does not have to be validated
+            return
+        assert profile_end_time >= end_time, (
+            f"end_time is not valid {profile_end_time}, should be >= {end_time}")
+
+    @staticmethod
+    def _validate_slot_length(profile: Dict[DateTime, float], slot_length: timedelta):
+        first_time = next(iter(profile))
```
spyrostz, line 39: I am tempted to ask you to abstract this in a property / method, however leave it as it is since there are only 2 occurrences of it.

spyrostz (dismissed):
~~Some comments, LGTM though ~~

Apologies, missed some commits, will resume it

gsy_framework/utils.py
```diff
@@ -585,3 +590,36 @@ def use_default_if_null(input_value: Union[int, str, float], default_value: Unio
...
     return input_value if input_value not in NULL_VALUES else default_value
+
+
+def _generate_time_slot_list(slot_length: timedelta, sim_duration: timedelta,
+                             start_date: datetime) -> List[datetime]:
+    return [start_date.add(minutes=slot_length.seconds / 60 * time_diff) for time_diff in
+            range(int(sim_duration.total_seconds() / slot_length.total_seconds()))]
+
+
+def resample_hourly_energy_profile(
```
spyrostz, line 601: I think it is a better fit to live with the other profile methods of gsy-framework, therefore IMO read_user_profile is a much better module for the function

gsy_framework/utils.py
```diff
@@ -585,3 +590,36 @@ def use_default_if_null(input_value: Union[int, str, float], default_value: Unio
                         ) -> Union[int, str, float]:
     """Return input_value if not null, else return provided default_value"""
     return input_value if input_value not in NULL_VALUES else default_value
+
+
+def _generate_time_slot_list(slot_length: timedelta, sim_duration: timedelta,
+                             start_date: datetime) -> List[datetime]:
+    return [start_date.add(minutes=slot_length.seconds / 60 * time_diff) for time_diff in
```
spyrostz, line 597: I think that should also be `slot_length.total_seconds()`

spyrostz, line 597: A good optimisation would be for this to return a generator and not a list. That way the returned list would not remain in memory, but only single datetimes.

gsy_framework/utils.py
```diff
@@ -585,3 +590,36 @@ def use_default_if_null(input_value: Union[int, str, float], default_value: Unio
...
+    """Resample hourly energy profile in order to fit to the set slot_length."""
+    slot_length_minutes = slot_length.total_seconds() / 60
+    if slot_length_minutes < 60:
+        assert 60 % slot_length_minutes == 0, "slot_length is not division of 1 hour"
+        scaling_factor = 60 / slot_length_minutes
+        out_dict = {}
+        for time_slot in _generate_time_slot_list(slot_length, sim_duration, start_date):
+            hour_time_slot = datetime(time_slot.year, time_slot.month, time_slot.day,
+                                      time_slot.hour)
+            out_dict[time_slot] = input_profile[hour_time_slot] / scaling_factor
```
spyrostz, line 613: The approach of generating all expected time slots and checking out whether they exist in the profile is fine by me. However, I am not sure that KeyError is the right way to react in case of missing datapoints. I would recommend to either wrap this line around KeyError, or to check whether the hour_time_slot is part of the input_profile. And if it is not, then raise a custom exception (maybe GSyProfileError like on the other comment) that the caller code can then catch.

gsy_framework/validators/profile_validator.py
hannesdiedrich: > I do not recall the exact condition when we copy to other days

we do here, here https://github.com/gridsingularity/gsy-framework/blob/285da741929e5731ebeb3caf3dd52c75a47585b2/gsy_framework/read_user_profile.py#L331

> Do we also check out for exactly one day in that part of the code too 

No, we do not, This would break the gsy-e that still allows gaps in the profiles.

gsy_framework/validators/profile_validator.py
hannesdiedrich: Let me do it actually.

gsy_framework/utils.py
```diff
@@ -585,3 +590,36 @@ def use_default_if_null(input_value: Union[int, str, float], default_value: Unio
...
+            range(int(sim_duration.total_seconds() / slot_length.total_seconds()))]
+
+
+def resample_hourly_energy_profile(
+        input_profile: Dict[datetime, float], slot_length: timedelta, sim_duration: timedelta,
+        start_date: datetime) -> Dict[datetime, float]:
+    """Resample hourly energy profile in order to fit to the set slot_length."""
+    slot_length_minutes = slot_length.total_seconds() / 60
+    if slot_length_minutes < 60:
+        assert 60 % slot_length_minutes == 0, "slot_length is not division of 1 hour"
```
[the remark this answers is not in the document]
hannesdiedrich: I agree. Will raise an exception.

gsy_framework/utils.py
hannesdiedrich: OK, this will prevent us from rebase returning profiles with holes.

gsy_framework/read_user_profile.py
```diff
@@ -329,14 +329,12 @@ def read_arbitrary_profile(profile_type: InputProfileTypes,
             GlobalConfig.IS_CANARY_NETWORK:
         profile = copy_profile_to_multiple_days(profile)
     if profile is not None:
-
-        zero_value_slot_profile = default_profile_dict(current_timestamp=current_timestamp)
-        filled_profile = _fill_gaps_in_profile(profile, zero_value_slot_profile)
-
         if profile_type in [InputProfileTypes.POWER_W, InputProfileTypes.REBASE_W]:
+            zero_value_slot_profile = default_profile_dict(current_timestamp=current_timestamp)
+            filled_profile = _fill_gaps_in_profile(profile, zero_value_slot_profile)
```
spyrostz, line 333: I like this a lot!

spyrostz (commented):
Some comments, mostly not that important, will rereview once dealt with. Thanks!

gsy_framework/validators/profile_validator.py
hannesdiedrich: If the end_time is not the expected end_time, the profile should be rejected.
Either it is exactly one day (minus the slot_length) or it is as long as the simulation duration.
If you like I could make it a bit less strict and also allow profiles that are longer than one day but shorter than the simulation duration. However, this will mess with the `copy_profile_to_multiple_days` here we only look at the hours and minutes and the latest available value that corresponds to the hour:minute combination will be used. So, if you just add one more time_slot to the profile (for the next days' 00:00), it will set the 00:00 to the next days, but not for the first day. IMO this is not ideal and I rather be strict here.
What do you think?

spyrostz: Convinced, better to remain strict here indeed. Thanks for the detailed explanation!

gsy_framework/read_user_profile.py
```diff
@@ -385,3 +389,47 @@ def read_and_convert_identity_profile_to_float(profile, current_timestamp: DateT
 
 def _convert_identity_profile_to_float(profile, _current_timestamp: DateTime = None):
     return {k: float(v) for k, v in profile.items()}
+
+
+def _generate_time_slots(
+        slot_length: timedelta, sim_duration: timedelta, start_date: datetime
+) -> Generator:
+    return (start_date + timedelta(seconds=slot_length.total_seconds() * time_diff_count)
```
spyrostz, line 397: You need to `yield` instead of return here I think

gsy_framework/read_user_profile.py
```diff
@@ -385,3 +389,47 @@ def read_and_convert_identity_profile_to_float(profile, current_timestamp: DateT
...
+        return profile[key]
+    except KeyError as ex:
+        raise GSyProfileException(f"The input profile does not contain the key {key}") from ex
+
+
+def resample_hourly_energy_profile(
+        input_profile: Dict[datetime, float], slot_length: timedelta, sim_duration: timedelta,
+        start_date: datetime) -> Dict[datetime, float]:
+    """Resample hourly energy profile in order to fit to the set slot_length."""
+    slot_length_minutes = slot_length.total_seconds() / 60
```
spyrostz, line 413: Minor, but I think that `slot_length.total_minutes()` would be preferable.

gsy_framework/read_user_profile.py
```diff
@@ -385,3 +389,47 @@ def read_and_convert_identity_profile_to_float(profile, current_timestamp: DateT
...
+    if slot_length_minutes < 60:
+        if 60 % slot_length_minutes != 0:
+            raise GSyProfileException("slot_length is not division of 1 hour")
```
spyrostz, line 416: I think that it is more explicit if you use the term "exact division" in order to denote that only divisors of 60 are allowed as slot length minute values.

spyrostz (approved):
Minor comments, LGTM already

gsy_framework/read_user_profile.py
hannesdiedrich: I would like to also support datetime.datetime objects that do not have the `total_minutes` method (only pendulum.datetime has it). Does it make sense?

gsy_framework/read_user_profile.py
hannesdiedrich: Thanks a lot! My limited english was the reason for this formulation. Will change it like you propose

gsy_framework/read_user_profile.py
```diff
@@ -332,13 +332,14 @@ def read_arbitrary_profile(profile_type: InputProfileTypes,
             GlobalConfig.IS_CANARY_NETWORK:
         profile = _copy_profile_to_multiple_days(profile)
     if profile is not None:
-        if profile_type in [InputProfileTypes.POWER_W, InputProfileTypes.REBASE_W]:
+        if profile_type is not InputProfileTypes.ENERGY_KWH:
```
spyrostz, line 335:
```suggestion
        if profile_type not in [InputProfileTypes.ENERGY_KWH, InputProfileTypes.IDENTITY]:
```
Unless I misunderstood here and you would prefer to fill identity profiles too. In this case though, why not include the energy profiles from `_fill_gaps_in_profile` ?

gsy_framework/read_user_profile.py
spyrostz: Clear, ignore the comment then, thanks!

gsy_framework/read_user_profile.py
hannesdiedrich: > Unless I misunderstood here and you would prefer to fill identity profiles too

Yes, this was my goal in order to be backwards compatible

> In this case though, why not include the energy profiles to _fill_gaps_in_profile

We discussed this and said that this is implicit behaviour and should not be done really. For the "new" energy profiles I thought we should be correct from the start.
Or do you think that we validate the gsy-web profiles anyway and there will not be any gaps from that side anyway?

gsy_framework/validators/profile_validator.py
```diff
@@ -50,3 +54,12 @@ def _validate_slot_length(self):
                 f"{self.slot_length.total_seconds() / 60} minutes "
                 f"(previous time_slot: {first_time})")
             first_time = time_slot
+
+    def _get_and_validate_time_diffs(self) -> timedelta:
+        first_time = self._profile_start_time
+        time_diffs = set()
+        for time_slot in list(self.profile.keys())[1:]:
+            time_diffs.add((time_slot-first_time).total_seconds())
+            first_time = time_slot
```
spyrostz, line 63: Equally performant, but a set comprehension is more concise IMO:

```suggestion
    timestamps = list(self.profile.keys())
    time_diffs = {end - start for start, end in zip(timestamps[:, -1], timestamps[1:])}
```
The map function would equally do the trick, however I think that this is less memory-intensive.

gsy_framework/read_user_profile.py
spyrostz: Ok , now I recall our discussion. The problem I have with this is that we treat energy and identity profiles differently, which might make debugging more difficult in the long run (one would have to remember that gap-filling only applies to identity and not energy profiles). I agree with you that the only problem are profiles that already exist on the database, for all new profiles we should not have this problem as far as I recall. Anyways, because this discussion is probably an overkill, it is fine to stay as is, we will probably anyways remove this from identity profiles later down the road.

spyrostz (commented):
One comment, LGTM once resolved!

spyrostz (approved):
LGTM

PR #441, opened 2023-04-27
Reverted the former fix where the removal of the = clause led to not …

spyrostz:
…being allowed strategies with the same initial and final value.


hannesdiedrich (approved):
I remember now, that I changed this already some months/years back and we also reverted it at some point :-D

PR #450, opened 2023-06-14
GSYE-560: Make AggregatorChannels.batch_commands independent on the a…

hannesdiedrich:
…ggregator uuid, because it is passed inside the payload anyway.


spyrostz (approved):
LGTM

PR #476, opened 2023-09-26
FE-443: Add validation of HP source type

hannesdiedrich:


gsy_framework/validators/heat_pump_validator.py
```diff
@@ -92,6 +90,23 @@ def validate_rate(cls, **kwargs):
...
             ]})
 
+    @staticmethod
+    def _validate_source_type(**kwargs):
+        if kwargs.get("source_type"):
+            allowed_source_type_ints = [st.value for st in HeatPumpSourceType]
+            if kwargs.get("source_type") not in allowed_source_type_ints:
+                raise GSyDeviceException(
+                    {"misconfiguration": [
+                        f"HeatPump source type not one of {allowed_source_type_ints}"]})
```
spyrostz, line 100: There is a more convenient way, showcased here https://github.com/gridsingularity/gsy-web/blob/master/simulation/graphql/configuration_objects.py#L137:

```suggestion
        try:
            if kwargs.get("source_type"):
                HeatPumpSourceType(kwargs.get("source_type"))
            except ValueError:
                raise GSyDeviceException(
                    {"misconfiguration": [
                        f"Incorrect heatpump source type."]})
```

hannesdiedrich: Ah yes, good idea. Are you also proposing to move this check into the gsy-web?
IMO it is good to have in the framework, right?

spyrostz: It is fine to stay here IMO, no need to move it to gsy-web since it can be reusable.

spyrostz (approved):
LGTM

spyrostz (approved):
LGTM

PR #478, opened 2023-10-19
GSYE-563: Remove IS_CANARY_NETWORK from GlobalConfig and remove colla…

hannesdiedrich:
…boration_uuid from the simulation_launch schema

### DIFF
--- a/gsy_framework/utils.py
+++ b/gsy_framework/utils.py
@@ -98,9 +98,9 @@ def generate_market_slot_list(start_timestamp=None):
     No input arguments, required input is only handled by a preconfigured GlobalConfig
     @return: List with market slot datetimes
     """
-    time_span = duration(days=PROFILE_EXPANSION_DAYS) \
-        if GlobalConfig.CONFIG_TYPE \
-        else min(GlobalConfig.sim_duration, duration(days=PROFILE_EXPANSION_DAYS))
+    time_span = (duration(days=PROFILE_EXPANSION_DAYS)
+                 if GlobalConfig.is_canary_network()
+                 else min(GlobalConfig.sim_duration, duration(days=PROFILE_EXPANSION_DAYS)))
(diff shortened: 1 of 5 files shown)


gsy_framework/utils.py
```diff
@@ -596,3 +597,8 @@ def use_default_if_null(input_value: Union[int, str, float], default_value: Unio
                         ) -> Union[int, str, float]:
     """Return input_value if not null, else return provided default_value"""
     return input_value if input_value not in NULL_VALUES else default_value
+
+
+def is_canary_network():
+    """Return if the GlobalConfig is set up for a Canary Network"""
+    return GlobalConfig.CONFIG_TYPE == ConfigurationType.CANARY_NETWORK.value
```
spyrostz, line 604: How about moving this method in the `GlobalConfig` class ? That way it will be grouped together with the configuration and will not be part of the generic `utils` module

gsy_framework/utils.py
```diff
@@ -99,7 +100,7 @@ def generate_market_slot_list(start_timestamp=None):
     @return: List with market slot datetimes
     """
     time_span = duration(days=PROFILE_EXPANSION_DAYS) \
-        if GlobalConfig.IS_CANARY_NETWORK \
+        if GlobalConfig.CONFIG_TYPE \
```
spyrostz, line 103: I think that collaborations should not be included here, because they also have a valid `sim_duration`.

```suggestion
        if is_canary_network() \
```

spyrostz (commented):
Some comments, will have another look once fixed, thanks!

gsy_framework/utils.py
hannesdiedrich: This is a bug, will fix it. Thanks

gsy_framework/utils.py
hannesdiedrich: Good idea, let me add this.

spyrostz (approved):
LGTM

PR #480, opened 2023-10-24
GSYE-651: Adapt heat pump validator to also allow parcial provision o…

hannesdiedrich:
…f price parameters and validate if update_interval is not set to zero

### DIFF
--- a/gsy_framework/validators/heat_pump_validator.py
+++ b/gsy_framework/validators/heat_pump_validator.py
@@ -67,17 +67,17 @@ class HeatPumpValidator(BaseValidator):
     @classmethod
     def validate_rate(cls, **kwargs):
         """Validate energy rate related arguments."""
-        if not(kwargs.get("initial_buying_rate")
-               and kwargs.get("final_buying_rate")
-               and kwargs.get("update_interval")):
-            return
-        if (not kwargs.get("initial_buying_rate")
-                or not kwargs.get("final_buying_rate")
-                or not kwargs.get("update_interval")):
+        if (kwargs.get("initial_buying_rate") is None
+                or kwargs.get("final_buying_rate") is None
+                or kwargs.get("update_interval") is None):
             raise GSyDeviceException(
                 {"misconfiguration": [
                     "All pricing parameters of heat pump should be provided:"
                     "initial_buying_rate, final_buying_rate, update_interval"]})
+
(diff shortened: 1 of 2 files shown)


gsy_framework/validators/heat_pump_validator.py
```diff
@@ -67,18 +67,22 @@ def _validate_temp(cls, **kwargs):
     @classmethod
     def validate_rate(cls, **kwargs):
         """Validate energy rate related arguments."""
-        if not(kwargs.get("initial_buying_rate")
-               and kwargs.get("final_buying_rate")
-               and kwargs.get("update_interval")):
+        if (kwargs.get("initial_buying_rate") is None
+                and kwargs.get("final_buying_rate") is None
+                and kwargs.get("update_interval") is None):
```
spyrostz, line 72: I like the fix, however I think that this is `or`, not `and`. The original clause says: if not all 3 are not-None. Therefore the converted clause should say: if any of the 3 is None

spyrostz: After looking at the following clause though, I think that this clause is redundant, because we raise an exception if any of the 3 is None. I would recommend to remove this completely actually.

spyrostz (commented):

gsy_framework/validators/heat_pump_validator.py
```diff
@@ -67,18 +67,18 @@ def _validate_temp(cls, **kwargs):
     @classmethod
     def validate_rate(cls, **kwargs):
         """Validate energy rate related arguments."""
-        if not(kwargs.get("initial_buying_rate")
-               and kwargs.get("final_buying_rate")
-               and kwargs.get("update_interval")):
```
hannesdiedrich, line 72: @spyrostz Sorry to drag you into this one more time.
We need this clause in order to allow the user to not provide pricing at all and let the strategy use the default values.
For this case, we do not need to validate.

hannesdiedrich: I rea-added it in the last commit

spyrostz: Understood and thank you for correcting the clause! It now is correct and reflects the condition of all being None, contrary to the former implementation which evaluated if one of them was None.

spyrostz (approved):
LGTM

PR #486, opened 2023-11-23
Feature/gsye 650 fix

spyrostz:


gsy_framework/sim_results/kpi.py
```diff
@@ -140,22 +147,24 @@ def _dissipate_infinite_consumption(self, trade: Dict):
...
+                    self._accumulate_infinite_consumption(trade)
+                    self._dissipate_infinite_consumption(trade)
 
     def update_area_kpi(self, area_dict: Dict, core_stats: Dict):
         """Update kpi after every market cycle"""
         self.total_energy_demanded_wh = 0
+
         self._accumulate_total_energy_demanded(area_dict, core_stats)
-        self._accumulate_energy_trace(core_stats)
+        self._accumulate_energy_trace(core_stats, "P2P" in area_dict["name"])
```
hannesdiedrich, line 167: Do we really want to add this solution permanently?
If yes, we should add a more unique string IMO.
e.g. `##P2P##` What do you think?

gsy_framework/sim_results/kpi.py
```diff
@@ -49,11 +49,13 @@ def __init__(self):
 
     def accumulate_devices(self, area_dict: Dict):
         """Accumulate all child devices with respective types"""
+        if "DH" in area_dict["name"]:
+            return
         for child in area_dict["children"]:
-            if is_producer_node_type(child):
+            if is_producer_node_type(child) or is_heatpump_node_type(child):
```
hannesdiedrich, line 55: why is the heat pump also a producer?

hannesdiedrich (commented):
Two comments, LGTM once addressed

gsy_framework/sim_results/kpi.py
spyrostz: Should not be and good catch, thanks!

gsy_framework/sim_results/kpi.py
spyrostz: Let me think of it a bit more and adapt it. Same goes for the "DH" string for distinguishing district heating.

hannesdiedrich:
@spyrostz We should also discuss whether the distinguishing between P2P and DH should end up in our codebase (whether we should push this)

PR #499, opened 2024-01-22
Reverted the clause that controls whether the profile is copied in mu…

spyrostz:
…ltiple days, and specified it even more in order to apply only in case the profile is exactly a daily one. If not, the profile should not be used in order to be copied to more days.

Root cause of the issue was this method `_copy_profile_to_multiple_days`. If this method is called with a daily profile, then the first/single day is used in order to copy to following days. If the method is called with a weekly profile though, it will use the last day of the profile, and overwrite the existing profile in order to all be copies of the last day of the profile. 
As a hotfix, corrected the clause in order for this method to get called only if the input is a daily profile and not a multi-day profile.


hannesdiedrich (approved):
LGTM

PR #500, opened 2024-01-22
Added one extra market slot in the read_arbitrary_profile.profile_dur…

spyrostz:
…ation calculation, in order to take into account the last market slot in the duration


hannesdiedrich (approved):
LGTM

PR #503, opened 2024-02-01
Feature/gsye 682

spyrostz:


gsy_framework/enums.py
```diff
@@ -25,6 +25,7 @@ class CoefficientAlgorithm(Enum):
     STATIC = 1
     DYNAMIC = 2
     NO_COMMUNITY_SELF_CONSUMPTION = 3
+    LIVE_STREAM = 4
```
hannesdiedrich, line 28: This will be implemented in the frame of GSYE-672, right?

hannesdiedrich (approved):
LGTM

gsy_framework/enums.py
spyrostz: Exactly. This is just a placeholder for the following ticket

PR #520, opened 2024-05-16
GSYE-690: Remove cloud_coverage

BigTava:


BigTava:
Can't commit if coverage not >= than previous? This awesome ! But now what do I do? 🤣

tests/schema/test_scenario_validator.py
```diff
@@ -119,7 +118,6 @@ def setup_method(self):
                       "panel_count": 1,
                       "fit_to_limit": True,
                       "power_profile": "{\"filename\": \"Berlin_pv_1m.csv\"}",
-                      "cloud_coverage": 4,
```
hannesdiedrich, line 122: We like to keep the cloud coverage for the PV asset. Only for the general settings we like to remove it. So, this is not correct.

hannesdiedrich (changes requested):
I think there is a misunderstanding here. We only wanted to remove the cloud coverage  in the GlobalSettings, not for the individual PVs (for now).

PR #525, opened 2024-06-04
Corrected AVRO schema for LoadHours strategy. Fixed simulation result…

spyrostz:
…s for Electric Blue to use avg_power_W and not the capacity_kW argument for load strategies.

### DIFF
--- a/gsy_framework/sim_results/electric_blue/all_results.py
+++ b/gsy_framework/sim_results/electric_blue/all_results.py
@@ -147,9 +147,11 @@ class ForwardResultsHandler:  # pylint: disable=too-many-instance-attributes
 
         asset_uuid = current_asset_stats.device_uuid
         if asset_uuid not in self.asset_volume_time_series:
+            # We assume that for EB, slot length is 1 hour, hence no need for conversion between
+            # power and energy
             asset_peak_kwh = (
                 asset_info["capacity_kW"] if "capacity_kW" in asset_info
-                else asset_info["avg_power_W"])
+                else asset_info["avg_power_W"] * 1000.0)
             self.asset_volume_time_series[asset_uuid] = {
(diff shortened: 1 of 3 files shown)


gsy_framework/sim_results/electric_blue/all_results.py
```diff
@@ -147,10 +147,13 @@ def _generate_asset_volume_time_series(
 
         asset_uuid = current_asset_stats.device_uuid
         if asset_uuid not in self.asset_volume_time_series:
+            asset_peak_kwh = (
+                asset_info["capacity_kW"] if "capacity_kW" in asset_info
+                else asset_info["avg_power_W"])
```
hannesdiedrich, line 152: I do not remember this part of the code, but the units look wrong. Here W or kW values are assigned to a kWh value?

hannesdiedrich (commented):
One question. once resolved, LGTM

gsy_framework/sim_results/electric_blue/all_results.py
spyrostz: There is a reason why you do not remember this part of the code, it is for Electric Blue :) I added a comment to signify that slot length is hourly in this case, and converted the units from W to kWh. Thanks for the remark.

hannesdiedrich: Thanks! And the variable is ok to be named "asset_peak_kwh" because this is a hourly resolution?

hannesdiedrich (approved):
LGTM

gsy_framework/sim_results/electric_blue/all_results.py
spyrostz: Yes, I left it as an extra comment even, in order to denote that no power -> energy conversion is needed here.

PR #536, opened 2024-09-09
Feature/gsye 723

hannesdiedrich:


spyrostz (dismissed):
LGTM, however I think that the saved state functionality will need special handling. I expect this to be in gsy-web though so will mention it in this repo if I see an issue. Thanks!

spyrostz (approved):
LGTM

PR #537, opened 2024-09-09
feature/GSYE-732: remove market_maker_rate, coefficient_percentage and feed_in_tariff from area represetation

BigTava:


hannesdiedrich (approved):
LGTM in general. But here also, why is the first commit part of this PR? Why not opening a separate one for better understanding?

spyrostz (approved):
LGTM

PR #542, opened 2024-10-16
GSYE-780: Added contracted power fee, contracted power cargo fee and …

spyrostz:
…energy cargo fee to the available fees.


hannesdiedrich (approved):
LGTM

PR #553, opened 2024-11-08
Feature/gsye 807

hannesdiedrich:


gsy_framework/community_datasheet/row_converters.py
```diff
@@ -23,10 +23,17 @@
...
-from gsy_framework.utils import use_default_if_null
+from gsy_framework.utils import use_default_if_null, convert_string_to_boolean
+from gsy_framework.enums import CoefficientAlgorithm
 
 logger = logging.getLogger(__name__)
 
+# this is in order to expose better naming to the user
+ADVANCED_SETTINGS_FIELD_MAPPING = {
+    "intracommunity_rate": "intracommunity_rate_base_eur",
+    "operational_hours_of_delay": "scm_cn_hours_of_delay",
```
BigTava, line 34: Operational hours of delay is a so much self-evident variable than scm_cn_hours_of_delay 😅. The intracommunity_rate is only valid for communities with fixed price right? and just need to calculate the energy bill from that. It should be optional since that is not the case for the communities we have now.

hannesdiedrich: > Operational hours of delay is a so much self-evident variable than scm_cn_hours_of_delay 😅.

yeah. this is why this mapping exists. How about adding a ticket for renaming `scm_cn_hours_of_delay` to `operational_hours_of_delay` ?

> The intracommunity_rate is only valid for communities with fixed price right? and just need to calculate the energy bill from that. 

Not sure what you mean. The `intracommunity_rate_base_eur` is already used on the gsy-e side. So we need to somehow pass it there. I would not distinguish between optional and required parameters here, as they are all in a separate Excel Tab which we could communicate to the user to only touch with our supervision. What do you think?

BigTava: Telling javier and jose that they can only edit the Advanced Tab with our supervision is IMO a big turn off. If the intracommunity_rate_base_eur is a property that should be set if the community managers want a fixed price for the community then it should be optional, since that may not be the case.

BigTava: changing scm_cn_hours_of_delay  to operational_hours_of_delay is a good move, and we can create a ticket for that and next time we change something in that table we pitch in that ticket too.

hannesdiedrich: > Telling javier and jose that they can only edit the Advanced Tab with our supervision is IMO a big turn off

I am fine with showing all the options to him and pre-populate them already in the default CDS.
If you think that this is too overwhelming, then we need to review the list of advanced settings to be exposed by default here.

Let me show how the Advanced Settings tab is currently looking:
image

hannesdiedrich: For now we decided to keep these settings in the CDS always visible to the user.

gsy_framework/community_datasheet/row_converters.py
```diff
@@ -76,7 +83,48 @@ def convert(cls, row: Dict) -> Dict:
...
+
+        The field name is converted to snake case.
+        """
+        cls._validate_row(row)
+
+        field_name, value = row[0], row[2]  # Each row is a tuple (setting-name, legend, value)
+        field_name = field_name.strip().lower().replace(" ", "_")
+        field_name = (
+            ADVANCED_SETTINGS_FIELD_MAPPING[field_name]
+            if field_name in ADVANCED_SETTINGS_FIELD_MAPPING
```
spyrostz, line 111: I would actually expand this logic and be more explicit actually. I would even include all the CDS string fields as-is, like this:
```
{"Coefficient Algorithm": "coefficient_algorithm", "Grid Fees Reduction": "grid_fees_reduction"}
```
not just for less mental hassle, but also for being able to debug in a more sane way (at the moment if you add invalid fields, to my recollection it is not immediately clear which field is incorrectly named) and also to be searchable (e.g. if I globally search for "Coefficient Algorithm" I want to find where is this CDS header defined)

gsy_framework/community_datasheet/row_converters.py
```diff
@@ -76,7 +83,48 @@ def convert(cls, row: Dict) -> Dict:
...
+            else field_name
+        )
+        if field_name == "coefficient_algorithm":
+            return {field_name: cls._get_coefficient_algorithm_enum(value).value}
+        if field_name.startswith("enable_"):
+            field_name = field_name.replace("enable_", "")
```
spyrostz, line 117: If you follow my suggestion, this string manipulation will also be removed.

spyrostz (commented):

gsy_framework/community_datasheet/row_converters.py
hannesdiedrich: OK, i get your point.
For this I need to omit the `field_name.strip().lower()` and extend the mentioned mapping (only writing it down here in order to sync)

PR #556, opened 2024-11-27
GSYE-719: build carbon footprint handler to be pluged in gsy-web rese…

BigTava:
…arch-scripts and gsy-e


gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,169 @@
...
+                    [target_timestamp], method="nearest"
+                )[0]
+                nearest_row = df_total_carbon_and_energy.iloc[nearest_index]
+
+                total_carbon_g = nearest_row["Total Carbon (gCO2eq)"]
+                total_energy_kWh = nearest_row["Total Energy (kWh)"]
+
+                imported_energy_from_grid_kWh = area_result["bought_from_grid"]
+                imported_energy_from_community_kWh = area_result[
+                    "energy_bought_from_community_kWh"
```
hannesdiedrich, line 137: With this, the carbon foodprint is only applicable for SCM. Don't we want to use also the `ImportedExportedEnergyHandler` or instead?

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,169 @@
...
+            country_code, start_date, end_date
+        )
+        for result in results_list:
+            current_market = result["current_market"]
+
+            for area_uuid in result["results"].keys():
+                area_result = result["results"][area_uuid]
+                target_timestamp = pd.Timestamp(current_market).tz_localize("UTC")
+
+                nearest_index = df_total_carbon_and_energy.index.get_indexer(
```
hannesdiedrich, line 127: The term "nearest" confused me at first (i thought you are dealing with locations here). With near you mean: "the closes time stamp to the `target_timestamp`, right?". Maybe you can add an inline comment for clarification.

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,169 @@
...
+                imported_energy_from_community_kWh = area_result[
+                    "energy_bought_from_community_kWh"
+                ]
+
+                carbon_generated_base_case = (
+                    total_carbon_g
+                    * (imported_energy_from_grid_kWh + imported_energy_from_community_kWh)
+                    / total_energy_kWh
+                )
+                carbon_generated_with_trading = (
```
hannesdiedrich, line 145: Minor, but we need to discuss the nomenclature here (talking about the suffix `_trading`). IMO we should be consistent use the suffix `_gsy` like in the savings calculation. Or, alternatively, `_p2p`, but this would also not be in line with the SCM.

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,169 @@
...
+from entsoe import EntsoePandasClient
+from entsoe.mappings import lookup_area, Area
+from entsoe.parsers import parse_generation
+
+from gsy_framework.sim_results.results_abc import ResultsBaseClass
+
+URL = "https://web-api.tp.entsoe.eu/api"
+
+GENERATION_PLANT_TO_CARBON_EMISSIONS = {
+    # source: https://www.ipcc.ch/site/assets/uploads/2018/02/ipcc_wg3_ar5_annex-iii.pdf#page=7
```
hannesdiedrich, line 15: Can you please explain why you hardcoded this here and not call the API? Is this because of your comment in line 27?

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,169 @@
...
+import requests
+import pandas as pd
+from pendulum import duration
+from entsoe import EntsoePandasClient
+from entsoe.mappings import lookup_area, Area
+from entsoe.parsers import parse_generation
+
+from gsy_framework.sim_results.results_abc import ResultsBaseClass
+
+URL = "https://web-api.tp.entsoe.eu/api"
```
hannesdiedrich, line 12:
```suggestion
ENTSOE_URL = "https://web-api.tp.entsoe.eu/api"
```

tests/test_sim_results/constants.py
hannesdiedrich, line 1: Not really a fan of the naming of this module. If I would look for test data, I would not look into "constants.py". But leave it as is. Global search will help me.

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,169 @@
...
+        df_numeric["Total Energy (kWh)"] = df_numeric.sum(axis=1) * 1000  # convert MWh to kWh
+        df_numeric["Total Carbon (gCO2eq)"] = df_numeric.mul(emissions_map, axis=1).sum(axis=1)
+
+        generation_types = df.iloc[0]
+        emissions_map = generation_types.map(
+            lambda x: GENERATION_PLANT_TO_CARBON_EMISSIONS.get(x, {}).get(stat, 0)
+        )
+        return df_numeric[["Total Energy (kWh)", "Total Carbon (gCO2eq)"]]
+
+    def update(self, country_code: str, time_slot: duration, results_list: List[Dict]):
```
hannesdiedrich, line 112: With putting the country_code in the interface of the method, you are assuming that all areas in the simulation are in the same country. Works for now, but how do you get the country code ? Curious about your gsy-web and gsy-e PRs.

hannesdiedrich (commented):
Left some comments and questions, most of them minor. Will review again once adapted.

gsy_framework/sim_results/carbon_emissions.py
BigTava: Yes, how do I adapt to use what you are saying. Let's catch up after standup.

gsy_framework/sim_results/carbon_emissions.py
BigTava: Thanks!

gsy_framework/sim_results/carbon_emissions.py
BigTava: Using _gsy. Though in gsy we also perform non-p2p/non-trading scenarios

gsy_framework/sim_results/carbon_emissions.py
BigTava: Yes, it follows the adaption for the code in https://github.com/EnergieID/entsoe-py. Once we update to python 3.11 all of these extra methods will be removed.

gsy_framework/sim_results/carbon_emissions.py
BigTava: Thanks

tests/test_sim_results/constants.py
BigTava: When we call pytest it looks for all files that start with "test_". When I was looking for a place to declare some constant I would never think that the best place would be in test_data because it means you are "testing something related to data". So I renamed to constants.py, but I don't mind reverting.

gsy_framework/sim_results/carbon_emissions.py
BigTava: I thought it would be best to accept the country code rather than geo coordinates, because it would simplify when we use it in research-scripts for example, but now that I think it further maybe it is best to calculate the country code here

tests/test_sim_results/constants.py
hannesdiedrich: I understand, ok. How about "fixtures.py"? But really, leave it if fixtures is not fine for you.

gsy_framework/sim_results/carbon_emissions.py
hannesdiedrich: Yes, we need the location of the area in any case and need to pass it here:  `ResultsCalculation._calculate_current_market_sim_results`

gsy_framework/sim_results/carbon_emissions.py
spyrostz: This is indeed not needed BTW if you use the ImportedExportedEnergyHandler

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def calculate_carbon_ratio(
+        self, country_code: str, start: pd.Timestamp, end: pd.Timestamp, stat: str = "median"
+    ) -> pd.DataFrame:
+        """Calculate the total carbon generated based on the specified statistic
+        (min, median, max)"""
+        if stat not in ["min", "median", "max"]:
+            raise ValueError("stat must be one of 'min', 'median', or 'max'")
+
+        df = self._query_generation_per_plant(country_code, start, end)
```
spyrostz, line 136: Please use better names for these pandas dataframes. Just naming the dataframe as `df` is equal to a normal Python variable being named as `a`

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,246 @@
...
+    "EE": "EE",
+    "LV": "LV",
+    "LT": "LT",
+    "LU": "LU",
+    "MT": "MT",
+    "CY": "CY",
+}
+
+
+class CarbonEmissionsHandler(EntsoePandasClient, ResultsBaseClass):
```
spyrostz, line 60: I do not particularly like the multiple inheritance here, since the same outcome can be achieved by composition. I think it will be much more readable if you create a separate EntsoeClient class that inherits from EntsoePandasClient, override whatever you want and construct it in the CarbonEmissionsHandler constructor. That way you will avoid multiple inheritance, and the dependencies between the Entsoe client and our logic will be far more decoupled.

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,246 @@
...
+        ]
+        return df_combined
+
+    def update(self, country_code: str, time_slot: duration, results_list: List[Dict]):
+        """Update the results_dict with carbon generated and carbon savings"""
+
+        start_date = pd.Timestamp(results_list[0]["current_market"].split("T")[0])
+        end_date = pd.Timestamp(results_list[-1]["current_market"].split("T")[0])
+        df_carbon_ratio = self.calculate_carbon_ratio(country_code, start_date, end_date)
+        df_carbon_ratio.to_csv("carbon_ratio.csv")
```
spyrostz, line 199: I assume that this is for debugging right ? Please remember to remove before merging

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,246 @@
...
+
+            for area_uuid in result["results"].keys():
+                area_result = result["results"][area_uuid]
+                target_timestamp = pd.Timestamp(current_market).tz_localize("UTC")
+
+                nearest_index = df_carbon_ratio.index.get_indexer(  # the closest timestamp to the target_timestamp
+                    [target_timestamp], method="nearest"
+                )[
+                    0
+                ]
```
spyrostz, line 211: IMO the time resolution of the carbon_ratio should not be calculated here but right after retrieving the data from ENTSOE, essentially the profiles from ENTSOE should be converted to timeseries adhering to the configuration slot_length (therefore, if slot_length=15, 00:00, 00:15, 00:30, 00:45 etc) right at the time that they are first received, and not rely on the "closest timestamp".

spyrostz (commented):
Several comments, will have another look once dealt with!

gsy_framework/sim_results/carbon_emissions.py
BigTava: I will need this method in gsy-e CLI

gsy_framework/sim_results/carbon_emissions.py
BigTava: Thanks for the roast. Changed aux variables naming

gsy_framework/sim_results/carbon_emissions.py
BigTava: Thanks for the suggestion. This was a really good catch. I replaced the multiple inheritance in with composition through the EntsoePandasClientAdapter

gsy_framework/sim_results/carbon_emissions.py
BigTava: Yes, removed, thanks

gsy_framework/sim_results/carbon_emissions.py
BigTava: Hmm... I expect ENTSOE data has the same slot_length in every request. What does not have the same slot_length is our results. So for every result I am finding the closest carbon_ratio data point. But I could indeed already provide the carbon_ratio dataframe with the right slot_length by duplicating the rows or applying an average depending on the slot_length. After I integrate with the gsy-e cli, I will see if using this method still makes sense, and if yes, I will proceed with your suggestion.

hannesdiedrich (commented):
Left comments, let's discuss

requirements/base.ini
```diff
@@ -11,3 +11,7 @@ tabulate
 tox
 websockets
 avro
+entsoe-py>=0.4.4
+pandas>=1.4.2
+geopy
+numpy==1.24.4
```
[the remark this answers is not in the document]
BigTava: It is a dependency of entsoe, but I need to lock this specific numpy version, in order to be compatible

gsy_framework/sim_results/imported_exported_energy.py
```diff
@@ -15,21 +15,26 @@
...
         if not self.should_export_plots:
             # do not accumulate over timestamps for the gsy-web path
             self._init_results_dict()
         for child_dict in area_dict["children"]:
             # loop over all communities
             if not child_dict.get("children"):
                 continue
-            self._accumulate_imported_exported_energy(child_dict, core_stats)
+            if _current_market_slot != "":
+                self._accumulate_imported_exported_energy(child_dict, core_stats)
```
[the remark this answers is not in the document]
BigTava: Well, it is a design choice. In this particular case, I interpret _market_slot as a property of the class and not really an argument for these private functions. Or at least, I am not interested in passing the _market_slot in the interface of these methods. But I can be totally wrong here

gsy_framework/sim_results/imported_exported_energy.py
```diff
@@ -58,9 +63,17 @@ def _prepopulate_results_dict_with_zeros(self, member_uuid_name_mapping: Dict):
             else member_uuid_name_mapping.keys()
         )
         for member_key in members:
+            if member_key not in self.imported_exported_energy:
+                self.imported_exported_energy[member_key] = {}
+                self.imported_exported_energy[member_key]["accumulated"] = {}
```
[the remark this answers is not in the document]
BigTava: You lost me here. I think we had only accumulated before. Let me know how you would like to proceed

gsy_framework/resources/carbon_ratio_per_country/carbon-intensity-electricity.csv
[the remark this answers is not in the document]
BigTava: Thanks!

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,396 @@
...
+import pandas as pd
+from entsoe import EntsoePandasClient
+from entsoe.mappings import lookup_area, Area
+from entsoe.parsers import parse_generation
+from geopy.geocoders import Nominatim
+
+
+ENTSOE_URL = "https://web-api.tp.entsoe.eu/api"
+
+GENERATION_PLANT_TO_CARBON_EMISSIONS = {
```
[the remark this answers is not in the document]
hannesdiedrich: I do not understand why it is handy to have them here in the same module as the code. Especially the GEOPY_TO_ENTSOE_COUNTRY_CODE and MONTHLY_CARBON_EMISSIONS_COUNTRY_CODES.
When you open the file, you first need to scroll through all these to get to the class. so yes, I still think that these belong to a constants.module

gsy_framework/sim_results/imported_exported_energy.py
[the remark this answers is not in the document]
hannesdiedrich: Yes, we had accumulated values before, because we wanted to export plots that show the imported/exported values per home easily for reports.
Now you said that a timeseries would be better for calculating the carbon footprint and I said: "Please keep both".
But now that I see this implementation, I think that only keeping the time-series is sufficient. If we need the accumulated value, we can simple accumulated in posts-processing. 
Does it make sense not?

requirements/base.ini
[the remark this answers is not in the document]
hannesdiedrich: Entoe itself does not require numpy. I think pandas does. Another pro-argument to not use pandas. ;-)

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,432 @@
...
+        df_generation = df_generation.loc[start:end]
+
+        return df_generation
+
+
+class CarbonEmissionsHandler:
+    """The most recent Enstsoe-y version (v0.6.16) is only compatible with python 3.9.
+    Therefore, I am using v0.4.4 whose URL does not work. So, I am overwriting some methods
+    from https://github.com/EnergieID/entsoe-py (branch v0.4.4) to make this work.
+    """
```
hannesdiedrich, line 235: This comment should be in EntsoePandasClientAdapter right? 

> The most recent Enstsoe-y version (v0.6.16) is only compatible with python 3.9

Might it be, that the raw client does not suffer from this limitation? Maybe only pandas expecting 3.11?

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
hannesdiedrich: You can handle this in V2

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,396 @@
...
+            right_index=True,
+            how="inner",
+        )
+        df_carbon_emissions["Carbon Generated (gCO2eq)"] = (
+            df_carbon_emissions["Value"] * df_carbon_emissions["Ratio (gCO2eq/kWh)"]
+        )
+        carbon_generated_g = df_carbon_emissions["Carbon Generated (gCO2eq)"].sum()
+
+        emissions = {"carbon_generated_g": carbon_generated_g}
+        return emissions
```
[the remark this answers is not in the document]
hannesdiedrich: Please answer here.

gsy_framework/sim_results/imported_exported_energy.py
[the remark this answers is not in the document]
hannesdiedrich: please remove the "accumulated" value. Thanks!

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
BigTava: Yes, forgot to move these to a separate file. Thanks!

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,396 @@
...
+
+    def calculate_from_gsy_imported_exported_energy(
+        self, country_code: str, imported_exported_energy: Dict
+    ):
+        """Calculate the carbon emissions from gsy-e imported exported energy
+        which includes imported energy from community and grid."""
+
+        # Find simulation start and end dates
+        for area in imported_exported_energy.values():
```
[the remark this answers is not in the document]
hannesdiedrich: You can remove this loop when you remove the accumulated value from the imported_exported_energy dict.

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,491 @@
...
+        # remove accumulated key
+        for area in imported_exported_energy.values():
+            area.pop("accumulated", None)
+
+        # Find simulation start and end dates
+        first_key = next(iter(imported_exported_energy))
+        area = imported_exported_energy[first_key]
+        dates = [datetime.strptime(timestr, "%Y-%m-%dT%H:%M:%S") for timestr in area.keys()]
+        start_date = min(dates).replace(tzinfo=pytz.UTC).replace(minute=0, second=0, microsecond=0)
+        end_date = max(dates).replace(tzinfo=pytz.UTC).replace(minute=0, second=0, microsecond=0)
```
hannesdiedrich, line 457: This deserves it's own (private) method even now IMO. Thanks!

gsy_framework/sim_results/imported_exported_energy.py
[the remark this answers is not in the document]
BigTava: We are using the [accumulated in gsy-web](https://github.com/gridsingularity/gsy-web/blob/c559e23d88bbcb91c0502e92c914185117fda94d/simulation_results/results_to_db.py#L178). Should I calculate the accumulated there?

gsy_framework/sim_results/imported_exported_energy.py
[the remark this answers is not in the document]
hannesdiedrich: Due to lack of time I will mark this as resolved. Still not a huge fan of this. I may remind you in v2 about it.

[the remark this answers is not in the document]
BigTava: True. Thanks for noticing

tests/test_sim_results/test_carbon_emissions.py
```diff
@@ -0,0 +1,120 @@
...
+        # Given
+        energy_handler = CarbonEmissionsHandler(entsoe_api_key="false_key")
+
+        # When
+        carbon_emissions = energy_handler.calculate_from_gsy_trade_profile(
+            country_code=country_code, trade_profile=self.trade_profile
+        )
+
+        # Then
+        assert carbon_emissions["carbon_generated_g"] > 0
```
hannesdiedrich, line 120: I think for this specific case you can add an expected value. Please do.

hannesdiedrich (commented):
Left some more minor comments and resolved the ones that are probably dealt with in v2. Most of the code here should be moved to gsy-web in v2. Only the multiplication of the  carbon ratios and the imported energy should live in the CarbonEmissionsHandler. With this we can also get rid of the dependency to pandas.
Thanks!

gsy_framework/sim_results/carbon_emissions.py
BigTava: Agree

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,396 @@
...
+                        "simulation_time": pd.to_datetime(timestamp).tz_localize("UTC"),
+                        "imported_from_grid": imported_energy["imported_from_grid"],
+                        "imported_from_community": imported_energy["imported_from_community"],
+                    },
+                    index=[0],
+                )
+                df_total_accumulated = (
+                    pd.concat([df_total_accumulated, df_imported_energy])
+                    .groupby("simulation_time", as_index=False)
+                    .sum()
```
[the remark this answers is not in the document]
hannesdiedrich: Please answer

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
hannesdiedrich: Sorry, this is not the answer to my question. Why are the returned values or structure different in these methods? I get that the input values are different, but the return value should look the same, right?

tests/test_sim_results/test_carbon_emissions.py
BigTava: This tests 72 different values. But I can add them, yes

gsy_framework/community_datasheet/row_converters.py
```diff
@@ -41,6 +41,8 @@
     "Enable grid import fee": "enable_grid_import_fee_const",
     "Enable grid export fee": "enable_grid_export_fee_const",
     "Enable taxes surcharges": "enable_taxes_surcharges",
+    "Enable electricity tax": "enable_electricity_tax",
+    "Enable fixed monthly fee": "enable_fixed_monthly_fee",
```
hannesdiedrich, line 45: these are duplicated now, right?

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
BigTava: I am concatenating the df_imported_energy, grouping by simulation_time, and if two instances have the same simulation_time, then it will sum

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,491 @@
...
+
+        return df_numeric
+
+    def calculate_carbon_ratio(
+        self, country_code: str, start: pd.Timestamp, end: pd.Timestamp, stat: Stat = Stat.MEDIAN
+    ) -> pd.DataFrame:
+        """Calculate the total carbon generated based on the specified statistic
+        (min, median, max)"""
+        if stat not in [Stat.MIN, Stat.MEDIAN, Stat.MAX]:
+            raise ValueError("stat must be one of 'min', 'median', or 'max'")
```
hannesdiedrich, line 301: The beauty of enums is that you do no need these kind of validators.

gsy_framework/community_datasheet/row_converters.py
BigTava: I don't remember why this here TBH. Probably some fix I did while testing the whole thing. I'll check

gsy_framework/sim_results/carbon_emissions.py
BigTava: True. Thanks

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
hannesdiedrich: Ah , I see now, you are accumualting over all child areas, right?
IMO this also deserves its own method for better readability.

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,491 @@
...
+        which includes imported energy from community and grid."""
+
+        # remove accumulated key
+        for area in imported_exported_energy.values():
+            area.pop("accumulated", None)
+
+        # Find simulation start and end dates
+        first_key = next(iter(imported_exported_energy))
+        area = imported_exported_energy[first_key]
+        dates = [datetime.strptime(timestr, "%Y-%m-%dT%H:%M:%S") for timestr in area.keys()]
```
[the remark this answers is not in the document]
BigTava: Thanks

gsy_framework/community_datasheet/row_converters.py
BigTava: Yes, this fix was already merged

gsy_framework/sim_results/carbon_emissions.py
```diff
@@ -0,0 +1,396 @@
...
+        """Calculate the carbon emissions from gsy-e imported exported energy
+        which includes imported energy from community and grid."""
+
+        # Find simulation start and end dates
+        for area in imported_exported_energy.values():
+            area.pop("accumulated")
+            dates_keys = list(area.keys())
+            dates = sorted(set(pd.to_datetime(list(dates_keys), format="%Y-%m-%dT%H:%M:%S").date))
+            start_date = pd.Timestamp(datetime.combine(dates[0], datetime.min.time())).round("H")
+            end_date = pd.Timestamp(datetime.combine(dates[-1], datetime.max.time())).round("H")
```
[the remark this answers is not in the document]
BigTava: Actually, I remember that I did use the DATE_TIME_FORMAT when I first wrote, but I got this error
'2024-01-01T00:00:00' does not match format 'YYYY-MM-DDTHH:mm'

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
BigTava: total_emissions = df_carbon_emissions.drop(columns=["simulation_time"]).sum().to_dict() returns {"carbon_generated_base_case_g", "carbon_savings_g", "imported_from_community", "imported_from_grid", "carbon_savings_g"}

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
BigTava: If you insist I will isolate in its own method. But since it is not being reused I would prefer not to, since it loose functions become harder to debug.

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
hannesdiedrich: Which method emitted this error? How did you apply the DATE_TIME_FORMAT ?

gsy_framework/sim_results/carbon_emissions/results.py
hannesdiedrich, line 1: Why did you move this method in the same commit as other changes? This will hide all small changes, because the whole module is marked as changed. Please omit doing that in the future.

tests/test_sim_results/test_carbon_emissions.py
hannesdiedrich: I soo now. testing it for two countries could have been sufficient, but ok. Thanks

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
hannesdiedrich: Ok, I get your point now, but where is `calculate_from_gsy_trade_profile` used? I can not find it any more.

gsy_framework/sim_results/carbon_emissions/results.py
```diff
@@ -0,0 +1,346 @@
...
+        """Calculate the carbon emissions from gsy-e imported exported energy
+        which includes imported energy from community and grid."""
+
+        # remove accumulated key
+        for area in imported_exported_energy.values():
+            area.pop("accumulated", None)
+
+        first_key = next(iter(imported_exported_energy))
+        area = imported_exported_energy[first_key]
+        start_date, end_date = self._find_start_and_end_dates(area.keys(), "%Y-%m-%dT%H:%M:%S")
```
hannesdiedrich, line 310: Why not moving line 308 and 309 also into `_find_start_and_end_dates`  and only pass the `imported_exported_energy` ?

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
hannesdiedrich: `loose functions become harder to debug` why is that? IMO the readability, testability and debugging ability increases when putting every step into separate methods. 
If you follow this approach, the `calculate_from_gsy_imported_exported_energy` really quick to understand, because it clearly consists of 4 steps:
1. get start and end date
2. calculate carbon ratio
3. accumulate energy over child areas
4. calculate_from_carbon_ratio_and_imported_energy
This will be important for later and for debugging.

gsy_framework/sim_results/carbon_emissions/results.py
```diff
@@ -0,0 +1,346 @@
...
+            ]
+
+        # fill df_carbon_ratio in case of missing hours using the last value
+        df_carbon_ratio = df_carbon_ratio[["Ratio (gCO2eq/kWh)"]]
+        df_carbon_ratio = df_carbon_ratio.reindex(
+            pd.date_range(start=start, end=end, freq="H", tz="UTC"), method="pad"
+        )
+        return df_carbon_ratio
+
+    def calculate_from_carbon_ratio_and_imported_energy(
```
hannesdiedrich, line 214: I only realised it now, when I was writing another comment, that this method name is unclear.
What is calculated? 
How about renaming it to 
`calculate_carbon_emissions_from_imported_energy `
and 
`calculate_from_gsy_trade_profile` to `calculate_carbon_emissions_from_trade_profile ` ?

gsy_framework/sim_results/imported_exported_energy.py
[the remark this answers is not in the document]
spyrostz: Agreed and already filed a ticket https://gridsingularity.atlassian.net/browse/GSYE-847

gsy_framework/sim_results/carbon_emissions/results.py
BigTava: Those were the original names. But, the name calculate_carbon_emissions_from_carbon_ratio_and_imported_energy is too long, so I shorted everything to calculate_from_[arguments]. Truth is, the class name is CarbonEmissions, so it is redundant to call calculate_carbon_emissions_from_[], IMO. Let me know, if you would like the renaming

gsy_framework/sim_results/carbon_emissions/results.py
BigTava: I am reusing _find_start_and_end_dates in other function which does not take a dict.

gsy_framework/sim_results/carbon_emissions/results.py
BigTava: Yes, it would have been clearer to do it in separate commits.

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
BigTava: This pd.to_datetime(list(dates_keys), format="%Y-%m-%dT%H:%M:%S") emitted. In fact, the timestamps take seconds, and DATE_TIME_FORMAT only considers the minutes

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
BigTava: It is used in gsy-e and at the moment in research-scripts

gsy_framework/sim_results/imported_exported_energy.py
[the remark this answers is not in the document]
BigTava: we are saving the accumulated value per market slot already. The results are being broadcasted to gsy-web every market slot.

gsy_framework/sim_results/carbon_emissions/results.py
hannesdiedrich: This is why I removed the "carbon_ratio" part in my suggestion, because this is needed in both cases anyway.

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
hannesdiedrich: I got that. However, why using pd for this? I showed you a solution last week (original comment).

gsy_framework/sim_results/carbon_emissions/results.py
BigTava: There are 3 main functions:
calculate_from_carbon_ratio_and_imported_energy
calculate_from_gsy_trade_profile
calculate_from_gsy_imported_exported_energy

Only the first needs the carbon_ratio as an argument

gsy_framework/sim_results/carbon_emissions.py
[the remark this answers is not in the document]
BigTava: Yes, I replaced it with datetime.strptime(timestr, datetime_format) as seen in line 115

gsy_framework/sim_results/carbon_emissions/results.py
hannesdiedrich: OK.

gsy_framework/sim_results/carbon_emissions/results.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def _find_start_and_end_dates(
+        self, dates: List, datetime_format: str = DATE_TIME_FORMAT
+    ) -> Tuple[datetime, datetime]:
+        """Find the start and end dates from a list of string timestamps"""
+
+        dates = [datetime.strptime(timestr, datetime_format) for timestr in dates]
+        start_date = min(dates).replace(tzinfo=pytz.UTC).replace(minute=0, second=0, microsecond=0)
+        end_date = max(dates).replace(tzinfo=pytz.UTC).replace(minute=0, second=0, microsecond=0)
+        return start_date, end_date
```
hannesdiedrich, line 118: If you always overwriting the `datetime_format` when calling this method, the default value does not make sense.
You are propably overwriting it because it did not work for `strptime`, right?
This is because this format string is for pendulum datetimes.
If you would followed my original suggestion, you would have been more lucky:
https://github.com/gridsingularity/gsy-framework/pull/556#discussion_r1925118474
Let me rewrite this method how I think it should look:
```suggestion
self, times: List) -> Tuple[datetime, datetime]:
        """Find the start and end dates from a list of string timestamps"""

        dates = [from_format(timestr, DATE_TIME_FORMAT) for timestr in times]
        start_date = min(dates).in_tz(TIME_ZONE).set(minute=0, second=0, microsecond=0)
        end_date = max(dates).in_tz(TIME_ZONE).set(minute=0, second=0, microsecond=0)
        return start_date, end_date
```
This will not only enforce our universally used DATE_TIME_FORMAT but also TIME_ZONE.

I have an additional question. Why do we select only full days here and not allow also to query for multiple hours or partial days?

gsy_framework/sim_results/carbon_emissions/results.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def calculate_from_gsy_trade_profile(self, country_code: str, trade_profile: Dict) -> Dict:
+        """Calculate the carbon emissions from gsy-e trade profile,
+        usually from InfiniteBus accummulated sold_energy"""
+
+        if len(trade_profile.keys()) == 0:
+            raise ValueError("Invalid trade profile")
+
+        start_date, end_date = self._find_start_and_end_dates(
+            trade_profile.keys(), "%B %d %Y, %H:%M h"
```
hannesdiedrich, line 268: Why does the trade profile have a different time format than the imported_exported energy?
We should definitely look into this. Can you please leave a TODO comment in code? Thanks!

hannesdiedrich (approved):
LGTM

PR #561, opened 2024-12-16
FE-550-2: create _default_grid_root method in CommunityDataSheetParser

BigTava:


gsy_framework/community_datasheet/community_datasheet_parser.py
```diff
@@ -141,11 +135,21 @@ def _create_grid(self) -> Dict:
                     "allow_external_connection": False,
                     "uuid": str(uuid.uuid4()),
                     "type": "Area",
-                    "children": grid,
+                    "children": None if children is None else children,
```
hannesdiedrich, line 138: I think you missed the case when the user uploads a CDS in which case the `_create_grid` method should be called here.

Additionally, the "children" member should never be None. We hat lot's of issues in the past with that. I think we fixed them all, but better be save than sorry and add an empty list here.

PR #563, opened 2024-12-18
feature/RESEARCH-315: add heat pump tab to CDS and improve CDS readibility

BigTava:


gsy_framework/community_datasheet/row_converters.py
```diff
@@ -309,84 +313,74 @@ def _validate_row(cls, row: Dict):
...
 
     @classmethod
     def convert(cls, row: Dict) -> Dict:
-        """
-        Convert the row parsed from the Excel file to the asset representation used in scenarios.
-        """
         cls._validate_row(row)
         return {
-            "name": row[PVSheetHeader.PV_NAME],
+            "name": row[PVSheetHeader.PV_NAME.value],
```
spyrostz, line 345: Even though I think that this will not crash anything, isn't PV_NAME a string already? If this is the case, then the `.value` is not needed, same for all other Header class variables.

spyrostz (commented):
Some comments, will LGTM once clarified

gsy_framework/community_datasheet/row_converters.py
BigTava: It might be a little bit more confusing, but I am overriting the value of Enum to lower each value so it is clear from the code what each enum means + allowing parsing the column headers in CDS with -- user errors

PR #578, opened 2025-02-27
GSYE-857: Corrected timezone related errors on the static carbon foot…

spyrostz:
…print parsing. Fixed carbon data profile path. Added carbon data profiles to the gsy-framework installation.

## Proposed changes

-

**GSY_WEB_BRANCH**=master

### DIFF
--- a/gsy_framework/sim_results/carbon_emissions/results.py
+++ b/gsy_framework/sim_results/carbon_emissions/results.py
@@ -1,6 +1,5 @@
 from typing import Dict, List, Tuple
 import csv
-import os
 
 import pendulum
 from pendulum import DateTime
@@ -8,6 +7,7 @@ from geopy.geocoders import Nominatim
 from iso3166 import countries
 
 from gsy_framework.constants_limits import TIME_ZONE
+from gsy_framework.resources import STATIC_FILES_DIR
 from gsy_framework.utils import str_to_pendulum_datetime
 from gsy_framework.sim_results.carbon_emissions.constants import (
     MONTHLY_CARBON_EMISSIONS_COUNTRY_CODES,
@@ -62,8 +62,7 @@ class CarbonEmissionsHandler:
             country_code in MONTHLY_CARBON_EMISSIONS_COUNTRY_CODES
         ):  # source: https://www.electricitymaps.com/data-portal
             file_path = (
-                f"{os.path.dirname(os.path.abspath(__file__))}/../../resources/"
-                f"carbon_ratio_per_country/{country_code}_2023_monthly.csv"
+                f"{STATIC_FILES_DIR}/carbon_ratio_per_country/{country_code}_2023_monthly.csv"
             )
 
             with open(file_path, mode="r", encoding="utf-8") as carbon_file:
--- a/gsy_framework/sim_results/carbon_emissions/results.py
+++ b/gsy_framework/sim_results/carbon_emissions/results.py
@@ -46,13 +46,16 @@ class CarbonEmissionsHandler:
         return [start.add(hours=hour) for hour in range(int((end - start).total_hours()) + 1)]
 
     def query_carbon_ratio_from_static_source(
-        self, country_code: str, start: DateTime, end: DateTime
+        self, country_code: str, start_time: DateTime, end_time: DateTime
     ) -> List[Dict]:
         """Calculate the total carbon generated based on the specified statistic"""
 
-        if start.tzinfo is None or end.tzinfo is None:
+        if start_time.tzinfo is None or end_time.tzinfo is None:
             raise ValueError("start and end must have timezone")
-        if start.astimezone().tzname() != TIME_ZONE or end.astimezone().tzname() != TIME_ZONE:
+        if (
+            start_time.astimezone().tzname() != TIME_ZONE
+            or end_time.astimezone().tzname() != TIME_ZONE
+        ):
             raise ValueError("start and end must be in UTC+0")
 
         if (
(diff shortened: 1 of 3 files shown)


gsy_framework/sim_results/carbon_emissions/results.py
```diff
@@ -42,31 +42,40 @@ def _find_start_and_end_dates(self, dates: List) -> Tuple[DateTime, DateTime]:
...
         return [start.add(hours=hour) for hour in range(int((end - start).total_hours()) + 1)]
 
     def query_carbon_ratio_from_static_source(
-        self, country_code: str, start: DateTime, end: DateTime
+        self, country_code: str, start: DateTime, end: DateTime, file_prefix=""
     ) -> List[Dict]:
         """Calculate the total carbon generated based on the specified statistic"""
 
+        if not file_prefix:
+            file_prefix = "gsy_framework"
```
hannesdiedrich, line 57: uff, I am not a fan of this solution unfortunately. 
How about doing something I did in my latest gsy-web PR:
1. create a variable that holds the path to the resources https://github.com/gridsingularity/gsy-web/pull/2285/files#diff-d37ae1bfda98224e3012b3ce64f0b49bf6e7ecf1d3016645a6e5bb85c05e4e13R3
2. re-use it by importing when opening a file: https://github.com/gridsingularity/gsy-web/pull/2285/files#diff-d43286a340dfb891017855642cac303271c1c8aa3318652e5d96d046fda405f3R154

gsy_framework/sim_results/carbon_emissions/results.py
```diff
@@ -42,31 +42,40 @@ def _find_start_and_end_dates(self, dates: List) -> Tuple[DateTime, DateTime]:
...
+
         if start.tzinfo is None or end.tzinfo is None:
             raise ValueError("start and end must have timezone")
-        if start.timezone_name != TIME_ZONE or end.timezone_name != TIME_ZONE:
+        if start.astimezone().tzname() != TIME_ZONE or end.astimezone().tzname() != TIME_ZONE:
```
hannesdiedrich, line 61: why was this needed? Arend the `start` and `end` pendulum times?

gsy_framework/sim_results/carbon_emissions/results.py
```diff
@@ -42,31 +42,40 @@ def _find_start_and_end_dates(self, dates: List) -> Tuple[DateTime, DateTime]:
...
         """Generate a list of datetime objects for each hour between start and end"""
+        if not isinstance(start, pendulum.DateTime):
+            start = pendulum.instance(start)
+        if not isinstance(end, pendulum.DateTime):
+            end = pendulum.instance(end)
         return [start.add(hours=hour) for hour in range(int((end - start).total_hours()) + 1)]
 
     def query_carbon_ratio_from_static_source(
-        self, country_code: str, start: DateTime, end: DateTime
+        self, country_code: str, start: DateTime, end: DateTime, file_prefix=""
```
hannesdiedrich, line 52: NIT: Let's also rename it here to
```suggestion
        self, country_code: str, start_time: DateTime, end_time: DateTime, file_prefix=""
```

hannesdiedrich (approved):
Feel free to merge in order to get done with it (approving already). However, I am not a fan of the implementation.

gsy_framework/sim_results/carbon_emissions/results.py
spyrostz: Let me check if this is no longer needed. It was crashing before I converted the start / end time from normal datetimes (received from GraphQL / graphene) to penulum datetimes, but I have meanwhile converted those to pendulum

gsy_framework/sim_results/carbon_emissions/results.py
spyrostz: Fine for here, and let me adapt it. However this solution will not work when importing gsy-framework in another repo (e.g. gsy-web) because this file ends up in the site-packages directory, therefore we need to refer it from the imported package and not from the project that imports it. 
Let me change it here though so that it is resolved.

gsy_framework/sim_results/carbon_emissions/results.py
spyrostz: It is required because this clause used to be wrong in the first place. TIME_ZONE is a string, and start / end are DateTime objects, therefore we really need the timezone string of the datetime for comparison.

gsy_framework/sim_results/carbon_emissions/results.py
spyrostz: Tested it and you were right, the relative path of the current file worked! Thanks for the suggestion

gsy_framework/sim_results/carbon_emissions/results.py
hannesdiedrich: Yes, and start.timezone_name returns a string which should have worked also in the original version.

gsy_framework/sim_results/carbon_emissions/results.py
hannesdiedrich: > Fine for here, and let me adapt it. However this solution will not work when importing gsy-framework in another repo (e.g. gsy-web) because this file ends up in the site-packages directory, therefore we need to refer it from the imported package and not from the project that imports it.

I understood, that my suggestion saves the absolute path of the static file directory to the STATIC_TEST_FILE_DIR when initialising. So, when you import it from another repo, this should not be untouched and point to the site-packages folder where it lives, or did I get it wrong?

spyrostz: Yes, but this is what I do not understand:
> that my suggestion saves the absolute path of the static file directory to the STATIC_TEST_FILE_DIR when initialising

I assume that by "static file directory" you mean the `gsy_framework/resources/` path?

hannesdiedrich: yes, sorry, I should not have referred to my but this situation here.

hannesdiedrich (dismissed):

PR #587, opened 2025-03-06
Early return in the read_arbitrary_profile method for the case of emp…

spyrostz:
…ty profile.

## Proposed changes

-

**GSY_WEB_BRANCH**=master


hannesdiedrich:
I am not sure if we mask a real issue with this. Let's keep it in our memories and remember when we see other issues in the profiles. Thanks!

PR #590, opened 2025-03-13
FE-605: AAdded community imported exported energy and not just the ho…

spyrostz:
…me one.

## Proposed changes

-

**GSY_WEB_BRANCH**=master

### DIFF
--- a/gsy_framework/sim_results/imported_exported_energy.py
+++ b/gsy_framework/sim_results/imported_exported_energy.py
@@ -81,10 +81,12 @@ class ImportedExportedEnergyHandler(ResultsBaseClass):
         )
         # Adding community name and uuid to the member uuid name mapping to include it in the
         # imported exported results
-        member_uuid_name_mapping[community_dict["uuid"]] = community_dict["name"]
-        self._prepopulate_results_dict_with_zeros(member_uuid_name_mapping)
+        self._prepopulate_results_dict_with_zeros(
+            {**member_uuid_name_mapping, community_uuid: community_dict["name"]}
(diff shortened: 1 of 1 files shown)


hannesdiedrich (dismissed):
Interesting!

gsy_framework/sim_results/imported_exported_energy.py
```diff
@@ -109,6 +112,12 @@ def _accumulate_imported_exported_energy(self, community_dict: Dict, core_stats:
                     "imported_from_grid",
                     trade["energy"],
                 )
+                # Track community imported energy from grid
+                add_or_create_key(
+                    self.imported_exported_energy[community_key][self._market_slot],
+                    "imported_from_grid",
```
hannesdiedrich, line 118: If the community is the seller, this should be: "exported_to_grid", right?
And in line 133 it should be "imported_from_grid". Or is my brain not working correctly?

spyrostz: No, it is actually the other way around: if the community is the seller name, at least with the current GSY Maps communities, this implies that for both the home and the community the energy is imported from the grid. In this market, the flow of such a trade is `InfiniteBus -> Community -> Home 1`. The export direction is `Home 1 -> Community -> Infinite Bus` in which case the Community is the buyer.

hannesdiedrich: Ah yes, this is the Community area agent that is connected to the higher market. I always confuse this. Thanks a lot for explaining!

PR #593, opened 2025-03-28
GSYE-875: Remove 'trades' from scm_simulation_raw_data avro schema be…

hannesdiedrich:
…cause it is not sent by the gsy-e any more

## Proposed changes

-

**GSY_WEB_BRANCH**=master


spyrostz (approved):
LGTM

PR #598, opened 2025-05-22
GSYE-860: Improve get_from_profile_same_weekday_and_time

hannesdiedrich:

## Proposed changes

-

**GSY_WEB_BRANCH**=master

### DIFF
--- a/gsy_framework/utils.py
+++ b/gsy_framework/utils.py
@@ -132,6 +132,13 @@ def generate_market_slot_list(start_timestamp=None):
     return market_slot_list
 
 
+def _find_first_full_day_in_profile(indict: Dict) -> Optional[DateTime]:
+    for time_slot in indict.keys():
+        if time_slot.hour == 0 and time_slot.minute == 0:
+            return time_slot
+    return None
+
+
 def get_from_profile_same_weekday_and_time(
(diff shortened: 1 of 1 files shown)


gsy_framework/utils.py
```diff
@@ -132,6 +132,13 @@ def generate_market_slot_list(start_timestamp=None):
     return market_slot_list
 
 
+def _find_first_full_day_in_profile(indict: Dict) -> Optional[DateTime]:
+    for time_slot in indict.keys():
+        if time_slot.hour == 0 and time_slot.minute == 0:
+            return time_slot
+    return None
```
spyrostz, line 139: Just to confirm, I do not recall if this method is triggered for the case that no rotation is performed. If not, then this method is good to go.

spyrostz (approved):
LGTM

gsy_framework/utils.py
hannesdiedrich: AFAIS, `get_from_profile_same_weekday_and_time` is only used inside of update_frequency (where all profiles are rotated) and the validation methods in the strategies in order to pass the rates to the update_frequency.
So, IMO we should be good to go with this. After the merge I will run some simulations and scms just in case

